PR #2, opened 2020-05-06
Move metadata code to load_omero_metadata

will-moore:
The image can still be opened if the metadata loading fails


joshmoore:
:+1: for more methods. I'll add some on top of this.

PR #4, opened 2020-05-06
is_ome_zarr: check for "multiscales" for resolution loading

joshmoore:
Otherwise, check if a Zarr is just an array and if so, load it.
No handling of non-multiscale groups is currently supported.


will-moore:
```$ napari '/tmp/6001240.zarr/0'```
works for me, showing a 2-channel image (instead of the napari default of showing single channel n-dim image).
👍

joshmoore:
This has been released as 0.0.1.

PR #10, opened 2020-05-08
Workaround pyramids of length 1

joshmoore:
Passing a pyramid of length 1 leads to the exception:

```
AttributeError: 'list' object has no attribute 'shape'
```

see: https://github.com/napari/napari/issues/1225


will-moore:
Makes sense. 👍

joshmoore:
Thanks. Releasing a quick 0.0.3

PR #11, opened 2020-05-08
Test image use zattrs for omero

will-moore:
Since we now store omero metadata in the zarr attrs, we have access to this data after download.
Updating the test image creation and the tests accordingly.


joshmoore:
Nice. Thanks, @will-moore

PR #12, opened 2020-05-11
Readme

jburel:
* Fix formatting
 * Ship correct version of License file


joshmoore:
Actually, let's go with BSD: see https://github.com/ome/ome-zarr-py/pull/15

PR #13, opened 2020-05-11
missing dependencies

jburel:
I installed the dev version in conda env 
Tried to run 
```
ome_zarr download https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001240.zarr/
```
This returns
```
 No module named 'pluggy'
```

This PR adds ``pluggy`` as a dependency

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,6 +19,7 @@ install_requires += ['s3fs'],
 install_requires += ['requests'],
 install_requires += ['toolz'],
 install_requires += ['vispy'],
+install_requires += ['pluggy'],
 
 
 setup(


joshmoore:
Thanks, @jburel. Merging.

For the napari-only dependencies, it might make sense to make the imports optional and revert this addition since pluggy _will_ be installed if napari is already installed.

will-moore:
@joshmoore Are you sure that ```pluggy``` is installed with napari? I seem to remember having to install it separately, and I don't see it in https://github.com/napari/napari/blob/master/requirements/default.txt

joshmoore:
It's a dependency of:

```
requirements/default.txt:napari-plugin-engine>=0.1.4
```

tlambert03:
Sorry, pluggy is not a dependency anywhere anymore.  The plugin engine is a fork/replacement of pluggy.  You shouldn’t need pluggy anywhere, (though if you install pytest, you’ll have it because they depend on pluggy)

tlambert03:
Oh, and we’re recommending that you install the plugin marker from napari_plugin_engine... (though technically they are still compatible, which is why your code works). We updated our docs too https://napari.org/docs/plugins/for_plugin_developers.html

PR #15, opened 2020-05-11
README & LICENSE

joshmoore:
Move to BSD license and migrate @jburel's change from https://github.com/ome/ome-zarr-py/pull/12.


sbesson:
:+1: for switching to BSD-2 clause

jburel:
License file now matches what is declared in the README
The switch of license also makes sense

PR #16, opened 2020-05-11
download: exit early if not an ome-zarr

joshmoore:
cc: @jburel


jburel:
This means that ``4007801.zarr`` will have to be regenerated to be an ``ome-zarr``

joshmoore:
It can't really live on minio due to the number of files per directory.

jburel:
That's fine

will-moore:
👍

PR #18, opened 2020-05-11
Depend on napari_plugin_engine

tlambert03:
This is a modification to #13.

You are welcome to reject this PR and continue depending on `pluggy` for the `HookImplMarker` if you'd like, we have no plans to break that backward compatibility.  But we are recommending that people use the `napari_plugin_engine` in case we add/remove capabilities going forward.  For instance, one feature that you can only get from napari_plugin_engine is the ability to mark functions as providing napari hookspec, [without renaming the function itself](https://napari.org/docs/plugins/for_plugin_developers.html#matching-hook-implementations-to-specifications).  (that feature merged into pluggy master a while ago, but the timeline of pluggy releases is unclear at the moment)

### DIFF
--- a/ome_zarr.py
+++ b/ome_zarr.py
@@ -24,7 +24,7 @@ from dask.diagnostics import ProgressBar
 from vispy.color import Colormap
 
 from urllib.parse import urlparse
-from pluggy import HookimplMarker
+from napari_plugin_engine import napari_hook_implementation
 
 import logging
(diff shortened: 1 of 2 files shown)


joshmoore:
Thanks @tlambert03. I've opened https://github.com/ome/ome-zarr-py/pull/20 where I try to build on top of this. Opinions welcome.

PR #20, opened 2020-05-12
Replace pluggy with optional napari_plugin_engine

joshmoore:
In addition to the switch from pluggy in https://github.com/ome/ome-zarr-py/pull/18 (thanks @tlambert03) this also makes the napari requirement optional.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ python:
   - "3.6"
 # command to install dependencies
 install:
-  - pip install .
+  - pip install .[napari]
   - pip install scikit-image  # only needed for tests
 # command to run tests
 script: pytest
--- a/ome_zarr.py
+++ b/ome_zarr.py
@@ -24,7 +24,7 @@ from dask.diagnostics import ProgressBar
 from vispy.color import Colormap
 
 from urllib.parse import urlparse
-from pluggy import HookimplMarker
+from napari_plugin_engine import napari_hook_implementation
 
 import logging
 # DEBUG logging for s3fs so we can track remote calls
@@ -38,9 +38,6 @@ LayerData = Union[Tuple[Any], Tuple[Any, Dict], Tuple[Any, Dict, str]]
 PathLike = Union[str, List[str]]
 ReaderFunction = Callable[[PathLike], List[LayerData]]
 # END type hint stuff.
-
-napari_hook_implementation = HookimplMarker("napari")
-
 
 
 @napari_hook_implementation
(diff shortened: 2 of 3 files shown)


joshmoore:
Note: currently the behavior seems to be that the plugin still gets detected but it does not get invoked:

```
(py36) /opt/ome-zarr-py $napari /tmp/6001240.zarr/
INFO:napari_plugin_engine.manager:loaded 3 plugins:
  <module 'napari.plugins._builtins' from '/opt/anaconda/envs/py36/lib/python3.6/site-packages/napari/plugins/_builtins.py'>
  <module 'napari_svg' from '/opt/anaconda/envs/py36/lib/python3.6/site-packages/napari_svg/__init__.py'>
  <module 'ome_zarr' from '/opt/ome-zarr-py/ome_zarr.py'>
  <module 'napari_plugin_engine' from '/opt/anaconda/envs/py36/lib/python3.6/site-packages/napari_plugin_engine/__init__.py'>
```

I'm unsure if we couldn't go further and have the plugin not detected at all.

tlambert03:
couple thoughts:
1. `napari_plugin_engine` is a small pure standard-lib dependency (it doesn't install napari).  It was designed to be a lightweight dependency.  (I definitely also understand wanting to keep the deps list as short as possible... just wanted you to know that it was strictly stdlib).  And that `hook_implementation` decorator does very little ... it just adds a small bit of metadata to the function.

2. If you _do_ decide to make all napari stuff an extra... then you might just consider making the extra actually just be napari proper: `"napari": ["napari"]` and not the plugin engine.  i suspect people using `pip install ome-zarr[napari]` actually intended to use napari... whereas with your current extra they would need to do `pip install napari ome-zarr` (and the `[napari]` extra would be irrelevant ... because `napari-plugin-engine` already comes with napari.

3. your current "noop" wrapper would actually nuke the function altogether (turn it into a pass).  maybe that's what you were looking to do... but if you just want a do-nothing decorator, I'd go with:
```python
def napari_hook_implementation(func, *args, **kwargs): 
    return func
```

> I'm unsure if we couldn't go further and have the plugin not detected at all.

not totally sure what you mean.  Currently, if you decorate your function with the @napari_hook_implementation it will be detected _and_ registered.  the only way to not have it detected at all is to just not decorate it.  I will mention that will will probably soon be [moving away from auto-registration](https://github.com/napari/napari/issues/1238)... leaving it up to the user which of the detected plugins they would actually like to register.

joshmoore:
Followed @tlambert03's suggestions and am now installing all of napari with the extras. Going to roll this out once the tests are green so everyone will be working from the same version for testing this week.

PR #22, opened 2020-05-13
better error handling

joshmoore:
Thanks to @tlambert03 and https://github.com/napari/napari/pull/1219, starting the plugin with a bad URL showed much more information about what went wrong. With this PR, we catch the exception and log it at `WARNING` which is _not_ shown to the user by default:

```
(py36) /opt/ome-zarr-py $napari http://localhost:8080/idr/zarr/v0.1/151.zarr
11:32:27 ERROR PluginError: Error in plugin 'builtins', hook 'napari_get_reader'
  Cause was: ValueError('Not a zarr dataset or group: http://localhost:8080/idr/zarr/v0.1/151.zarr',)
    in file: /opt/anaconda/envs/py36/lib/python3.6/site-packages/napari/utils/io.py
    at line: 272
    package: napari
    version: 0.3.1

11:32:27 WARNING No plugin found capable of reading 'http://localhost:8080/idr/zarr/v0.1/151.zarr'.
11:32:27 ERROR (1) error occured in plugins: 'builtins'. See full error logs in "Plugins → Plugin Errors..."
```

### DIFF
--- a/ome_zarr.py
+++ b/ome_zarr.py
@@ -120,7 +120,7 @@ class BaseZarr:
             data = da.from_zarr(f"{self.zarr_path}")
             return [(data, {'channel_axis': 1})]
 
-    def load_omero_metadata(self):
+    def load_omero_metadata(self, assert_channel_count=None):
         """Load OMERO metadata as json and convert for napari"""
         metadata = {}
         try:
@@ -129,7 +129,20 @@ class BaseZarr:
             if rdefs:
                 model = rdefs.get('model', 'unset')
 
-            channels = self.image_data.get('channels', {})
+            channels = self.image_data.get('channels', None)
+            if channels is None:
+                return {}
+
+            count = None
+            try:
+                count = len(channels)
+                if assert_channel_count:
+                    if count != assert_channel_count:
+                        LOGGER.error((f"unexpected channel count: "
+                                      "{count}!={assert_channel_count}"))
+                        return {}
+            except:
+                LOGGER.warn(f"error counting channels: {channels}")
 
             colormaps = []
             contrast_limits = [None for x in channels]
@@ -192,7 +205,8 @@ class BaseZarr:
 
         if len(pyramid) == 1:
             pyramid = pyramid[0]
-        metadata = self.load_omero_metadata()
+
+        metadata = self.load_omero_metadata(data.shape[1])
         return (pyramid, {'channel_axis': 1, **metadata})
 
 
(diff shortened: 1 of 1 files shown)


tlambert03:
Glad that came in handy!  Just in case you missed it, another recently merged feature that you might find useful is the ability to pass a plugin like `--plugin ome-zarr` on the CLI: https://github.com/napari/napari/pull/1220

(Still no plugin params though)

joshmoore:
@mtbc @will-moore : I've add on some more fault-handling code here since this PR has the logging in place (probably best to get it merged sooner rather than later).

I could use some feedback on https://github.com/ome/ome-zarr-py/pull/22/commits/c20db1758bc64f6382d94010590cb79033a6c89e:

 - should omero-ms-zarr be passing back `label` at all cost?
 - are there default values we can specify for start/end that would be more robust?
 - what values can safely be missing from the metadata that's passed to napari?

mtbc:
All good questions! For each key it could be good to specify from where in OMERO to get the value and, if missing and required, what to invent and how.

jburel:
If you fail to retrieve the rendering settings, you could also fall back to the pixels type to set the range but that will mean implementing some logic to set the start/end interval

will-moore:
I think all of the metadata can be omitted and napari will happily use defaults. The only exception is probably window start/end which are used for contrast setting.
If we don't set these then napari may try to load a ton of data to try and establish min/max.
As @jburel says, we could try to return min/max based on pixel type (since we have some data in hand, but as we were discussing for OMERO, the full range can give a poor experience with rendering sliders. We could potentially get a sample plane (mid Z/T index) and use that to estimate range (e.g. min = 0 and max = plane.max * 2), which could give a more useful range that pixelType max.
But choosing the plane is maybe a bit trickier with multi-resolution images (unless we use the lowest resolution which might be a single tile). I think napari loads that anyway to use for the channel thumbnail.

joshmoore:
@will-moore : :+1: are there any concrete changes you would make to this patch? (Would also be good to eventually encapsulate that in the spec.md as @mtbc suggests)

will-moore:
I don't know if we want to check that the ```len(channels)``` is correct (same length as the C-dimension of the pixel data), or just assume it is?
If it's not, napari fails as above.

joshmoore:
That all makes sense and I'll update, but I think I was more concerned about what happens if an internal element is missing. (In this case, a missing `label` key was throwing an exception.)

will-moore:
OK, understood.
It seems from my test above that napari is not grabbing all the remote data to choose contrast limits. So you can probably leave ```metadata['contrast_limits']``` unset if you don't have a 'end' value for every channel? Probably safer than passing an incomplete list.

joshmoore:
Ok. Added those checks. Have a look at the last two commits.

jburel:
test failed

will-moore:
Code changes look good, thanks

will-moore:
👍

PR #28, opened 2020-06-02
omit 'channel_axis' for non-ome_zarrs (fix #25)

### LINKED ISSUE #25: Don't hardcode channel axis, or any other metadata really, if array is not OME?
Hardcoding the dimensions to TCZYX and channel_axis=1 makes sense when the array can be ascertained to be an OME array, but it seems pretty bold to do it on any random array, of any shape, too. Example:

```python
import numpy as np
import zarr

fout = zarr.open('demo-zarr-3d.zarr', mode='w', shape=(5, 10, 10), dtype=np.float32) 
fout[:] = np.random.random((5, 10, 10)) 
```

If I then do:

```python
napari demo-zarr-3d.zarr
```

in the command line, having installed `ome-zarr`, I get:

napari-zarr

joshmoore:
cc: @will-moore

### DIFF
--- a/ome_zarr.py
+++ b/ome_zarr.py
@@ -122,7 +122,7 @@ class BaseZarr:
 
         elif self.zarray:
             data = da.from_zarr(f"{self.zarr_path}")
-            return [(data, {'channel_axis': 1})]
+            return [(data,)]
 
     def load_omero_metadata(self, assert_channel_count=None):
         """Load OMERO metadata as json and convert for napari"""


will-moore:
Code looks good with respect to #25.
Testing with the code example there I get:

Screenshot 2020-06-03 at 13 37 20

Good to merge.

joshmoore:
Thanks, merging. @jni: note, this defers the decision on whether to handle non ome-zarr arrays at all for a later day.

jni:
Whoo! I love postponed decisions. =D Thanks @joshmoore!

PR #30, opened 2020-07-03
Support opening of masks

will-moore:
Currently supports opening of local zarr masks.

To test:
 - Export masks as described in https://github.com/ome/omero-cli-zarr/pull/11
 - Open image in napari e.g. ```$ omero napari view Image:5514375```
 - In `napari` console, do ```viewer.open('path/to/5514375.zarr/masks/')```

Screenshot 2020-07-03 at 16 51 59

### DIFF
--- a/ome_zarr.py
+++ b/ome_zarr.py
@@ -258,10 +258,11 @@ class BaseZarr:
                 color_dict = mask_attrs.get('color')
                 colors = {int(k):self.to_rgba(v) for (k, v) in color_dict.items()}
             data = da.from_zarr(mask_path)
-            # mask data is 5D (t, c, z, y, x) but each layer in napari is 4D (no C)
-            # NB: Assume we want 'first Channel'
-            data = data[:,0,:,:,:]
-            masks.append((data, {'name': name, 'color': colors}, 'labels'))
+            # Split masks into separate channels, 1 per layer
+            for n in range(data.shape[1]):
+                masks.append((data[:,n,:,:,:],
+                              {'name': name, 'color': colors},
+                              'labels'))
         return masks
 
 
--- a/ome_zarr.py
+++ b/ome_zarr.py
@@ -102,6 +102,9 @@ class BaseZarr:
     def is_ome_zarr(self):
         return self.zgroup and "multiscales" in self.root_attrs
 
+    def is_ome_mask(self):
+        return self.zarr_path.endswith('masks/') and self.get_json('.zgroup')
+
     def get_json(self, subpath):
         raise NotImplementedError("unknown")
 
@@ -109,6 +112,10 @@ class BaseZarr:
         if not self.is_zarr():
             raise Exception(f"not a zarr: {self}")
         return self.reader_function
+
+    def to_rgba(self, v):
+        """Get rgba (0-1) e.g. (1, 0.5, 0, 1) from integer"""
+        return [x/255 for x in v.to_bytes(4, signed=True, byteorder='big')]
 
     def reader_function(self, path: PathLike) -> List[LayerData]:
         """Take a path or list of paths and return a list of LayerData tuples."""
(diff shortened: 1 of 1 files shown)


will-moore:
With that last commit, and with `fsspec` patched as described with @joshmoore's workaround from https://github.com/intake/filesystem_spec/issues/342#issuecomment-654314336

```
napari 'https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001240.zarr/'
```
loads masks along with the Image:

Screenshot 2020-07-06 at 17 18 31

And this works locally (with images exported as above) loading masks with the Image:

```
$ napari 5514375.zarr
```

Screenshot 2020-07-06 at 17 21 24

joshmoore:
In testing against a local array, I was failing to load the masks, @will-moore. I tried unifying the logic and this now works. This, however, removes your directory listing code. Two thoughts:

 - shall I add it back _optionally_? i.e. if the server doesn't allow it it isn't loaded
 - _or_ do we need to provide a file-browser for the the zarr _itself_, i.e. if enabled, you can see all the arrays that are in the zarr and open/overlay those that you want to.

cc: @tlambert03

ome_zarr.py
```diff
@@ -219,6 +246,24 @@ def load_ome_zarr(self):
...
+            mask_path = os.path.join(self.zarr_path, name)
+            mask_attrs = self.get_json(f'{name}/.zattrs')
+            colors = {}
+            if 'color' in mask_attrs:
+                color_dict = mask_attrs.get('color')
+                colors = {int(k):self.to_rgba(v) for (k, v) in color_dict.items()}
+            data = da.from_zarr(mask_path)
+            # mask data is 5D (t, c, z, y, x) but each layer in napari is 4D (no C)
+            # NB: Assume we want 'first Channel'
+            data = data[:,0,:,:,:]
```
manics, line 263: Shouldn't this be handled on the napari side, instead of in this library?

tlambert03: ah interesting.  There does seem to be a limitation here in napari if you have multiple (image) channels along with a mask for each one.  In the near(ish) future, we'll have layer "groups" that would let you associate each labels layer with a given image layer... but for now, you basically have no choice but to have nC x 2 unassociated layers in napari if you have a mask for every image channel.

So for here, doing `data[:, 0, :, :, :]` will make sure you only have a single labels layer, but will obviously omit some data.  So you could return a list of `[(data[:,n,:,:,:],meta,'labels') for n in range(nChannels)]` if you want to (and, until we provide layer groups, give them names that indicate which image layer they go with?)

lastly, singleton dimensions are handled well in napari, so if nC == 1, you wouldn't need to squeeze it.

will-moore:
@joshmoore Your logic from ec13afd for detecting masks relies on https://github.com/ome/omero-cli-zarr/pull/12 ?
I think I was missing that when writing the local logic.

joshmoore:
> [ome/omero-cli-zarr#12](https://github.com/ome/omero-cli-zarr/pull/12) ?

Yes. Happy to roll that back if you need to get this in before 12 lands.

ome_zarr.py
will-moore: Without that line, the masks shape is 5D: `(40, 1, 31, 256, 256)` whereas the Image channels have 4D shape `(40, 31, 256, 256)`, so you get a mismatch of dimensions (masks and Image have independent T sliders):

Screenshot 2020-07-07 at 15 50 13

In 9f7f878 I've split mask channels into layers, as suggested by @tlambert03 - Thanks.

will-moore:
This works now (loading masks):
```
$ napari https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001240.zarr
```

joshmoore:
Merging this as is. I'll updated https://github.com/ome/omero-ms-zarr/pull/55 to list labeled images as zarrays rather than multiscale zgroups and then we can make that move in a follow up. (It'll be the first test of a breaking change!)

Issue #32, opened 2020-07-23
parseurl fails to parse ome data with windows file paths 


NHPatterson:
Hi, I have generated some local zarr stores with ome formatting on windows but ome_zarr.parseurl returns these as a remote store...

The problem is the scheme separates out the windows drive letter as the ```result.scheme```.

```python
from urllib.parse import urlparse
from ome_zarr import LocalZarr, RemoteZarr

def parse_url(path):
    result = urlparse(path)
    if result.scheme in ("", "file"):
        # Strips 'file://' if necessary
        return LocalZarr(result.path)
    else:
        return RemoteZarr(path)

path = "C:/biomic/devdata/test_zarr100.zarr"
zarr_store = parse_url(path)
type(zarr_store)


Out[91]: ome_zarr.RemoteZarr
```

It seems for windows compatibility it may be best to check if the directory exists before trying to parse as a url...here is what  I added to make it work...

```python
from urllib.parse import urlparse
from ome_zarr import LocalZarr, RemoteZarr
from pathlib import Path

def parse_url(path):
    try:
        if Path(path).is_dir():
            return LocalZarr(path)
    except OSError:
        result = urlparse(path)
        if result.scheme in ("", "file"):
            # Strips 'file://' if necessary
            return LocalZarr(result.path)
        else:
            return RemoteZarr(path)

path = "C:/biomic/devdata/test_zarr100.zarr"
zarr_store = parse_url(path)
type(zarr_store)

Out[92]: ome_zarr.LocalZarr
```

Happy to open a pull request.

Thanks!

joshmoore:
Thanks, @NHPatterson. A PR would be great. We will need to get tests running on Windows to keep this from happening again. ~J.

PR #33, opened 2020-07-23
add check for directory for LocalZarr for Windows compatability 

NHPatterson:
See issue : https://github.com/ome/ome-zarr-py/issues/32#issue-664496745
All tests pass, but I had to make the napari reader check for a ```None``` because checking only for```instance.is_zarr()``` errored when the file path returns a ```None```. when parse_url instance returns None as it does with ```"fake.file"``` in the test.


joshmoore:
Re-opening to trigger the new GH Action.

joshmoore:
The GH Action on Windows is now green as opposed to in the mainline. Awesome. Thanks for this, @NHPatterson!

PR #34, opened 2020-07-28
Re-use omero-py travis config for Windows build

joshmoore:
For testing, https://github.com/ome/ome-zarr-py/pull/33

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -1,0 +1,29 @@
+name: Build for Windows
+
+on:
+  pull_request:
+
+jobs:
+  test:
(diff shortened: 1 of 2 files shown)


joshmoore:
Migrating omero-py's build may still be useful, but for Windows it was not sufficient (due to the installation of OpenGL on Windows). Instead a GitHub action was activated on the mainline. See:

 - https://github.com/ome/ome-zarr-py/runs/922312042?check_suite_focus=true
 - https://github.com/ome/ome-zarr-py/commit/4d39c2bd3abedbcf12b9c7ae18b283b0db1e36e7

PR #38, opened 2020-08-03
Add linter.yml

joshmoore:
see: https://github.com/github/super-linter

### DIFF
--- a/.github/workflows/linter.yml
+++ b/.github/workflows/linter.yml
@@ -1,0 +1,52 @@
+---
+###########################
+###########################
+## Linter GitHub Actions ##
+###########################
+###########################
+name: Lint Code Base
+
+#
+# Documentation:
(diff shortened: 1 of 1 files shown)


joshmoore:
Need to find a way to add requirements into the linter docker.

manics:
Could add a pre-commit config for black and flake8 instead? There's an upstream action https://github.com/pre-commit/action

joshmoore:
Certainly no objection. Give it a try. You'd have it run on every push to the branch or only on merge? I'm not exactly sure how the synchronization with upstream is going to work in the latter case.

joshmoore:
> Could add a pre-commit config for black and flake8 instead?

I missed the "instead". Still no objections, but part of my goal was to have something that would lint more than just Python that we could use across all repos _a la_ https://docs.github.com/en/actions/configuring-and-managing-workflows/sharing-workflow-templates-within-your-organization

PR #37, opened 2020-08-03
Logging cleanup

joshmoore:
Migrates nearly all print statements (other than one in `download`)
to `logging` calls. `info` now defaults to INFO and `download` to
DEBUG.

The S3/HTTP connection pool logging is still an issue making any
DEBUG level configuration difficult to read.

PR #39, opened 2020-08-03
Add Precommit for linting

manics:
Adds black autoformatter, flake8, yamllint.

Also adds commented config for mypy since there are some type hints but they need to be fixed.

Best reviewed one commit at a time.

### DIFF
--- a/.github/workflows/precommit.yml
+++ b/.github/workflows/precommit.yml
@@ -1,0 +1,11 @@
+name: Precommit
+on: [push, pull_request]
+jobs:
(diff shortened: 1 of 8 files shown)


joshmoore:
Changes look fine. It probably would have been easier had this been on top of #37 but I'll take care of that. I do have some concern about how to keep the versions of the individual pre-commit hooks up-to-date across all of our repositories. If you see anything about having a single shared version (e.g. `ome-pre-commit@v1.0`) that might be a worthwhile follow up.

PR #42, opened 2020-08-13
mask fault-tolerance (Proposed 0.0.12)

joshmoore:
## Invalid colors ##

e.g.

```
{"image":{"array":"../../0/"},"color":{"true":-2139062144}}
```

see https://github.com/ome/omero-ms-zarr/pull/60

## missing chunks ##

See https://github.com/intake/filesystem_spec/issues/342#issuecomment-654314336

PR #43, opened 2020-08-14
Masks not visible when image opened in napari

will-moore:
I think this will prevent loading of Masks in `napari` until the user makes the labels layer visible.
To test:

```
napari 'https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001240.zarr/'
```

Labels layer is not visible. I think `napari` won't try to load the mask until the layer is shown.

Screenshot 2020-08-14 at 13 30 21

### DIFF
--- a/ome_zarr.py
+++ b/ome_zarr.py
@@ -278,14 +278,12 @@ class BaseZarr:
             data = da.from_zarr(mask_path)
             # Split masks into separate channels, 1 per layer
             for n in range(data.shape[1]):
-                masks.append((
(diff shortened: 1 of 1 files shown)


will-moore:
Ah, I see I was beaten to it!
https://github.com/ome/ome-zarr-py/pull/42/files/6690fc0328ddbf643a8e7393c683c6abd18dfa1e..64551841728c94910c36fd0e3cf0dc798430a889

joshmoore:
Doh. Sorry. :)

PR #44, opened 2020-08-18
Rename 'masks' group to 'labels'

joshmoore:
see: https://github.com/ome/omero-ms-zarr/issues/59

PR #45, opened 2020-08-19
Multiscale labeled images

joshmoore:
ome-zarr now expects a multiscale zgroup rather than a single zarray as the content of a label. Along with the previous round of changes (https://github.com/ome/ome-zarr-py/pull/44#issuecomment-676303785) this means to modify an existing ome-zarr with masks you need to:

 - rename masks/ to labels/
 - run `ome_zarr scale 6001240.py/labels/{0,1}`
 - merge the metadata from labels/0/.zattrs into labels/1/.zattrs
 - add `1` to the `labels/.zattrs` "names" element, optionally removing `0`

(This will be put into a script to be run against s3.embassy.ebi.ac.uk)

Interestingly, with the current changes, loading is still successful without the changes above since the the mask is treated as a raw zarr array.

*Summary of other user changes:*
 - It's now possible to pass the `labels/` directory to napari. The mask is opened visible, and the image is loaded invisible. This is the first usage of the link `{"image": { "array" ...`. See https://github.com/ome/omero-cli-zarr/pull/19 . Future work will likely make it possible to point at the label itself and have the images loaded, too.
 - New `ome_zarr create` command for creating sample data. Note: colors are not working correctly yet.
 - New `ome_zarr scale` command
 - `ome_zarr info` now displays different output.
 - `ome_zarr download` now downloads recursively.

*Summary of other developer changes:*
 - Split main script into napari.py, cli.py, utils.py, reader.py (beginning of refactoring for code reuse; multiple napari plugins may also be a good path forward).
 - Splitting of BaseZarr into BaseZarrLocation and Layer/Spec types (significantly simplifies different starting points)
 - Enforce mypy in all method definitions (aided with refactoring).


ome_zarr/reader.py
```diff
@@ -264,41 +265,31 @@ def load_ome_zarr(self):
...
+    def load_ome_label_metadata(self, name: str) -> Dict:
+        # Metadata: TODO move to a class
+        label_attrs = self.get_json(f"{name}/.zattrs")
+        colors: Dict[Union[int, bool], List[float]] = {}
+        color_dict = label_attrs.get("color", {})
+        if color_dict:
+            for k, v in color_dict.items():
+                try:
+                    if k in ("true", "false"):
+                        k = bool(k)
```
mtbc, line 277: @manics noticed that,
```python
[bool(x) for x in ["true", "false"]] == [True, True]
```

joshmoore: Pushed.

ome_zarr/data.py
```diff
@@ -0,0 +1,166 @@
...
+    pyramid, labels = method()
+
+    store = zarr.DirectoryStore(zarr_directory)
+    grp = zarr.group(store)
+    write_multiscale(pyramid, grp)
+
+    if pyramid[0].shape[CHANNEL_DIMENSION] == 1:
+        image_data = {
+            "channels": [{"window": {"start": 0, "end": 1}}],
+            "rdefs": {"model": "grayscale"},
```
mtbc, line 121: In the OMERO metadata JSON I think we previously had "greyscale" matching `RenderingModel`. I don't know if this matters or what it should be.

ome_zarr/io.py
```diff
@@ -0,0 +1,150 @@
...
+    """
+
+    def __init__(self, path: str) -> None:
+        self.zarr_path: str = path.endswith("/") and path or f"{path}/"
+        self.zarray: JSONDict = self.get_json(".zarray")
+        self.zgroup: JSONDict = self.get_json(".zgroup")
+        self.__metadata: JSONDict = {}
+        self.__exists: bool = True
+        if self.zgroup:
+            self.__metadata = self.get_json(".zattrs")
```
mtbc, line 37: not `.zgroup`?

joshmoore: No. It's wanting the flexible metadata stuffs for multiscales, etc.

ome_zarr/data.py
joshmoore: Happy to defer to @will-moore, but I agree at least the three zarr related repos should use uniform spelling.

ome_zarr/reader.py
```diff
@@ -0,0 +1,335 @@
...
+            LOGGER.info("root_attr: %s", k)
+            LOGGER.debug(v)
+
+    def lookup(self, key: str, default: Any) -> Any:
+        return self.zarr.root_attrs.get(key, default)
+
+
+class Labels(Spec):
+    """Relatively small specification for the well-known "labels" group which only
+    contains the name of subgroups which should be loaded an labeled images."""
```
mtbc, line 105:
```suggestion
    contains the name of subgroups which should be loaded on labeled images."""
```
?

ome_zarr/scale.py
```diff
@@ -0,0 +1,206 @@
...
+            self.__assert_values(pyramid)
+
+        grp = self.__create_group(store, base, pyramid)
+
+        if self.copy_metadata:
+            print(f"copying attribute keys: {list(base.attrs.keys())}")
+            grp.attrs.update(base.attrs)
+
+    def __check_store(self, output_directory: str) -> MutableMapping:
+        """Return a Zarr store if it doesn't not already exist."""
```
mtbc, line 86:
```suggestion
        """Return a Zarr store if it doesn't already exist."""
```

ome_zarr/scale.py
```diff
@@ -0,0 +1,206 @@
...
+        return list(reversed(rv))
+
+    #
+    # Helpers
+    #
+
+    def _by_plane(
+        self, base: np.ndarray, func: Callable[[np.ndarray, int, int], np.ndarray],
+    ) -> np.ndarray:
+        """Loop over 3 of the 5 dimensions of and apply the func transform."""
```
mtbc, line 188: of?

ome_zarr/io.py
mtbc: :+1: Okay, just wanted to make sure given the avoidance of "or".

will-moore:
Testing on merge-ci.
First, add masks using `omero napari...`
Imported an image https://merge-ci.openmicroscopy.org/web/webclient/?show=image-139615 and opened in napari:
```
$ omero napari view Image:139615
```
Drew labels and click "Save ROIs to OMERO"

Export image with masks.
Using omero-cli-zarr at the last commit before https://github.com/ome/omero-cli-zarr/pull/22 (renaming to labels)  5fe97fe.

```
$ omero zarr export Image:139615
Using session for user-3@10.0.48.9:4064. Idle timeout: 10 min. Current group: read-only-1
loading plane c:0, t:0, z:0
loading plane c:1, t:0, z:0
loading plane c:2, t:0, z:0
Created 139615.zarr

$ omero zarr masks Image:139615
Using session for user-3@10.0.48.9:4064. Idle timeout: 10 min. Current group: read-only-1
Export Masks on Image: plate1_1_013 [Well 1, Field 1 (Spot 1)]_merged_z01_t01.png
Found 10 mask shapes in 10 ROIs
...
Created 139615.zarr/masks/0

```

Using `ome-zarr-py` before renaming to labels, I can open image with masks in napari:
```
git checkout 208952c
napari 139615.zarr/
```

To test this PR....

 - First renamed `/masks/` to `/labels/`
 - Also updated `lables/.zattrs
```
{
    "labels": [
        "0"
    ]
}
```
- Using the master of ome-zarr-py, this image opened with labels as before in napari.
- Checked-out this branch and needed to re-install: `$ pip install -e .` to fix `ModuleNotFoundError: No module named 'ome_zarr_cli'`
- `$ ome_zarr scale 139615.zarr/labels/0/`
`ome_zarr scale: error: the following arguments are required: output_directory`
- Repeated:
- `ome_zarr scale 139615.zarr/labels/0/ 139615.zarr/scaled/`
- This created dirs 1-4 in that location.
- Trying to get the new dirs `1-4` as siblings to the original `0` with: 
`$ ome_zarr scale 139615.zarr/labels/0/ 139615.zarr/labels/`
Didn't seem to do anything - no new dirs appeared.
- Copied /scaled/{1-4} into /labels/ alongside the 0 dir.
- Updated `/labels/.zattrs/` to replace "0" with 1-4: ```
{
    "labels": [
        "1",
        "2",
        "3",
        "4",
    ]
}
```
- copied the /labels/0/.zattrs content and duplicated it to create /labels/{1-4}/.zattrs which were previously empty.
Trying to open in napari... 
```
$ napari 139615.zarr/
14:20:49 ERROR PluginError: Error in plugin 'ome_zarr', hook 'napari_get_reader'
  Cause was: JSONDecodeError('Expecting value: line 7 column 5 (char 74)')
    in file: /Users/wmoore/opt/miniconda3/envs/omero-napari/lib/python3.7/json/decoder.py
    at line: 355
     author: The Open Microscopy Team
    package: ome-zarr
        url: https://github.com/ome/ome-zarr-py
    version: 0.0.13.dev0

14:20:49 ERROR (1) error occured in plugins: 'ome_zarr'. See full error logs in "Plugins → Plugin Errors..."
  File "/Users/wmoore/opt/miniconda3/envs/omero-napari/bin/napari", line 8, in <module>
    sys.exit(main())
  File "/Users/wmoore/opt/miniconda3/envs/omero-napari/lib/python3.7/site-packages/napari/__main__.py", line 171, in main
    **kwargs,
  File "/Users/wmoore/opt/miniconda3/envs/omero-napari/lib/python3.7/site-packages/napari/view_layers.py", line 208, in view_path
    path=path, stack=stack, plugin=plugin, layer_type=layer_type, **kwargs
  File "/Users/wmoore/opt/miniconda3/envs/omero-napari/lib/python3.7/site-packages/napari/components/add_layers_mixin.py", line 859, in open
    _path, kwargs, plugin=plugin, layer_type=layer_type
  File "/Users/wmoore/opt/miniconda3/envs/omero-napari/lib/python3.7/site-packages/napari/components/add_layers_mixin.py", line 925, in _add_layers_with_plugins
    new = self._add_layer_from_data(*_data)
  File "/Users/wmoore/opt/miniconda3/envs/omero-napari/lib/python3.7/site-packages/napari/components/add_layers_mixin.py", line 997, in _add_layer_from_data
    layer = add_method(data, **(meta or {}))
  File "/Users/wmoore/opt/miniconda3/envs/omero-napari/lib/python3.7/site-packages/napari/components/add_layers_mixin.py", line 230, in add_image
    return self.add_layer(layers.Image(data, **kwargs))
  File "/Users/wmoore/opt/miniconda3/envs/omero-napari/lib/python3.7/site-packages/napari/layers/image/image.py", line 159, in __init__
    multiscale, data = guess_multiscale(data)
  File "/Users/wmoore/opt/miniconda3/envs/omero-napari/lib/python3.7/site-packages/napari/layers/image/_image_utils.py", line 52, in guess_multiscale
    shapes = [d.shape for d in data]
  File "/Users/wmoore/opt/miniconda3/envs/omero-napari/lib/python3.7/site-packages/napari/layers/image/_image_utils.py", line 52, in <listcomp>
    shapes = [d.shape for d in data]
AttributeError: 'list' object has no attribute 'shape'
```

Not sure where I went wrong. Will try to figure it out...

File: ome_zarr/scale.py (hunk 6 above)
Comment (joshmoore) [reply ome_zarr/scale.py:188]:
```suggestion
        """Loop over 3 of the 5 dimensions and apply the func transform."""
```

will-moore:
Sorry - can't work out how to get masks working.
I tried deleting all the `labels/` dir, to see if I could get my 3-channel, single-plane image to open in napari,
but in `napari.py` `def transform` at `shape = data[0].shape` my shape is `(3, 1, 1024, 1344)` so this check fails
```
if shape[CHANNEL_DIMENSION] > 1:
    metadata["channel_axis"] = CHANNEL_DIMENSION
```
Then napari fails with 
`TypeError: Received sequence for argument 'colormap', did you mean to specify a 'channel_axis'? `

If I hard-code the channel_axis back in here, the 3-channel image opens in napari.
But I can't get the labels to show up, when I add back the '/labels/` dir...

joshmoore:
Hmmm.... thanks for the testing, @will-moore. I did push a few more fixes yesterday which you might not have yet, but I never saw these particular failures. Which version of napari are you working from? An RC for 0.3.7 came out yesterday that I've been working from (but you need one commit on top of it from origin/master to fix `-vvv` output) You might also want to try the `environment.yml` I've added for setting up a conda environment.

will-moore:
I was using `napari==0.3.6`. Now I've updated to latest `origin/master`
```
* 1b761a0 (HEAD, origin/master, origin/HEAD) Add SELECTED color-mode for individual label visibility (#1555)
* ad149be fix ipython raise behavior (#1595)
* c71644a Restore multiple command-line arguments (#1597)
* ebdb9d5 Minor updates to custom mouse functionality examples (#1592)
* 26540ce (tag: v0.3.7rc0) 0.3.7 release notes (#1589)
```

Also updated to latest changes in `multiscale` branches of ome-zarr-py and omero-cli-zarr:
```
$ pip freeze | grep ome
-e git+git@github.com:tlambert03/napari-omero.git@d4add2f928af5a42bec43c14a5b57e912e7ee393#egg=napari_omero
-e git+git@github.com:ome/ome-zarr-py.git@69288fc47126973f7f3c199928fb38c81d0ea249#egg=ome_zarr
-e git+git@github.com:ome/omero-cli-zarr.git@32bf5de78dad0a8018e9968c370f040e05f93811#egg=omero_cli_zarr
omero-napari==0.1.0
-e git+git@github.com:ome/omero-py.git@b7e29643d1ad7cabdbef0eba81a500ab02c9c35e#egg=omero_py
omero-rois==0.3.0
```

Repeating steps from above with https://merge-ci.openmicroscopy.org/web/webclient/?show=image-139615

```
$ omero zarr export Image:139615
$ omero zarr masks Image:139615
```
This gave me 2 resolutions of labels:
```
$ ls 139615.zarr/labels/0/
0	1
```
Tried to open image with labels in `napari`
```
$ napari 139615.zarr
```
Saw same error as above:
Screenshot 2020-09-01 at 12 47 44

ome_zarr/napari.py
```diff
@@ -0,0 +1,82 @@
...
+            if data is None or len(data) < 1:
+                LOGGER.debug(f"skipping non-data {layer}")
+            else:
+                LOGGER.debug(f"transforming {layer}")
+                shape = data[0].shape
+
+                layer_type: str = "image"
+                if layer.load(Label):
+                    layer_type = "labels"
+                    if "colormaps" in metadata:
```
will-moore, line 63: I think the key here is `colormap` not `colormaps`. Same for below.

ome_zarr/napari.py
```diff
@@ -0,0 +1,82 @@
...
+        results: List[LayerData] = list()
+
+        for layer in layers:
+            data: List[Any] = layer.data
+            metadata: Dict[str, Any] = layer.metadata
+            if data is None or len(data) < 1:
+                LOGGER.debug(f"skipping non-data {layer}")
+            else:
+                LOGGER.debug(f"transforming {layer}")
+                shape = data[0].shape
```
will-moore, line 58: The `shape` I have here is `(3, 1, 1024, 1344)` so the `channel_axis` is not added below.

will-moore:
If I convert `colormaps` to `colormap` in `napari.py` (see my comment above) then I can open, but I see a slider for the 3 channels instead of them being separate layers:


Screenshot 2020-09-01 at 13 10 51

ome_zarr/napari.py
joshmoore: Cheers. I'll push a commit.

joshmoore:
> but I see a slider for the 3

Looks like this is caused by small images. When there's only one array to the pyramid, something is dropping the outer `list()`. I'll look into it.

will-moore:
So, I didn't test the upgrade steps today. `ome_zarr scale 139615.zarr/labels/0/ 139615.zarr/scaled/` etc.
Does that still need testing again?

ome_zarr/io.py
```diff
@@ -0,0 +1,150 @@
...
+        if self.zarray:
+            suffix += " [zarray]"
+        return f"{self.zarr_path}{suffix}"
+
+    def exists(self) -> bool:
+        """Return true if zgroup or zarray metadata exists."""
+        return self.__exists
+
+    def is_zarr(self) -> Optional[JSONDict]:
+        """Return true if either zarray or zgroup metadata exists."""
```
mtbc, line 57: I don't quite get what's being communicated by "either" here but not for the previous.

joshmoore: A node in the Zarr hierarchy must have one or the other.

mtbc: Aha, but that's not true for `exists`, it could be both? Okay.

ome_zarr/napari.py
joshmoore: Pushed a fix.

will-moore:
Testing as above with latest commits from here and `omero-cli-zarr` is working:

```
$ omero zarr export Image:139615
$ omero zarr masks Image:139615
$ napari 139615.zarr
```

Screenshot 2020-09-02 at 13 26 10

I noticed the labels layer is visible initially. Is that a change of behaviour?
Also, I wonder if we want to rename `masks -> labels` for:
```
$ omero zarr masks Image:139615
```
?

will-moore:
I tried downloading:
```
$ ome_zarr download https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001240.zarr/
WARNING:ome_zarr.io:unreachable: https:/s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001240.zarr/labels/.zarray -- details logged at debug
WARNING:ome_zarr.io:unreachable: https:/s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001240.zarr/labels/.zgroup -- details logged at debug
ERROR:ome_zarr.cli:. already exists!
```
I guess the s3 data hasn't been updated from masks -> labels yet, but even if labels are missing, should the image data itself still download OK?
e.g. When I try `napari 'https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001240.zarr/'` even though there are no labels, this opens the image fine.

joshmoore:
> I noticed the labels layer is visible initially. Is that a change of behaviour?

Yeah, I noticed that as well. One of my bug fixes must have reverted. I assume it's not overly critical, but I'll look into it.

> I tried downloading ... I guess the s3 data hasn't been updated from masks -> labels yet

Definitely. `omero zarr masks` is the only source of test data at the moment. Once these PRs go in, I can update the public data. Or we hold off for the color changes.

will-moore:
My point was not so much that the 'masks -> labels` updates need to happen on s3, but that any image fails to download if it doesn't have any labels.
Or, maybe it's not the lack of labels that is breaking download, but something else.
Trying another ID from https://downloads.openmicroscopy.org/presentations/2020/Dundee/Workshops/NGFF/zarr_diagram/ which I don't think has any masks/labels

```
$ ome_zarr download https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/179706.zarr/
WARNING:ome_zarr.io:unreachable: https:/s3.embassy.ebi.ac.uk/idr/zarr/v0.1/179706.zarr/labels/.zarray -- details logged at debug
WARNING:ome_zarr.io:unreachable: https:/s3.embassy.ebi.ac.uk/idr/zarr/v0.1/179706.zarr/labels/.zgroup -- details logged at debug
ERROR:ome_zarr.cli:. already exists!
```

will-moore:
Ah, this was a lack of output dir.
This works:
```
$ ome_zarr download https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/179706.zarr/ --output test
```

ome_zarr/io.py
joshmoore: Pushed a removal of `is_zarr()` in favor of `exists()`

ome_zarr/reader.py
```diff
@@ -47,22 +50,64 @@ def __init__(
...
+        self,
+        zarr: BaseZarrLocation,
+        prepend: bool = False,
+        visibility: Optional[bool] = None,
+    ) -> "Optional[Node]":
+        """Create a child node if this location has not yet been seen.
+
+        Returns None if the node has already been processed.
+
+        By setting prepend, the addition will be considered as higher priority
```
mtbc, line 94:
```suggestion
        By setting prepend, the addition will be considered to be a higher priority
```

ome_zarr/reader.py
```diff
@@ -47,22 +50,64 @@ def __init__(
...
+        node which will likely be turned into a lower layer for display.
+
+        By unsetting visible, the node (and in turn the layer) can be
+        deactivated for initial display. By default, the value of the current
```
mtbc, line 98: Slightly confused: "by default" suggests a means of override but I am not seeing what that is.

will-moore:
The `labels` layer is not visible when opened in napari now. 👍 

For `$ ome_zarr download https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/179706.zarr/`, should `--output` be a required parameter (since it will fail with the default output dir of `.`)?

ome_zarr/reader.py
```diff
@@ -0,0 +1,379 @@
...
+        self.__visible = visibility
+
+        # Likely to be updated by specs
+        self.metadata: JSONDict = dict()
+        self.data: List[da.core.Array] = list()
+        self.specs: List[Spec] = []
+        self.pre_nodes: List[Node] = []
+        self.post_nodes: List[Node] = []
+
+        # TODO: this should be some form of plugin infra over subclasses
```
sbesson, line 43: I assume another todo might be to add some validation for incompatible specs i.e. I assume a node cannot be both a `Label` and `Labels`

ome_zarr/data.py
```diff
@@ -0,0 +1,166 @@
...
+
+
+def write_multiscale(pyramid: List, group: zarr.Group) -> None:
+    """Write a pyramid with multiscale metadata to disk."""
+    paths = []
+    for path, dataset in enumerate(pyramid):
+        group.create_dataset(str(path), data=pyramid[path])
+        paths.append({"path": str(path)})
+
+    multiscales = [{"version": "0.1", "datasets": paths}]
```
sbesson, line 102: Is this where the version is supposed to be ? In which case we might need to update https://github.com/ome/omero-ms-zarr/pull/70 which suggests it is a key under the `OMERO` spec.

ome_zarr/data.py
```diff
@@ -0,0 +1,166 @@
...
+    group.attrs["multiscales"] = multiscales
```
sbesson, line 103: This made me wonder where this logic should live in future improvements of this API. The two immediate options I can think of would be:

- have each `Spec` subclass implementing its own serialization method `Multiscales.write(paths)`
- have an `ome_zarr.writer` module mirroring the reader one
- have a

sbesson (commented):
:+1: for the overall refactoring of this class into more tractable and testable modules. Are we effectively approaching an `0.1.0`?

ome_zarr/data.py
joshmoore: Yeah, there is definitely more refactoring and encapsulating to be done: I just had to draw the line somewhere. Hopefully at this point it's fairly straightforward to see where something should live (in this case I'd say the `Spec` subclass and if we want to then wrap that with a Writer we can).

ome_zarr/data.py
joshmoore: There are multiple version elements, one per Spec object (at least currently). This is the multiscales version.

ome_zarr/reader.py
joshmoore: If the `Labels` pointed to itself with `.` it might just work...

ome_zarr/data.py
joshmoore: Yup. OMERO has `greyscale` everywhere. napari `grayscale`. I'll update to use the former and if we need to, we can transform in `ome_zarr.napari`.

NB: ome-model primarily uses "e", "a" appears only twice in docs. Bio-Formats leans more towards "a" including the CLI argument.

joshmoore:
> should `--output` be a required parameter (since it will fail with the default output dir of `.`)?

Ah thanks. I've fixed this to work again without the flag.

ome_zarr/reader.py
joshmoore: Passing a value would override the None. I'm pushing a commit to rework the wording. Let me know if you think it's better.

ome_zarr/data.py
sbesson: All fine with me. I think we should make this explicit in the `spec.md`. I can open a follow-up PR after https://github.com/ome/omero-ms-zarr/pull/70 unless this is already included in your follow-up commits.

ome_zarr/data.py
sbesson: :+1: not a blocker. My initial feeling would also be the `Spec` subclass which might lead us to moving it to a separate module e.g. `ome_zarr.specifications` that could be consumed both by readers and writers

joshmoore:
Non-Windows tasks are all now passing. It's going to be fiddly to get Windows passing, so I'd like to do that in a new PR.

Any objections to merging?

mtbc:
GFI. I think any fixes will be much smaller increments than this.

PR #46, opened 2020-09-04
Implement new color suggestion

joshmoore:
see: https://github.com/ome/omero-ms-zarr/pull/71

### DIFF
--- a/ome_zarr/reader.py
+++ b/ome_zarr/reader.py
@@ -7,7 +7,6 @@ from typing import Any, Dict, Iterator, List, Optional, Type, Union, cast
 import dask.array as da
 from vispy.color import Colormap
 
-from .conversions import int_to_rgba
 from .io import BaseZarrLocation
 from .types import JSONDict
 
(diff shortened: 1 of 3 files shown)


joshmoore:
Fixing the build will require #47 to be merged.

ome_zarr/reader.py
```diff
@@ -182,13 +181,14 @@ class Label(Spec):
     @staticmethod
     def matches(zarr: BaseZarrLocation) -> bool:
         """If label-specific metadata is present, then return true."""
-        # FIXME: this should be the "label" metadata soon
-        return bool("colors" in zarr.root_attrs or "image" in zarr.root_attrs)
+        return bool("image-label" in zarr.root_attrs)
```
sbesson, line 184: Should the `Spec` be renamed to `ImageLabel`?

joshmoore:
Re-opening to trigger a rebuild with #47

ome_zarr/reader.py
joshmoore: Interesting. No immediate objections, but to some extent I think we're "safer" in ome-zarr-py in the sense that it's clearly our namespace whereas in the json someone else could propose a "label" key.

joshmoore:
_grrr_... even the repeated clones are now failing.

joshmoore:
_whew_ Green on a manual re-run of the Windows workflow.

joshmoore:
Merging along with ome/omero-ms-zarr#71 in order to enable a basic workflow.

PR #47, opened 2020-09-04
Cross-platform path handling

joshmoore:
Removes the `zarr_path` variable from BaseZarrLocation, using a
pathlib.Path field in LocalZarrLocation and a URL string in
RemoteZarrLocation to properly handle the creation of subpaths across
multiple platforms. In general, use of os.path and especially
os.path.join should be avoided in favor of delegating to methods on the
zarr location objects.

### DIFF
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -28,6 +28,10 @@ jobs:
           python-version: ${{ matrix.python-version }}
 
       - name: Clone gl-ci-helpers
+        env:
+          GIT_CURL_VERBOSE: 1
+          GIT_TRACE: 1
+          GIT_TRACE_PACKET: 1
         run: git clone --depth 1 git://github.com/vtkiorg/gl-ci-helpers.git
 
       - name: Install opengl
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -32,7 +32,9 @@ jobs:
           GIT_CURL_VERBOSE: 1
           GIT_TRACE: 1
           GIT_TRACE_PACKET: 1
-        run: git clone --depth 1 git://github.com/vtkiorg/gl-ci-helpers.git
+        run: >
+            git clone --depth 1 git://github.com/vtkiorg/gl-ci-helpers.git ||
+            git clone --depth 1 git://github.com/vtkiorg/gl-ci-helpers.git
 
       - name: Install opengl
         shell: powershell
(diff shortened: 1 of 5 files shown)


joshmoore:
_Whew_. Finally. So a summary of steps taken and lessons learned to get the build passing:
 - extra `git clone` steps (at least of gl-ci-helpers on Windows) fail regularly.
 - `action/checkout` can't be used separately -- it will delete the previous check.
 - Qt on Windows and Ubuntu are going to take more work.
But ultimately, I'm _fairly_ confident that GH actions are now properly testing the the cross-platform support.

tlambert03:
> Qt on Windows and Ubuntu are going to take more work.
But ultimately, I'm fairly confident that GH actions are now properly testing the the cross-platform support.

not sure if it will be useful to you... but I have a branch of napari working with GH actions on all platforms (there may of course be other OMERO specific stuff that still need to be figured out): https://github.com/tlambert03/napari/blob/feature/github-tests/.github/workflows/tests.yml

joshmoore:
Thanks, @tlambert03. I'll review the diff and see if I can get the Qt working. :+1:

.github/workflows/windows.yml
```diff
@@ -28,7 +28,13 @@ jobs:
...
 
       - name: Clone gl-ci-helpers
-        run: git clone --depth 1 git://github.com/vtkiorg/gl-ci-helpers.git
+        env:
+          GIT_CURL_VERBOSE: 1
+          GIT_TRACE: 1
+          GIT_TRACE_PACKET: 1
+        run: >
+            git clone --depth 1 git://github.com/vtkiorg/gl-ci-helpers.git ||
+            git clone --depth 1 git://github.com/vtkiorg/gl-ci-helpers.git
```
mtbc, line 37: It makes sense that these two lines clone the same thing right?

joshmoore: Yes, the `||` is the important bit. GHA doesn't currently have "retry" functionality.

sbesson (approved):
I have close to little experience with `pathlib` but from a quick reading, it looks like it fulfills  the Windows/Unix path support and has the benefit of being a core Python library.

Tests are passing, the code rewriting looks generally sensible although I haven't looked at the details.

A side question is whether we would foresee switching to `pathlib` in `omero-py` as a replacement for the `path.py` fork?

joshmoore:
> A side question is whether we would foresee switching to `pathlib` in `omero-py` as a replacement for the `path.py` fork?

I could definitely see that. Happy for this to be something of a test run.

PR #50, opened 2020-09-18
Include packages in setup.py

perlman:
I was unable to use `ome_zarr` when installating through PiPy (`pip install ome_zarr`) or locally `pip install .`:

```bash
% ome_zarr 
  File "/Users/eric/nobackup/jax/ENV/bin/ome_zarr", line 5, in <module>
    from ome_zarr.cli import main
ModuleNotFoundError: No module named 'ome_zarr'
```

Developer mode works fine (`pip install -e .`) as this creates the required link (`ENV/lib/python3.8/site-packages/ome-zarr.egg-link`.)

Reproducible with python3.8 under ubuntu 20.04 (docker) and MacOS with a clean virtualenv.

One work around is to use `packages`. I don't know why `py_modules` is insufficient.

### DIFF
--- a/.bumpversion.cfg
+++ b/.bumpversion.cfg
@@ -4,14 +4,14 @@ commit = True
 tag = True
 sign_tags = True
 parse = (?P<major>\d+)\.(?P<minor>\d+)\.(?P<patch>\d+)(\.(?P<release>[a-z]+)(?P<build>\d+))?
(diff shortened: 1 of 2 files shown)


joshmoore:
Thanks, @perlman. I'll roll out a patch release once the above is green.

Issue #52, opened 2020-10-12
.zattrs and .zgroup should be downloaded first, not last


jni:
The download code leaves downloading of the .zattrs and .zgroup files for last. However, because partially downloaded zarrs are valid thanks to the fact that zarr is a sparse format, they should be downloaded *first* so that the user is left with a valid zarr even if the download is interrupted.

Reference code here:

https://github.com/ome/ome-zarr-py/blob/db60b8272e0fe005920f8a296d4828b7a32e663e/ome_zarr/utils.py#L71-L94

PR #53, opened 2020-10-16
Download .z files first (close gh-52)

joshmoore:
Implements the suggestion from @jni to download the smaller
metadata files before downloading the binary data.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: python
 jobs:
   include:
     - stage: deploy
-      python: "3.7"
+      python: "3.6"
       if: tag IS present
       install: skip
       script: skip
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ language: python
 jobs:
   include:
     - stage: deploy
-      python: "3.6"
+      python: "3.7"
       if: tag IS present
       install: skip
       script: skip
--- a/.github/workflows/posix.yml
+++ b/.github/workflows/posix.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         platform: [ubuntu-latest, macos-latest]
-        python-version: [3.7, 3.8]
+        python-version: [3.6, 3.7, 3.8]
 
     steps:
       - name: Checkout
--- a/.github/workflows/posix.yml
+++ b/.github/workflows/posix.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         platform: [ubuntu-latest, macos-latest]
-        python-version: [3.6, 3.7, 3.8]
+        python-version: [3.7, 3.8]
 
     steps:
       - name: Checkout
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -12,7 +12,7 @@ jobs:
       fail-fast: false
       matrix:
         platform: [windows-latest]
-        python-version: [3.7, 3.8]
+        python-version: [3.6, 3.7, 3.8]
 
 
     steps:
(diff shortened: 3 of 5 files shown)


joshmoore:
@sbesson @manics : notice the dropping of Python 3.6. Minimally tifffile and napari have dropped support.

.travis.yml
```diff
@@ -5,7 +5,7 @@ language: python
 jobs:
   include:
     - stage: deploy
-      python: "3.6"
+      python: "3.7"
```
sbesson, line 8: As a RFE, this should probably be migrated to GitHub actions similarly to https://github.com/ome/omero-py/blob/cda0978e1041e1d4ef7927d7232a15eaf8fab2ed/.github/workflows/workflow.yml#L45-L62

sbesson:
> @sbesson @manics : notice the dropping of Python 3.6. Minimally tifffile and napari have dropped support.

Slightly inconvenient but certainly not a surprise given the choices made by these communities. Moving forward, if Python versions continue to be dropped (too?) quickly, there might be an argument for extracting a base library with wider Python 3.x support and a napari component which could be tied to the upstream Python requirements.

jni:
> Moving forward, if Python versions continue to be dropped (too?) quickly

I'm not sure if you're aware of [NEP-29](https://numpy.org/neps/nep-0029-deprecation_policy.html), which states 3.5y of support for Python minor versions. So there is a policy on which you can base this decision — if 3.5y is too fast.

I am, however, confused about why the CI is failing! Certainly the intent of the policy is *not* to force the entire ecosystem to upgrade at this cadence. Installing tifffile or napari on 3.6 should simply give you the last version to support 3.6 — and of course there are recent versions of both that do that! So there is something amiss with how pip or conda are handling things here...

joshmoore:
I can reproduce with the entire `environment.yml` followed by the `pip install --upgrade ... .[napari]` command. I haven't been able to reproduce with just one library yet.

Not sure if it's related but on creating the original environment I see:

```
Preparing transaction: done
Verifying transaction: \
SafetyError: The package for napari located at /usr/local/anaconda3/pkgs/napari-0.3.6-py_0
appears to be corrupted. The path 'site-packages/napari/utils/colormaps/colormaps.py'
has an incorrect size.
  reported size: 13610 bytes
  actual size: 13643 bytes


done
Executing transaction: done
```

sbesson:
@jni Thanks for the NEP-29 link and apologies if my comment about the dropping policy came across the wrong way. The OME version support policy has been historically driven by the versions shipped with the default packages of the supported operating systems.

@joshmoore thanks for the testing the conda workflow. I separately tested the installation workflow without conda. On a base CentOS 7 with Python 3.6, the following commands successfully install `napari` with the last Python 3.6 compatible version of `tifffile`;

```
yum install python3 python3-devel gcc
python3 -mvenv venv && venv/bin/pip install napari
```

joshmoore:
After installing the environment (before any calls to `pip install --upgrade`) I see this:

```
pipdeptree -p tifffile -r
...
tifffile==2020.10.1
  - napari==0.3.6 [requires: tifffile>=2020.2.16]
```

jni:
@joshmoore so is it maybe that tifffile does not declare Python 3.6 incompatibility in its conda package? Maybe the fix is to limit the tifffile version manually in the environment.yml?

jni:
@sbesson no worries! :heart: We are big fans of your work. I just don't want napari to get a reputation for moving fast and breaking people's stuff! We will always aim to follow community standards.

joshmoore:
Green

joshmoore:
@jni and that means it works for you and you can start accessing the dataset earlier during a transnational download? :smile:

jni:
Sure, if you merge + release! =P

The motivation for writing this issue was a download that failed towards the end (later timepoints) of the highest resolution download, so it was not a big deal (hooray for zarr sparse support!) — *if* we had the .z files! Which we manually downloaded separately.

PR #55, opened 2020-10-20
Add scikit-image dependency to setup.py

jni:
Installing napari and ome-zarr results in a broken plugin, because napari does not depend on scikit-image.


joshmoore:
Ta. I'll roll this into the impending release.

sbesson:
Thanks @jni I also bumped into this issue while consuming `ome-zarr`. I will test things work with the latest release without having to redeclare `scikit-image` in downstream modules.

PR #57, opened 2020-10-22
Plates

will-moore:
Test sets:

 - https://minio-dev.openmicroscopy.org/idr/idr0033-rohban-pathways/41744_illum_corrected/pr_43/5966.zarr (big plate, 9 fields per well)
 - https://minio-dev.openmicroscopy.org/idr/idr0002-heriche-condensation/plate1_1_013/pr_43/422.zarr (multi-T)
 - https://minio-dev.openmicroscopy.org/idr/idr0004-thorpe-rad52/P101/pr_43/1751.zarr (sparse plate)
 - https://minio-dev.openmicroscopy.org/idr/idr0001-graml-sysgro/JL_120731_S6A/pr_45/2551.zarr (multi-Z, multiple acquisitions)

To test: Open via e.g:
```
$ napari https://minio-dev.openmicroscopy.org/idr/idr0002-heriche-condensation/plate1_1_013/pr_43/422.zarr 
```
 - Plates should be displayed in a grid with Well labels overlaid.
 - You can view an individual Well by appending e.g. `/A/1/` to the URL (actually, this will only work with the last plate above. The others were generated with an acquisition level in the hierarchy and need to add e.g. `0/A/1` or `Run 422/A/1` for idr0002.

Screenshot 2020-10-23 at 07 18 04

PR #59, opened 2020-10-29
Cap napari extra requirement depending on the Python version

sbesson:
See https://github.com/ome/ome-zarr-py/pull/57#issuecomment-718708701

This attempts at solving the compatibility issue by adding environment markers to `extra_require`.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -54,6 +54,9 @@ setup(
         "napari.plugin": ["ome_zarr = ome_zarr.napari"],
         "pytest11": ["napari-conftest = napari.conftest"],
     },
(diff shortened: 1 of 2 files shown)


will-moore:
👍  Great - once this is merged, my plates PR should go green again.

sbesson:
Only caveat is I could not manage to update the `test_napari.test_viewer` topass on OSX against all versions of `napari/Python`. That being said, I am happy to only run this on Python 3.7+ to match the napari roadmap.

joshmoore (approved):
LGTM. Eventually, we'll probably even need to strip the "3.7" bits. Onward and upwards.

Issue #60, opened 2020-10-31
Add class annotations and/or other metadata properties to labels


DragaDoncila:
Currently the labels spec supports the declaration of a `label-value` and its associated `color`.

Commonly, label values have other associated information including the most obvious, the class name. napari also supports display of label properties, so this would be a nice additional feature for the reader plugin.

I think the critical requirements for these properties should be:
- Supporting an easy mapping between a given property and the `label-value`/s it is associated with
- Enforcing as few rules as possible on what kinds of properties can be accepted
- Supporting an arbitrary number of properties 

There are three ways I can see the spec supporting these additional properties:

1. Arbitrary number of lists of max length n for a label image containing n label values, each corresponding to a property. The index in the list corresponds to the integer `label-value` e.g.

```
    "image-label": {
        "version": "0.1",
        "colors": [
            {
                "label-value": 1,
                "rgba": [
                    255,
                    100,
                    100,
                    255
                ]
            },
            {
                "label-value": 2,
                "rgba": [
                    0,
                    40,
                    200,
                    255
                ]
            },
            {
                "label-value": 3,
                "rgba": [
                    148,
                    50,
                    165,
                    255
                ]
            }
        ],
        "properties": [
            {
                "class": [
                    "Urban",
                    "Water",
                    "Agriculture"
                ],
                "area_m2":
                [
                    "400",
                    "1532",
                    "590"
                ]
            }
        ]
    }
```
I think this is least explicit, and less intuitive than the next approaches.

2. Declare another group similar to colors, where each `label-value` has its own associated properties:
```
{
    "multiscales": [
        {
            "datasets": [
                {
                    "path": "0"
                },
                {
                    "path": "1"
                },
                {
                    "path": "2"
                },
                {
                    "path": "3"
                }
            ],
            "version": "0.1"
        }
    ],
    "image-label": {
        "version": "0.1",
        "colors": [
               ...
        ],
        "properties": [
            {
                "label-value": 1,
                "class": "Urban",
                "area_m2": "400"

            },
            {
                "label-value": 2,
                "class": "Water",
                "area_m2": "1532"
            },
            {
                "label-value": 3,
                "class": "Agriculture",
                "area_m2": "590"

            }
        ]
    }
}
```
This is explicit, but has the disadvantage of duplicating the `label-value` definitions.

3. Make color another property e.g.
```
        "properties": [
            {
                "label-value": 1,
                "rgba": [
                    255,
                    100,
                    100,
                    255
                ],
[...]

PR #61, opened 2020-11-16
Add support for label properties in zattrs and plugin

DragaDoncila:
Add support for label properties of the form

```
        "properties": [
            {
                "label-value": 1,
                "class": "Urban",
                "m2": "100"
            },
            {
                "label-value": 2,
                "class": "Water",
                "m2": "200"

            },
            {
                "label-value": 3,
                "class": "Deciduous Woody Horticulture"

            },
            {
                "label-value": 4,
                "class": "Evergreen Woody Horticulture"

            },
            ...
```
as discussed in #60 .

Read these properties in as dictionary of `label-value` to properties.

Transform these properties in the napari reader to:
```
{
    "index": [
        1,
        2,
        3,
        4,
       ...
    ],
    "class": [
        "Urban",
        "Water",
        "Deciduous Woody Horticulture",
        "Evergreen Woody Horticulture",
        ...
    ],
    "m2": [
        "100",
        "200",
        "None",
        "None",
        "None",
       ...
    ]
}
```

This is the very first pass, so would appreciate feedback on the general approach before I move on to updating spec, documentation, tests, etc.

We probably also wish to consider what types of property keys and values we want to accept.

### DIFF
--- a/ome_zarr/napari.py
+++ b/ome_zarr/napari.py
@@ -62,6 +62,24 @@ def transform(nodes: Iterator[Node]) -> Optional[ReaderFunction]:
                     layer_type = "labels"
                     if "colormap" in metadata:
                         del metadata["colormap"]
+                    if "properties" in metadata:
+                        props = metadata["properties"]
+                        reader_props = {}
+                        label_indices = list(props.keys())
+                        reader_props["index"] = label_indices
+
+                        # properties may be ragged, so we need to know all possible properties
+                        all_keys = set()
+                        for index in label_indices:
+                            all_keys = all_keys.union(set(props[index].keys()))
+
+                        # napari expects lists of equal length so we must fill with None
+                        for prop_key in all_keys:
+                            reader_props[prop_key] = [
+                                props[i][prop_key] if prop_key in props[i] else "None"
+                                for i in label_indices
+                            ]
+                        metadata['properties'] = reader_props
 
                 elif shape[CHANNEL_DIMENSION] > 1:
                     metadata["channel_axis"] = CHANNEL_DIMENSION
--- a/ome_zarr/napari.py
+++ b/ome_zarr/napari.py
@@ -76,7 +76,7 @@ def transform(nodes: Iterator[Node]) -> Optional[ReaderFunction]:
                         # napari expects lists of equal length so we must fill with None
                         for prop_key in all_keys:
                             reader_props[prop_key] = [
-                                props[i][prop_key] if prop_key in props[i] else "None"
+                                props[i][prop_key] if prop_key in props[i] else None
                                 for i in label_indices
                             ]
                         metadata['properties'] = reader_props
--- a/ome_zarr/napari.py
+++ b/ome_zarr/napari.py
@@ -6,7 +6,7 @@ It implements the ``napari_get_reader`` hook specification, (to create a reader 
 
 import logging
 import warnings
-from typing import Any, Callable, Dict, Iterator, List, Optional
+from typing import Any, Callable, Dict, Iterator, List, Optional, Set
 
 from .data import CHANNEL_DIMENSION
 from .io import parse_url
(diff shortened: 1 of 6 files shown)


ome_zarr/reader.py
```diff
@@ -216,13 +216,22 @@ def __init__(self, node: Node) -> None:
                 except Exception as e:
                     LOGGER.error(f"invalid color - {color}: {e}")
 
+        properties: Dict[int, Dict[str, str]] = {}
+        props_list = image_label.get('properties', [])
+        if props_list:
+            for props in props_list:
+                label_val = props['label-value']
+                del props['label-value']
+                properties[label_val] =  props
```
manics, line 225: Could you explain what's happening here? We'll need to check whether the underlying dictionary is a deep-copy of the original, or whether it returns a reference which means this may update the properties seen by other code.

will-moore: Yes, if I do `image_label.get('properties', [])` after this `for` loop, I get the properties without the `label-value`s so you probably should create a copy of the `props` before editing.
This seems to work:
```
                properties[label_val] = dict(props)
                del properties[label_val]['label-value']
```

ome_zarr/napari.py
```diff
@@ -62,6 +62,24 @@ def f(*args: Any, **kwargs: Any) -> List[LayerData]:
...
+
+                        # properties may be ragged, so we need to know all possible properties
+                        all_keys = set()
+                        for index in label_indices:
+                            all_keys = all_keys.union(set(props[index].keys()))
+
+                        # napari expects lists of equal length so we must fill with None
+                        for prop_key in all_keys:
+                            reader_props[prop_key] = [
+                                props[i][prop_key] if prop_key in props[i] else "None"
```
will-moore, line 79: This can probably be an actual `None` value here instead of a string `"None"` (unless it *has* to be a string for some reason). Since if some code is testing for the value being `None`, then the string won't be recognised as `None`.

ome_zarr/napari.py
```diff
@@ -62,6 +62,24 @@ def f(*args: Any, **kwargs: Any) -> List[LayerData]:
                     layer_type = "labels"
                     if "colormap" in metadata:
                         del metadata["colormap"]
+                    if "properties" in metadata:
```
will-moore, line 65: This will always be True since you are always adding `"properties": properties,` in the reader, even if the `properties` is an empty dict. In which case your `reader_props` below will be `{'index': []}`, which doesn't break anything but could be cleaner.

will-moore:
This works fine for me. Can see the values in `napari` (bottom left) when hovering over the mask:

Screenshot 2020-11-16 at 14 05 01

Few minor comments but generally looking good.

ome_zarr/napari.py
DragaDoncila: I'll change this so we don't add properties if there's none to add

joshmoore:
Pushed a commit to update the miniconda action since it looks like your PR was running into:

```
Error: Unable to process command '::set-env name=CONDA_PKGS_DIR::/home/runner/conda_pkgs_dir' successfully.
25
Error: The `set-env` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
```

DragaDoncila:
> Pushed a commit to update the miniconda action since it looks like your PR was running into:
> 
> ```
> Error: Unable to process command '::set-env name=CONDA_PKGS_DIR::/home/runner/conda_pkgs_dir' successfully.
> 25
> Error: The `set-env` command is disabled. Please upgrade to using Environment Files or opt into unsecure command execution by setting the `ACTIONS_ALLOW_UNSECURE_COMMANDS` environment variable to `true`. For more information see: https://github.blog/changelog/2020-10-01-github-actions-deprecating-set-env-and-add-path-commands/
> ```

Thanks for this, pre-commit was giving me the run around...

DragaDoncila:
@joshmoore  I've added a test to `test_napari.py` for label properties. I was going to add one to `test_reader.py` as well but it looks like that doesn't really check metadata. Is there anywhere else I should add tests?

joshmoore:
Ah, @DragaDoncila. Sorry, I see I owe(d) you a comment regarding the spec & more tests. Your PR against omero-ms-zarr was perfect and against the right repo at the time. :100: You saw that I migrated it. Further PRs welcome against ome/zarr. :smile: Regarding tests, I imagine there will need to be a new file created for what you were wanted to accomplish. Feel free to open a new PR if you're interested. All the best, ~Josh

Merging along with https://github.com/ome/ngff/pull/3

PR #63, opened 2020-11-19
csv to label properties

will-moore:
Takes a csv file with 1 row per Mask (e.g. exported from OMERO with `Batch_ROI_Export`) and uses it to
add properties to the labels metadata in `ome-zarr`.
You need to specify which column of the csv contains an ID that matches a specified ID of each properties dict.

E.g. if you have exported labels from OMERO with https://github.com/ome/omero-cli-zarr/pull/50 then each properties dict will have e.g. `"omero:shapeId": 123` and this will correspond to the `shape_id` column of the Batch_ROI_Export.csv.

You also want to specify columns from the csv to use, e.g. `"area (pixels),X,Y,Width,Height"`.
You can also specify the column types (as in https://github.com/ome/omero-metadata/) to specify the data-type for each column (string by default).

 - `d`: `DoubleColumn`, for floating point numbers
 - `l`: `LongColumn`, for integer numbers
 - `s`: `StringColumn`, for text
 - `b`: `BoolColumn`, for true/false

Use e.g. `#d` as a suffix in the column name to denote a `float` column, no spaces etc:
```
"area (pixels)#d,well_label#s,Width#l,Height#l"
```

If you have a Plate exported to ome-zarr as "251.zarr", then you can do this:

```ome_zarr csv_to_labels plate251_ch1.csv shape_id "area (pixels)#d,well_label#s,Width#l,Height#l" 251.zarr omero:shapeId```

Then you can open a Well (field 0) in napari, and the properties will be read by https://github.com/ome/ome-zarr-py/pull/61 cc @DragaDoncila and you will be able to see the properties from the csv (bottom left) when you mouse over a label

```napari 251.zarr/A/1/0```

Screenshot 2020-11-19 at 09 13 55

You can then use the properties to update label colors:

```
layer = viewer.layers[-1]

for c, value in enumerate(layer.properties["area (pixels)"]):
    index = layer.properties["index"][c]
    if value > 2500:
        layer.color[index] = np.array([1, 0, 1, 1], dtype=np.float32)
    elif value < 2000:
        layer.color[index] = np.array([1, 1, 1, 1], dtype=np.float32)
    else:
        layer.color[index] = np.array([0, 1, 0, 1], dtype=np.float32)
   
# force refresh
layer.color_mode = "direct"
```

Screenshot 2020-11-19 at 13 28 40

TODO:

 - [ ] Also support single Images in ome-zarr instead of whole Plate
 - [x] Parse csv number columns as numbers

### DIFF
--- a/.github/workflows/posix.yml
+++ b/.github/workflows/posix.yml
@@ -19,7 +19,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Setup miniconda
-        uses: conda-incubator/setup-miniconda@v1
+        uses: conda-incubator/setup-miniconda@v2
         with:
           auto-update-conda: true
           channels: conda-forge,ome
--- a/.github/workflows/windows.yml
+++ b/.github/workflows/windows.yml
@@ -20,7 +20,7 @@ jobs:
         uses: actions/checkout@v2
 
       - name: Setup miniconda
-        uses: conda-incubator/setup-miniconda@v1
+        uses: conda-incubator/setup-miniconda@v2
         with:
           auto-update-conda: true
           channels: conda-forge,ome
--- a/README.rst
+++ b/README.rst
@@ -85,6 +85,41 @@ the metadata::
 
     $ napari '/tmp/6001240.zarr/0'
 
+csv to labels
+=============
+
+The `csv_to_labels` command uses a CSV file to add key:value properties to labels
+under an OME-Zarr Image or Plate.
+
+The OME-Zarr labels metadata must already contain a `properties`
+list of `{key:value}` objects, each with a unique key:ID. This key is `omero:shapeId`
+in the example below.
+
+This ID can be used to identify a single row of the CSV table by specifying the name of
+a column with unique values, e.g. `shape_id` below.
+This row is used to add additional column_name:value data to the label properties.
+
+You also need to specify which columns from the CSV to use, e.g. `"area,X,Y,Width,Height"`.
+You can also specify the column types (as in https://github.com/ome/omero-metadata/)
+to specify the data-type for each column (string by default).
+
+ - `d`: `DoubleColumn`, for floating point numbers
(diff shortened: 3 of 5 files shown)

### CODE REVIEW

**Comment by joshmoore** [comment]
NB: there's something odd about the relative paths, though I don't assume that's critical:

```
/data/josh/miniconda3/envs/o/bin/ome_zarr csv_to_labels /data/josh/idr0001_field0_ROIs.csv shape_id "area (µm)#d,min#l,max#l,mean#d" 2551.zarr omero:shapeId
...
PermissionError: [Errno 13] Permission denied: '/data/plates/2551.zarr/E/10/0/labels/0'
```
(when it should have been looking for `2551.zarr` in the current directory.)

**Comment by joshmoore** [comment]
Something like this was needed to omit missing labels:

```
(o) [jamoore@pilot-zarr1-dev 2551.zarr]$ (cd /data/josh/ome-zarr-py/; git diff)
diff --git a/ome_zarr/csv.py b/ome_zarr/csv.py
index de287e3..bdcb012 100644
--- a/ome_zarr/csv.py
+++ b/ome_zarr/csv.py
@@ -127,7 +127,10 @@ def dict_to_zarr(

     for path_to_labels in labels_paths:

-        label_group = zarr_open(path_to_labels)
+        try:
+            label_group = zarr_open(path_to_labels)
+        except:
+            continue
         attrs = label_group.attrs.asdict()
         properties = attrs.get("image-label", {}).get("properties")
         if properties is None:
```

will-moore:
Ah, Thanks. Yes, I've just viewed the plate with labels and noticed that some Wells are missing labels:

`napari https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/plates/2551.zarr`

Screenshot 2020-11-29 at 22 08 40

joshmoore:
Considering the test of 2551 successful since the source plate (2551 above) is also missing those labels. Merging for the next tag.

PR #64, opened 2020-12-01
Fix labels of differing dtype

joshmoore:
The first scale of the label was uint16 while following
levels were floating point:

```
cat 6001237.zarr/labels/masks/*/.zarray | grep dtype
    "dtype": "<i8",
    "dtype": "<f8",
    "dtype": "<f8",
    "dtype": "<f8",
    "dtype": "<f8",
```

cc: @will-moore @tischi

### DIFF
--- a/environment.yml
+++ b/environment.yml
@@ -4,7 +4,6 @@ channels:
   - ome
   - conda-forge
 dependencies:
-  - pyside2
   - napari
   - flake8
   - ipython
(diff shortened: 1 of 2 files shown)


joshmoore:
Re-opening to use the update build. A regeneration of the data on S3 is likely necessary.

joshmoore:
Interestingly the old jobs that have been removed are still running. cc: @sbesson

PR #51, opened 2020-10-01
napari test requirement

joshmoore:
Registering napari-conftest as a plugin in setup.py
makes a hard-requirement on napari. This PR moves the
requirement to tests only and registers the pytest
plugin at runtime.

With thanks to @Czaki

see #https://github.com/napari/napari/issues/1661

Note: that the environment.yml file is currently not tested by this PR and I in fact needed to actively remove `tox-conda` to get things to work. We will likely need to either remove it (in favor of an ome-zarr package on conda?) or add in separate testing.

PR #66, opened 2020-12-02
Add testing of Zarr development branch

Carreau:
Only on linux and a single version of Python.

This would help to catch potential regressions.

### DIFF
--- a/.github/workflows/zarr-dev.yml
+++ b/.github/workflows/zarr-dev.yml
@@ -1,0 +1,47 @@
+---
+name: Build using Zarr development branch
+
+on: [push, pull_request]
+
+jobs:
+  test:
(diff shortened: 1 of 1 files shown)


joshmoore:
Thanks, @Carreau. Happy to have this in. Also, good to see another working version after the fun of gh-51. Depending on what happens with `build.yml`, I may refactor this into the existing workflow and use another build dimension for different zarr versions.

Carreau:
> I may refactor this into the existing workflow and use another build dimension for different zarr versions.

I thought about doing this but I believed that testing on only ome zarr version and only linux was sufficient, which would be annoying with adding a dimension.

joshmoore:
We'll see ;) We also have a build that switches to a French Locale once a week to help detect bugs.

PR #67, opened 2020-12-04
Register napari conftest

Czaki:
@joshmoore I test this locally and (with removed skip) all tests passed. 
May you think to change this if to detect if the test is run on CI? 

There is `GITHUB_ACTIONS` in the environment.

```
GITHUB_ACTIONS = "GITHUB_ACTIONS" in os.environ
```


joshmoore:
Thanks, @Czaki. Certainly looking better in #51.

> May you think to change this if to detect if the test is run on CI?

You're saying that you would disable Qt tests on CI completely?

Czaki:
> You're saying that you would disable Qt tests on CI completely?

No. If the problem with test is only on CI (because of GL drive or no screen), then test should be disabled only on CI.

joshmoore:
Of course! I had hoped the fix would be quick and we could remove that but unfortunately that hasn't been the case.

Czaki:
> Of course! I had hoped the fix would be quick and we could remove that but unfortunately that hasn't been the case.

In my program, I disable part of tests that use napari viewer because of problems on CI, where I do not have time to investigate what wrong is with the configuration of environments. MacOS cover all tests.

joshmoore:
Understood. That may be the next step if I can't get it working this morning. Another option I've been considering is to extract the napari plugin out of ome-zarr-py (e.g. `napari-ome-zarr`).

joshmoore:
Closing now. Thanks again!

PR #69, opened 2020-12-04
attempts to create a writer

glyg:
This code base does not offer a clear way to _create_  zarr images from scratch (appart for the function in the `data` module). Although bioimage2raw works fine for already existing data, at some point data will be recorded directly in zarr, especially in custom microscope setups.

So, following a discussion with Josh on image.sc, here is an attempt at a `write_image` function.

This is very early, in particular I think the `ZarrLocation` classes should be adapted to writing and better manage paths and groups.

Thanks for any input!


ome_zarr/writer.py
```diff
@@ -0,0 +1,76 @@
...
+      the group within the zarr store to store the data in
+    chunks: int or tuple of ints,
+      size of the saved chunks to store the image
+    byte_order: str or list of str, default "tczyx"
+      combination of the letters defining the order
+      in which the dimensions are saved
+    """
+    zarr_location = parse_url(path)
+
+    ## I assume this should be dealt with in
```
joshmoore, line 43: Agreed. I think what's needed first is a way to pass `mode` through to the `parse_url` and `*Location` classes.

ome_zarr/writer.py
```diff
@@ -0,0 +1,76 @@
...
+            "z": image.shape[2],
+            "height": image.shape[3],
+            "width": image.shape[4],
+            }
+
+    metadata["omero"] = omero
+
+    root = zarr.group(store)
+    if group is not None:
+        grp = root.create_group(group)
```
joshmoore, line 69: `require_group` should do what you want.

joshmoore (commented):
Thanks, @glyg. Generally looks good. Let me know if you want to try the integration into `ome_zarr.io` or not. I like the `metadata` argument :+1: The `byte_order` argument may go through a refactoring soon as we loosen the dimension order early next year, but this makes good sense for the moment.

I should have the github actions happy again after lots of work in #51. I'll try to push something today to make it go green.

Cheers.

joshmoore:
Pushed commits to:
 - merge master to disable test_napari.py for zarr-dev
 - fix the output of `pre-commit run -a`

joshmoore:
Thanks for the update, @glyg. Do you want to run `pre-commit run -a` and push, or shall I? (I can recommend running `pre-commit install` in any repo that requires it)

glyg:
Hi @joshmoore I did not know `pre-commit` I'll apply it. I'm trying to integrate the `ZarrLocation` mechanisms to the writer, will push (and run pre-commit before :) once I have something functionnal

joshmoore:
Sounds great! Thanks again, @glyg.

glyg:
Ok, I have some issues

First pre-commit fails but I don't know how to make mypy happy

Second, when it seems that when only OMERO specs are found by the reader, `node.data` is not set (although the array is there). 
I'm not sure whether this is intended or how to manage it.

Finally I assume the writer should be coded as an object, but I'm not sure how.

joshmoore:
> First pre-commit fails but I don't know how to make mypy happy

```
ome_zarr/writer.py:52: error: Argument "zarr" to "Node" has incompatible type "Optional[BaseZarrLocation]"; expected "BaseZarrLocation"
```

This means that the variable somewhere could conceivably be `None`. You can either perform the `if zarr == None` test, or make sure it's never `None`.

```
ome_zarr/writer.py:65: error: Incompatible types in assignment (expression has type "Tuple[Any, ...]", variable has type "Union[Tuple[int], int, None]")
```
You might want to create a new variable like `_chunks: Tuple[int] = ...` then the "compiler" should know what you're trying to do.

glyg:
*  _I had to tell the OMERO spec to put the data in the node at read time, I'm not sure it does not break anything_ (although the tests pass)

* Is  a class needed for the writer?

* random channel colors might be ugly

joshmoore:
> random channel colors might be ugly

Seems like that can be handled in the future.


> Is  a class needed for the writer?

I was wondering about that. I don't think it's a MUST. Did you find any state that you would have felt useful to keep in such an object?

> I had to tell the OMERO spec to put the data in the node at read time, I'm not sure it does not break anything

I'll take a look.

joshmoore:
> I had to tell the OMERO spec to put the data in the node at read time,

Do you think this is a problem with the order of operations? i.e. if the data nodes ran before the metadata nodes?

glyg:
HI Josh,

> Do you think this is a problem with the order of operations?

Yes, it is not clear to me when data is appended to `node.data`. If you have pyramid images, then the `Multiscale` spec will access the data, but the `OMERO` spec will also be present. If it's always accessed last, then testing if `node.data` is empty as I did should be enough? 

> Did you find any state that you would have felt useful to keep in such an object?

The scenario I was thinking might justify the use of a class is if you want to have a `Writer` instance in which you append children / nodes at data generation time. e.g. during the course of an acquisition. Then some kind of object persistence would be interesting?

ome_zarr/writer.py
```diff
@@ -0,0 +1,112 @@
...
+
+from .io import parse_url
+from .reader import Node
+from .types import JSONDict
+
+LOGGER = logging.getLogger("ome_zarr.writer")
+
+
+def write_image(
+    path: str,
```
joshmoore, line 21: I could imagine passing a `zarr.Group` here rather than path+name+group

ome_zarr/writer.py
```diff
@@ -0,0 +1,112 @@
...
+    image: np.ndarray,
+    name: str = "0",
+    group: str = None,
+    chunks: Union[Tuple[int], int] = None,
```
joshmoore, line 25: If this method is to handle downsampling internally, I wonder if we want to pass multiple chunk sizes or keep them consistent.

ome_zarr/writer.py
```diff
@@ -0,0 +1,112 @@
...
+    node = Node(zarr=zarr_location, root=[])
+
+    if image.ndim > 5:
+        raise ValueError("Only images of 5D or less are supported")
+
+    shape_5d: Tuple[Any, ...] = (*(1,) * (5 - image.ndim), *image.shape)
+    image = image.reshape(shape_5d)
+
+    if chunks is None:
+        image = da.from_array(image)
```
joshmoore, line 63: If no reshaping is necessary, it _might_ be preferred to leave the input array as is and use a convenience zarr function to improve speed, but I'd need to do some testing.

ome_zarr/writer.py
```diff
@@ -0,0 +1,112 @@
...
+                {
+                    "color": "".join(f"{i:02x}" for i in color),
+                    "window": {"start": 0, "end": 1},
+                    "active": True,
+                }
+                for color in colors
+            ]
+            omero["rdefs"] = {"model": "color"}
+
+    metadata["omero"] = omero
```
joshmoore, line 96: This is currently missing the `multiscales` metadata. I'd propose moving `write_multiscales` from `ome_zarr.data` to your writer module.

joshmoore (commented):
Hi @glyg. I just tried this out while trying to come up with an example for @oeway, do you mind if I push a few commits for your review? Alternatively, do you have any changes that are coming yourself?

glyg:
> Hi @glyg. I just tried this out while trying to come up with an example for @oeway, do you mind if I push a few commits for your review? Alternatively, do you have any changes that are coming yourself?

Hi @joshmoore of course feel free to push on this PR. I was actually  waiting for your feedback before pursuing this, so there is nothing coming from me.
G.

ome_zarr/writer.py
glyg: yes makes sense  (I realize the group argument here is not used). 
Do you think both way of specifying the output should co-exist or it's better to leave the job of creating the group to the user all the time?

ome_zarr/writer.py
glyg: seems to me it's easier to keep the same chunks for all sample sizes. Or scale the number of chunks - maybe in the x, y dimensions - to have a consistent chunk size across scales? I'd assume the specified chunk size is for the 1:1 scale.

It sounds complicated to ask the user to input one chunk size for each scale - imho it should be as transparent as possible. 

Maybe it will get clearer when `write_multiscales` is integrated here?

ome_zarr/writer.py
glyg: I'll let you the judge of that :)

glyg:
Hi @joshmoore - finally, should I apply the changes you suggest? I can do that today or early tomorrow.

Best,

Guillaume

joshmoore:
Hi Guillame, sorry meetings have run long today. 🙂  Let me open a PR with my work-in-progress and you can give me feedback, trash it, etcl. ~J.

joshmoore:
See https://github.com/glyg/ome-zarr-py/pull/1

ome_zarr/writer.py
joshmoore: Let's start with leaving it to the user since there are so many options (local v. remote, etc.) but if we realize we're doing the same thing over and over again, we'll provide a helper.

ome_zarr/writer.py
joshmoore: Yeah, I had gotten that far ;) I think you're right that it's more difficult than one would want in general. Is it worth though providing all the options, i.e.

```
Union[Tuple[Tuple[int]], Tuple[int], int]
```

ome_zarr/writer.py
glyg: osund reasonable :)

ome_zarr/io.py
```diff
@@ -189,7 +193,7 @@ def get_json(self, subpath: str) -> JSONDict:
             return {}
 
 
-def parse_url(path: str) -> Optional[BaseZarrLocation]:
+def parse_url(path: str, mode: str = "r") -> Optional[BaseZarrLocation]:
```
glyg, line 196: As this function is not used in the write method, maybe I should revert the changes in `io.py`

joshmoore: I was working on a simple example for Wei (his use case was "how to turn an existing zarr array into an ome-zarr"). Let's keep this until we have a simple working user example. Maybe `parse_url` can still be the starting point.

<details><summary>previous example</summary>

```
#!/usr/bin/env python
import numpy as np
import zarr as zr

from ome_zarr.reader import OMERO, Reader

import argparse
parser = argparse.ArgumentParser()
parser.add_argument("path")
ns = parser.parse_args()

dtype = np.uint8
mean = 10
shape = (1, 3, 1, 256, 256)
rng = np.random.default_rng(0)
data = rng.poisson(mean, size=shape).astype(dtype)
store = zr.DirectoryStore(ns.path)
grp = zr.group(store)
write_image(data, grp, chunks=(128, 128))
```

</details>

joshmoore:
@glyg very sorry to have let this slipped. Getting back to my example for @oeway:

```
#!/usr/bin/env python
import numpy as np
import zarr as zr

import argparse
parser = argparse.ArgumentParser()
parser.add_argument("path")
ns = parser.parse_args()

# Set up your numpy, whatever it is
dtype = np.int8
mean = 1000
shape = (1, 3, 10, 4096, 4096)
rng = np.random.default_rng(0)
data = rng.poisson(mean, size=shape).astype(dtype)

# Write it out to a Zarr group
store = zr.DirectoryStore(ns.path)
grp = zr.group(store)
write_image(data, grp, chunks=(128, 128))
```

looks like the only thing missing was the downsampling. I went ahead and added that in a commit. We may need to change the API slightly to let someone choose _not_ to downsample, e.g. by passing `scaler=None` or by passing a pre-computed pyramid, but as far as I can tell, things are all working.

If my last commit goes green, let's get this in and start hammering on it!

Thanks for your patience. (Though feel free to kick me in the future.)

ome_zarr/writer.py
```diff
@@ -0,0 +1,108 @@
...
+
+
+def write_multiscale(
+    pyramid: List, group: zarr.Group, chunks: Union[Tuple[Any, ...], int] = None,
+) -> None:
+    """Write a pyramid with multiscale metadata to disk."""
+    paths = []
+    for path, dataset in enumerate(pyramid):
+        # TODO: chunks here could be different per layer
+        group.create_dataset(str(path), data=pyramid[path], chunks=chunks)
```
will-moore, line 23: `dataset` variable is not used, unless you want to do `data=dataset` ?

will-moore (approved):
Looks good. Running test locally and inspecting the output looks as expected.
Happy for this to be merged and I'll try using it 'for real'

ome_zarr/writer.py
joshmoore, line 23:
```suggestion
        group.create_dataset(str(path), data=dataset, chunks=chunks)
```

ome_zarr/writer.py
```diff
@@ -0,0 +1,108 @@
...
+        raise ValueError("Only images of 5D or less are supported")
+
+    shape_5d: Tuple[Any, ...] = (*(1,) * (5 - image.ndim), *image.shape)
+    image = image.reshape(shape_5d)
+
+    if chunks is not None:
+        chunks = _retuple(chunks, shape_5d)
+    omero = metadata.get("omero", {})
+    # Update the size entry anyway
+    omero["size"] = {
```
will-moore, line 63: I don't think we need to have `size` metadata here since we have that info in the `.zarray` and it's not part of the OME-Zarr spec.

ome_zarr/writer.py
```diff
@@ -0,0 +1,108 @@
...
+        if size_c == 1:
+            omero["channels"] = [{"window": {"start": 0, "end": 1}}]
+            omero["rdefs"] = {"model": "greyscale"}
+        else:
+            rng = np.random.default_rng(0)
+            colors = rng.integers(0, high=2 ** 8, size=(image.shape[1], 3))
+            omero["channels"] = [
+                {
+                    "color": "".join(f"{i:02x}" for i in color),
+                    "window": {"start": 0, "end": 1},
```
will-moore, line 81: It's probably better to omit all the `omero['channels']` metadata if we're not provided with it.
Generating these numbers will cause viewers to assume they are based on the data, which could lead to poor visualisation (e.g. render levels 0 - 1 will likely be saturated)

joshmoore: @glyg : any thoughts from your initial use case?

will-moore:
In testing this, I adapted `omero-cli-zarr` to use it for writing - See https://github.com/ome/omero-cli-zarr/pull/58.
Not sure we want to use this approach (see PR) but it seems to be working.

ome_zarr/writer.py
glyg: I think I just tried to emulate the code I saw in the tests or elsewhere, I agree that by default metadata entries should as small as possible

ome_zarr/reader.py
```diff
@@ -354,7 +354,9 @@ def __init__(self, node: Node) -> None:
             node.metadata["visible"] = visibles
             node.metadata["contrast_limits"] = contrast_limits
             node.metadata["colormap"] = colormaps
+
         except Exception as e:
+            raise e
```
glyg, line 359: Should this be there? The LOGGER line below will never be called

joshmoore: Pretty likely debugging code that crept in. Thanks, @glyg

ome_zarr/writer.py
```diff
@@ -0,0 +1,108 @@
...
+    chunks: Union[Tuple[Any, ...], int], shape: Tuple[Any, ...]
+) -> Tuple[Any, ...]:
+
+    _chunks: Tuple[Any, ...]
+    if isinstance(chunks, int):
+        _chunks = (chunks,)
+    else:
+        _chunks = chunks
+
+    return (*shape[: (5 - len(_chunks))], *_chunks)
```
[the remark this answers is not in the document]
joshmoore: Good point, but let's handle that along with #39. Ultimately I think the rules we'll have to be laid out, the most flexible being "once the shape is defined (e.g. `len(shape)==4`), then everything else (e.g. chunks) should stay in that shape & order.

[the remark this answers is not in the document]
glyg: Ok great!

joshmoore:
@will-moore @glyg : pushed a few fixes:
 * Remove the metadata as agreed
 * Fixed the `raise e` issue
 * Allowed `scaler` to be set to `None`

A few things that I think *need* to happen, but can be outside this PR:
 * Pyramids should be pre-computable, but that conflicts with the dimension detection, so we can do that with the other dimension fixes.
 * I think the `Scaler` class needs a `__call__` method which defaults to using `nearest()` internally. So then a scaler is basically just a function which takes an array and generates a pyramid.

will-moore:
@joshmoore You said "pushed a few fixes", but the last commit I see is https://github.com/ome/ome-zarr-py/pull/69/commits/b832df5bb940fd4d1a0fa352adf941eed9a950f3 (1 commit yesterday)?
Am I missing some?

joshmoore:
Doh! Sorry, @will-moore. I've made that mistake several times now. Pushed to _my_ `writer` branch, rather than @glyg's. Commits are there now.

will-moore:
Looks good.
Tests are passing etc.
Tried with/without `omero` metadata and with `scaler=None` from my `omero-cli-zarr` PR above. All works as expected.

joshmoore:
_Whew_. Merging :tada: Thanks for your effort (& patience!) here, @glyg.

glyg:
thanks a lot @joshmoore and @will-moore , very happy to be of some help :)

PR #70, opened 2020-12-06
Fix labels of differing dtype

joshmoore:
The first scale of the label was uint16 while following
levels were floating point:

```
cat 6001237.zarr/labels/masks/*/.zarray | grep dtype
    "dtype": "<i8",
    "dtype": "<f8",
    "dtype": "<f8",
    "dtype": "<f8",
    "dtype": "<f8",
```

A regeneration of the data on S3 is likely necessary.

cc: @will-moore @tischi


joshmoore:
Tested by Tischi with 6001237. Merging and releasing.

PR #72, opened 2021-02-04
pre-releases.yml: run tox with --pre periodically

joshmoore:
Based on https://github.com/napari/napari/blob/master/.github/workflows/test_prereleases.yml with thanks to @Czaki

### DIFF
--- a/.github/workflows/pre-releases.yml
+++ b/.github/workflows/pre-releases.yml
@@ -1,0 +1,73 @@
+# An "early warning" cron job that will install dependencies
+# with `pip install --pre` periodically to test for breakage
+# (and open an issue if a test fails)
+name: --pre Test
+
+on:
+  push:
+    branches:
+      - master
+  schedule:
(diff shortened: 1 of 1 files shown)


tlambert03:
I'm not sure this is going to work for you... if you're going to use this strategy, you're minimally going to need the template that I added here [`.github/TEST_FAIL_TEMPLATE.md`](https://github.com/napari/napari/blob/master/.github/TEST_FAIL_TEMPLATE.md).  Also, all of the environment variables in this test.yml are specifically catered to the napari tox.ini file, so I'm not sure this is going to run at all?

joshmoore:
Yeah, it wasn't intended to be merged. Just one too many things going on. It's definitely been one of those days. I'll just push the commits away and start with another PR.

joshmoore:
These commits have been rolled back, but I'll keep testing out the general strategy. Thanks, both.

PR #80, opened 2021-04-27
Upgrade v0.1 to v0.2

joshmoore:
* [x] Create `Format` hierarchy with `FormatV01` and `FormatV02` classes
 * [x] Auto-detect format and substitute the correct version
 * [ ] ~Add CLI method to perform upgrade~ (see follow-up)


joshmoore:
Ok. I've reviewed what I still had locally. A few updates:

 * I've pushed 2 commits which should be pretty harmless / remain green.
 * Locally, I have an `def upgrade()` but it still has issues.
   * These are related to https://github.com/zarr-developers/zarr-python/issues/772.
   * Basically, as far as I know, there's no method to migrate one store to another taking metadata into account. If anyone knows otherwise, I'm all ears.
   * I'll open a follow PR with all of that so everyone can see the issue.
* Once the build goes green, I'll merge so @constantinpape can build on top of this as discussed in #89 which will hopefully shake out any of the issues in the versioning API.

PR #82, opened 2021-05-04
Pin Zarr to 2.8.1 or above

joshmoore:
In concert with https://github.com/zarr-developers/zarr-python/issues/735 this should improve the user experience in napari for nested datasets.


joshmoore:
Also added minimal fsspec version here based on https://github.com/zarr-developers/zarr-python/issues/735#issuecomment-831956172, i.e. if zarr doesn't have a minimal version then we will need to do it here.

joshmoore:
Have seen others having issues with this. Going to go ahead and roll it out.

PR #89, opened 2021-06-11
Update reader.py to v0.3

constantinpape:
Changes for upcoming multiscale spec v0.3

PR #91, opened 2021-06-14
Cleanup builds

joshmoore:
After the extraction of the napari plugin to `napari-ome-zarr`,
a number of the matrix builds can be eliminated. This PR attempts
to drop any dependencies on Qt-related code which should all be in
the plugin repository.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -10,33 +10,16 @@ jobs:
       fail-fast: false
       matrix:
         include:
-
-          - {os: windows-latest, python_Version: '3.6', toxenv: 'py36-PyQt5'}
-          - {os: windows-latest, python_Version: '3.6', toxenv: 'py36-PySide2'}
-          - {os: windows-latest, python_Version: '3.7', toxenv: 'py37-PyQt5'}
-          - {os: windows-latest, python_Version: '3.7', toxenv: 'py37-PySide2'}
-          - {os: windows-latest, python_Version: '3.8', toxenv: 'py38-PyQt5'}
-          - {os: windows-latest, python_Version: '3.8', toxenv: 'py38-PySide2'}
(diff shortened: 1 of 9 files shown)


joshmoore:
@will-moore : in order to get rid of the vispy dependency, can you see just adding these values raw rather than wrapping them with a `Colormap`?

```
                color = ch.get("color", None)
                if color is not None:
                    rgb = [(int(color[i : i + 2], 16) / 255) for i in range(0, 6, 2)]
                    # TODO: make this value an enumeration
                    if model == "greyscale":
                        rgb = [1, 1, 1]
                    colormaps.append(Colormap([[0, 0, 0], rgb]))
```

will-moore:
A `Colormap` was added here because that's what `napari` was expecting.

It looks like `napari` is still expecting a [napari colormap](https://napari.org/docs/0.4.7/_modules/napari/utils/colormaps/colormap.html#Colormap) in `add_image`: https://napari.org/docs/0.4.7/api/napari.Viewer.html#napari.Viewer.add_image but this won't be needed
now that napari support is moved to napari-ome-zarr.

joshmoore:
So store `[[0, 0, 0], rgb]` and then transform in napari-ome-zarr? Or introduce an object here?

will-moore:
I would think that `[[0, 0, 0], rgb]` would be fine, but happy to use an Object if it's clearer etc.

joshmoore:
Failure is caused by the recently merged https://github.com/zarr-developers/zarr-python/pull/781

constantinpape (approved):
@joshmoore this looks good. Could we merge this now and then merge main into #89 in order to simplify the tests?

joshmoore:
Yeah, let's do that. Then we can build up our confidence in this a bit more in this PR before cutting a release.

PR #93, opened 2021-06-21
update link

jburel:
Point to the new location


joshmoore:
Thanks for finding that, @jburel. As a side note, I needed to explicitly _not_ use `https://ngff.openmicroscopy.org/latest` recently in favor of the current "stable" version (v0.2). Not sure if that would be a URL we'd want to redirect from.

PR #109, opened 2021-09-03
Update Embassy URLs to use new Embassy Cloud object storage

sbesson:
See https://forum.image.sc/t/hcs-ome-zarr-data-unable-to-download/57114/6

The current S3 URLs are regularly hitting 503 errors which might be related to the fact these resources are being EOL'ed.

The three  OME-Zarr samples have been sync'ed to the new Embassy Cloud v4 storage.

Opening this PR as a prererequisite of the overall migration

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -45,7 +45,7 @@ info
 Use the `ome_zarr` command to interrogate Zarr datasets::
 
     # Remote data
-    $ ome_zarr info https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001240.zarr/
+    $ ome_zarr info https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001240.zarr/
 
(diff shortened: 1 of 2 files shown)


joshmoore:
Thanks, @sbesson. I still have _some_ hope for the redirects from `s3.*`, but this does raise a long-term question. Not sure if it wouldn't be safer to go with our own proxy (like idr-analysis).

sbesson:
I agree that if we are going to tell our community they need to change their URLs/bookmarks, we might want to ensure they only update once.

As a starting point, there is some proxying infrastructure already in place with `idr-s3.openmicroscopy.org` (see https://github.com/IDR/deployment/pull/280). We could certainly try to build on top of that (potentially move it as part of the `prodXX-proxy`) and test it in various context.

joshmoore:
Merging this to have things more stable, but we may want to re-address with a proxy in the future.

PR #108, opened 2021-09-02
Fix HCS loading (close #96)

joshmoore:
* [x] Load first known well rather than image 0
* [x] append `/` to URLs on subpath loading (unclear why this was needed)

PR #111, opened 2021-09-16
Fix hcs 96

will-moore:
This is on top of #108 (attempt to fix HCS data viewing in v0.3 spec).

However, I have also just observed that the fixes in that PR are also needed for displaying labels with:

```
napari https://minio-dev.openmicroscopy.org/idr/v0.3/idr0062-blin-nuclearsegmentation/6001247.zarr
```

The extra commit on this branch is a cleanup of support for some HCS data generated before v0.1 (containing 'acquisitions').

It would be good to have labels working again in a release, to use as a public example at https://forum.image.sc/t/prepping-and-including-roi-masks/48750/17

### DIFF
--- a/ome_zarr/io.py
+++ b/ome_zarr/io.py
@@ -163,11 +163,14 @@ class ZarrLocation:
 
     def subpath(self, subpath: str = "") -> str:
         if self.__store.fs.protocol == "file":
-            path = Path(self.__path) / subpath
-            path = path.resolve()
-            return str(path)
+            filename = Path(self.__path) / subpath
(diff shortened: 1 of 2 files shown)


joshmoore:
Merging under the assumption that Will and I have reviewed each other's work. Proposing to release quickly as 0.1.1 (unless anyone thinks the removal of acquisitions is bumping...)

will-moore:
👍  No, acquisitions was just something we tried out and decided wasn't what we wanted in the spec. I only left it in temporarily so we could view a couple of the plates that I'd created in the meantime.

PR #118, opened 2021-10-05
add dispatch workflow

jburel:
Dispatch event to run the test.
We can add other repositories like napari one when we are ready
cc @will-moore


joshmoore (approved):
Looks fine to me.

PR #123, opened 2021-10-21
Validate axes for writer

### LINKED ISSUE #122: write_image assumes particular axisorder
Heyo, I just tried to naively write an ndarray to a an ome.zarr file, but didn't manage. My array was something like `numpy.zeros((100, 200, 1), dtype='uint8')` where I considered axes to be `yxc`.

of course when calling `write_image` I specified the `axes` argument. The `scaler`, however, will assume a certain axisorder. This leads to weird errors that are not super obvious.

I'd say this should be documented, or circumvented (e.g. making `Scaler` aware of axes, using a temporary adding the appropriate axes, etc...)

will-moore:
Fixes #122

This uses @constantinpape code from https://github.com/ome/ome-ngff-prototypes/blob/05b55d2516941e2eaf8fa82b722cad4371f99b5f/single_image/prototypes/v03.py#L19 to validate axes.

Also adds the required dimension order to the docstring for `write_image()` and `write_multiscale()`

cc @k-dominik 

To test, try the code sample below: https://github.com/ome/ome-zarr-py/pull/123#issuecomment-949691217 as it is, and with various shapes and valid/invalid axes. (invalid axes are not in the `tczyx` order, e.g. `xyz`. This should throw an `AssertionError`).
NB: This previously failed to scale with unhelpful error message (since the sizeX is 1), but error should now be improved.

### DIFF
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -38,17 +38,21 @@ def validate_axes_names(
         # from https://github.com/constantinpape/ome-ngff-implementations/
         val_axes = tuple(axes)
         if ndim == 2:
-            assert val_axes == ("y", "x"), str(val_axes)
+            if val_axes != ("y", "x"):
+                raise ValueError(f"2D data must have axes ('y', 'x') {val_axes}")
         elif ndim == 3:
-            assert val_axes in [("z", "y", "x"), ("c", "y", "x"), ("t", "y", "x")], str(
-                val_axes
-            )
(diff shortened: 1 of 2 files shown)


joshmoore:
Barring feedback from @k-dominik, is this waiting on anything?

will-moore:
No, I think it's good to go

k-dominik (approved):
Awesome, I really like the early failing with a super helpful error message! thx a lot!

sbesson:
Thanks @k-dominik and @will-moore

PR #124, opened 2021-10-28
Validate axes types v0.4

will-moore:
This builds on top of #123 to add validation of new axes dicts, based on the axes types.

Based on spec at https://github.com/ome/ngff/pull/57 (not merged yet).

There is a fair bit of validation code here cc @glyg 

Breaking changes:
 - reader `axes` now returns a list of dicts instead of list of str
 - OME-Zarr data is written as v0.4

But the writing API isn't breaking. Anything that was valid for v0.3 is also valid for v0.4, since we automatically add `type` info for `tczyx` dimensions.

### DIFF
--- a/ome_zarr/axes.py
+++ b/ome_zarr/axes.py
@@ -13,10 +13,15 @@ class Axes:
         axes: Union[List[str], List[Dict[str, str]]],
         fmt: Format = CurrentFormat(),
     ) -> None:
+        """
+        Constructor, transforms axes and validates
 
+        Raises ValueError if not valid
+        """
         if axes is not None:
             self.axes = self._axes_to_dicts(axes)
         self.fmt = fmt
+        self.validate()
 
     def validate(self) -> None:
         """Raises ValueError if not valid"""
--- a/ome_zarr/axes.py
+++ b/ome_zarr/axes.py
@@ -1,0 +1,100 @@
+"""Axes class for validating and transforming axes
+"""
+from typing import Any, Dict, List, Union
+
+from .format import Format
+
+KNOWN_AXES = {"x": "space", "y": "space", "z": "space", "c": "channel", "t": "time"}
+
+
+class Axes:
+    def __init__(self, axes: Union[List[str], List[Dict[str, str]]]) -> None:
+        self.axes = self._axes_to_dicts(axes)
+
+    def validate(self, fmt: Format) -> None:
+
+        # check names (only enforced for version 0.3)
+        if fmt.version == "0.3":
+            self._validate_axes_03()
+            return
+
+        self._validate_axes_types()
+
+    def get_axes(self, fmt: Format) -> Union[List[str], List[Dict[str, str]]]:
+        if fmt.version == "0.3":
+            return self._get_names()
+        return self.axes
+
+    @staticmethod
+    def _axes_to_dicts(
+        axes: Union[List[str], List[Dict[str, str]]]
+    ) -> List[Dict[str, str]]:
+        """Returns a list of axis dicts with name and type"""
+        axes_dicts = []
+        for axis in axes:
+            if isinstance(axis, str):
+                axis_dict = {"name": axis}
+                if axis in KNOWN_AXES:
+                    axis_dict["type"] = KNOWN_AXES[axis]
+                axes_dicts.append(axis_dict)
+            else:
+                axes_dicts.append(axis)
+        return axes_dicts
+
+    def _validate_axes_types(self) -> None:
+        """
+        Validate the axes types according to the spec, version 0.4+
+        """
+        axes_types = [axis.get("type") for axis in self.axes]
+        known_types = list(KNOWN_AXES.values())
+        unknown_types = [atype for atype in axes_types if atype not in known_types]
+        if len(unknown_types) > 1:
+            raise ValueError(
+                "Too many unknown axes types. 1 allowed, found: %s" % unknown_types
+            )
+
+        def _last_index(item: str, item_list: List[Any]) -> int:
+            return max(loc for loc, val in enumerate(item_list) if val == item)
+
+        if "time" in axes_types and _last_index("time", axes_types) > 0:
+            raise ValueError("'time' axis must be first dimension only")
+
+        if axes_types.count("channel") > 1:
+            raise ValueError("Only 1 axis can be type 'channel'")
+
+        if "channel" in axes_types and _last_index(
+            "channel", axes_types
+        ) > axes_types.index("space"):
+            raise ValueError("'space' axes must come after 'channel'")
+
+    def _get_names(self) -> List[str]:
+        """Returns a list of axis names"""
+        axes_names = []
+        for axis in self.axes:
+            if "name" not in axis:
+                raise ValueError("Axis Dict %s has no 'name'" % axis)
+            axes_names.append(axis["name"])
+        return axes_names
+
+    def _validate_axes_03(self) -> None:
+
+        val_axes = tuple(self._get_names())
+        if len(val_axes) == 2:
+            if val_axes != ("y", "x"):
+                raise ValueError(f"2D data must have axes ('y', 'x') {val_axes}")
+        elif len(val_axes) == 3:
+            if val_axes not in [("z", "y", "x"), ("c", "y", "x"), ("t", "y", "x")]:
+                raise ValueError(
+                    "3D data must have axes ('z', 'y', 'x') or ('c', 'y', 'x')"
+                    " or ('t', 'y', 'x'), not %s" % (val_axes,)
+                )
+        elif len(val_axes) == 4:
+            if val_axes not in [
+                ("t", "z", "y", "x"),
+                ("c", "z", "y", "x"),
+                ("t", "c", "y", "x"),
+            ]:
+                raise ValueError("4D data must have axes tzyx or czyx or tcyx")
+        else:
+            if val_axes != ("t", "c", "z", "y", "x"):
+                raise ValueError("5D data must have axes ('t', 'c', 'z', 'y', 'x')")
--- a/ome_zarr/axes.py
+++ b/ome_zarr/axes.py
@@ -2,20 +2,30 @@
 """
 from typing import Any, Dict, List, Union
 
-from .format import Format
+from .format import CurrentFormat, Format
 
 KNOWN_AXES = {"x": "space", "y": "space", "z": "space", "c": "channel", "t": "time"}
 
 
 class Axes:
-    def __init__(self, axes: Union[List[str], List[Dict[str, str]]]) -> None:
-        self.axes = self._axes_to_dicts(axes)
+    def __init__(
+        self,
+        axes: Union[List[str], List[Dict[str, str]]],
+        fmt: Format = CurrentFormat(),
+    ) -> None:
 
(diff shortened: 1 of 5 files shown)


ome_zarr/writer.py
```diff
@@ -13,18 +13,75 @@
...
 LOGGER = logging.getLogger("ome_zarr.writer")
 
+KNOWN_AXES = {"x": "space", "y": "space", "z": "space", "c": "channel", "t": "time"}
 
-def _validate_axes_names(
-    ndim: int, axes: Union[str, List[str]] = None, fmt: Format = CurrentFormat()
-) -> Union[None, List[str]]:
-    """Returns validated list of axes names or raise exception if invalid"""
+
+def _axes_to_dicts(
```
sbesson, line 19: The growth and the complexity of the different methods here make me wonder whether:

1. we want to create an object to handle the creation/validation/conversion
2. we want to split the format specificity into `AxesV03`, `AxesV04`

ome_zarr/writer.py
```diff
@@ -13,18 +13,75 @@
...
+    axes: Union[List[str], List[Dict[str, str]]]
+) -> List[Dict[str, str]]:
+    """Returns a list of axis dicts with name and type"""
+    axes_dicts = []
+    for axis in axes:
+        if isinstance(axis, str):
+            axis_dict = {"name": axis}
+            if axis in KNOWN_AXES:
+                axis_dict["type"] = KNOWN_AXES[axis]
```
sbesson, line 28: What should happen otherwise? Error? default type?

ome_zarr/reader.py
```diff
@@ -282,9 +282,9 @@ def __init__(self, node: Node) -> None:
                 "version", "0.1"
             )  # should this be matched with Format.version?
             datasets = multiscales[0]["datasets"]
-            # axes field was introduced in 0.3, before all data was 5d
-            axes = tuple(multiscales[0].get("axes", ["t", "c", "z", "y", "x"]))
-            if len(set(axes) - axes_values) > 0:
+            axes = multiscales[0].get("axes")
+            if version == "0.3" and len(set(axes) - axes_values) > 0:
+                # TODO: validate axis for > V0.3 ?
```
sbesson, line 287: We can probably define some scenarios that should be validated e.g. there are not enough spatial axes.

Can the `_validate_axes_types` method be re-used here?

ome_zarr/writer.py
will-moore: The types are subsequently checked in `_validate_axes_types()`.
Current spec 0.4 PR says `The "axes" MUST contain 2 or 3 entries of "type:space" and MAY contain one additional entry of "type:time" and MAY contain one additional entry of "type:channel" or a null / custom type.`
So a single missing type may not be invalid.

sbesson: but you proposed it should be a requirement in https://github.com/ome/ngff/pull/57/files#r772303584 no ?

ome_zarr/reader.py
will-moore: I'll reuse the full `_validate_axes()` to do validation for any version

will-moore: Done in bd16dfd

ome_zarr/writer.py
will-moore: Ah, yes - good catch! Although that statement is not quite right since we do allow 1 axis of unknown type.
It's probably better to have `type` as optional if we're going to allow it to be `null`.
Again, this PR really needs to have the spec PR to be concrete first.
I'll add a comment there...

will-moore: I updated the statement, and will fix the code here to use  `axis_dict["type"] = None`...

will-moore: Having tried to implement the `axis_dict["type"] = None` option, this breaks `mypy` in a lot of places, and means the `axes` dict is type `List[Dict[str, Optional[str]]]` everywhere, which means that `axes["name"]` can also be `None`. It doesn't make sense for "type" to be required when it can be `None`. Updated comment on PR to reflect that.

will-moore: To summarise, the code here doesn't need to change.

ome_zarr/writer.py
will-moore: Created an Axes class in 9f75ae3

ome_zarr/axes.py
```diff
@@ -0,0 +1,100 @@
...
+    def validate(self, fmt: Format) -> None:
+
+        # check names (only enforced for version 0.3)
+        if fmt.version == "0.3":
+            self._validate_axes_03()
+            return
+
+        self._validate_axes_types()
+
+    def get_axes(self, fmt: Format) -> Union[List[str], List[Dict[str, str]]]:
```
sbesson, line 23: Minor. `Axes.get_axes()` reads a bit redundant/confusing. If the output was a string, I would suggest to use Python built-in `__repr__` . I don't know if we have something that would clarify what this API does. `tolist()`?

ome_zarr/reader.py
```diff
@@ -275,17 +277,16 @@ def matches(zarr: ZarrLocation) -> bool:
...
             )  # should this be matched with Format.version?
             datasets = multiscales[0]["datasets"]
-            # axes field was introduced in 0.3, before all data was 5d
-            axes = tuple(multiscales[0].get("axes", ["t", "c", "z", "y", "x"]))
-            if len(set(axes) - axes_values) > 0:
-                raise RuntimeError(f"Invalid axes names: {set(axes) - axes_values}")
+            axes = multiscales[0].get("axes")
+            fmt = format_from_version(version)
+            # Raises ValueError if not valid
+            validate_axes(None, axes, fmt)
```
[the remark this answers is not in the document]
will-moore: Ah - good point. Looking at this again, I realise that `validate_axes(None, axes, fmt)` is really designed for writing. So, what's returned will be valid, even if what it's passed isn't valid. E.g. it will convert "tczyx" to an axis array, and will allow v0.3 to have `None` if 2D or 5D.
What we really want here is `Axes(axes, fmt).validate()`

ome_zarr/writer.py
```diff
@@ -160,7 +145,7 @@ def write_multiscales_metadata(
     group: zarr.Group,
     paths: List[str],
     fmt: Format = CurrentFormat(),
-    axes: List[str] = None,
+    axes: Union[str, List[str], List[Dict[str, str]]] = None,
```
sbesson, line 148: https://github.com/ome/omero-cli-zarr/pull/93/files#diff-8ebd9fed3089ce55b4b0d98840330514c910c84da182036847677d2edb02ed54R333 also introduces `transformations` as an extra keyword argument to the equivalent method in `omero-cli-zarr`. Is that something we should cater for? Or not applicable to this PR?

ome_zarr/writer.py
```diff
@@ -2,29 +2,32 @@
...
 from .types import JSONDict
 
 LOGGER = logging.getLogger("ome_zarr.writer")
 
 
-def _validate_axes_names(
-    ndim: int, axes: Union[str, List[str]] = None, fmt: Format = CurrentFormat()
-) -> Union[None, List[str]]:
-    """Returns validated list of axes names or raise exception if invalid"""
+def validate_axes(
```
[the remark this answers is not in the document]
sbesson: What's the outcome of https://github.com/ome/ome-zarr-py/pull/124#discussion_r782875057. Should the name of the method be updated to reflect this is a writer/constructor rather than a validator? Since this API is moved to be a public API, this is increasingly important. Alternatively, we can keep it prefixed with `_` for now.

ome_zarr/reader.py
```diff
@@ -275,17 +277,16 @@ def matches(zarr: ZarrLocation) -> bool:
...
             )  # should this be matched with Format.version?
             datasets = multiscales[0]["datasets"]
-            # axes field was introduced in 0.3, before all data was 5d
-            axes = tuple(multiscales[0].get("axes", ["t", "c", "z", "y", "x"]))
-            if len(set(axes) - axes_values) > 0:
-                raise RuntimeError(f"Invalid axes names: {set(axes) - axes_values}")
+            axes = multiscales[0].get("axes")
+            fmt = format_from_version(version)
+            # Raises ValueError if not valid
+            Axes(axes, fmt).validate()
```
sbesson, line 289: Someway related to https://github.com/ome/ome-zarr-py/pull/124#discussion_r782875057, are there implications to having `validate()` part of the `Axes` constructor?

sbesson (commented):
A few additional comments mostly revolving around API names and behaviors. But I think the bulk of the work looks good

As discussed this morning, we will need to get this consumed by downstream clients e.g. `omero-cli-zarr` and/or `napari-ome-zarr`. To respect the fact the 0.4 specification is not yet published, I don't think this can land into a public release of ome-zarr-py. Instead, can we try to make the minimal decision to target a 0.3 pre-release?

ome_zarr/axes.py
```diff
@@ -0,0 +1,110 @@
...
+    def _get_names(self) -> List[str]:
+        """Returns a list of axis names"""
+        axes_names = []
+        for axis in self.axes:
+            if "name" not in axis:
+                raise ValueError("Axis Dict %s has no 'name'" % axis)
+            axes_names.append(axis["name"])
+        return axes_names
+
+    def _validate_03(self) -> None:
```
joshmoore, line 89: I need to read more here to get my head around the new class, but my immediate reaction to methods that are version specific is that I'd hope they could live in the `Format` object.

joshmoore (commented):
Generally looks good. One minor detail question below and one higher-level here: Does anyone have a feeling for when we might want to start adding submodules, e.g. are we approaching a "util" or "internals" with his PR?

ome_zarr/axes.py
will-moore: I'm trying to think this through... It's possible for everything to become version specific if it changes in a future version, and we wouldn't want the API to change when methods move to a `Format` object.
Everything in the `Axes` class is version specific, (e.g. `_validate_axes_types` will soon become more liberal). So do we move those internal methods to `Format` objects when they change?
This will mean that eventually all the code for everything will migrate `Format` objects.
Are the methods of `Format` objects (e.g. `validate_axes()`) expected to be part of the external API?
What's the use-case or benefit of all version-specific code being in `Format` objects?

ome_zarr/reader.py
will-moore: Done in 3037b33

ome_zarr/writer.py
will-moore: It's not "axes" specific, but I think it still makes sense to include in this PR since it's all "v0.4"

will-moore: Added in 5607786

ome_zarr/axes.py
sbesson: Trying to think of another example with the 0.4 HCS changes to the `wells` element discussed in https://github.com/ome/ome-zarr-py/pull/157, I could certainly imagine migrating `_validate_well_images(wells)` by `fmt.validate_well_images(wells)` or similar and let `FormatV04` vs `FormatV{01,02,03}` override the methods as appropriate.

sbesson:
As discussed yesterday, merging to move forward towards a pre-release of `ome-zarr-py 0.3.0` and we can capture some of the outstanding comments as issues.

PR #142, opened 2021-12-02
Initial validate support, with --strict option

will-moore:
Work in progress...

I am testing the JSON schemas from ngff repo, downloading them to a cache with https://github.com/allenai/cached_path.
This PR adds a `validate` command that validates the JSON at each Node (currently only working for Multiscales spec).

NB: some validation occurs during the init of the Nodes e.g. axes matching xyzct etc. so if that fails then we don't even get to validate against the schemas.

We use `--strict` flag to use `strict_` schemas to show warnings where strict rules are not met.

cc @glyg 

```
$ ome_zarr validate https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001247.zarr --strict
WARNING:ome_zarr.io:version mismatch: detected:FormatV01, requested:FormatV03
WARNING:ome_zarr.io:version mismatch: detected:FormatV03, requested:FormatV01
WARNING:ome_zarr.io:version mismatch: detected:FormatV01, requested:FormatV03
WARNING:ome_zarr.io:version mismatch: detected:FormatV03, requested:FormatV01
Validating Multiscales spec at https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001247.zarr/ [zgroup]
Using Multiscales schema version 0.1
WARNING 'name' is a required property
WARNING 'type' is a required property
WARNING 'metadata' is a required property
Validating Multiscales spec at https://s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001247.zarr/labels/masks/ [zgroup]
Using Multiscales schema version 0.1
WARNING 'name' is a required property
WARNING 'type' is a required property
WARNING 'metadata' is a required property
```


glyg:
Hi @will-moore thanks for the ping. I'm waiting a bit to advance on the rest of the validation, mainly because I'm not sure of what will be left to validate once you're done :)
But maybe I'm wrong and the ticked list [here](https://github.com/ome/ngff/pull/69) is all there is?

If so I'll try to get that started soon

will-moore:
@glyg I wouldn't claim that the list there is complete - It's just what I picked up from the spec, looking for `MUST` rules. But not every detail of the spec is spelt out (it's possible to imagine a file that passed all those rules and was still invalid for some other reason).
However, it's also likely that some of those unchecked items are already covered by `ome-zarr-py`. E.g. some axes validation at https://github.com/ome/ome-zarr-py/blob/a3f50286bcaebfd1d6f03f8e82c6cb9ee0310393/ome_zarr/reader.py#L287 (when you try to parse any zarr image) and https://github.com/ome/ome-zarr-py/blob/a3f50286bcaebfd1d6f03f8e82c6cb9ee0310393/ome_zarr/writer.py#L17 (currently only used when writing an OME-Zarr).

@sbesson The method I've used here (adding more restrictions to the schema via merging python dicts) seems to work so far and could be extended for various levels. It's still limited by the JSON schema validation, so the logic can't get too complex, but it seems worth exploring in more depth.

will-moore:
@sbesson I've updated the approach to match the `allOf` method of combining schemas adopted at https://github.com/ome/ngff/pull/76

Re: required, optional and recommended: The spec uses `MUST` (required), `SHOULD` (recommended) and `MAY` (optional) rules. I don't think this is tool/application specific.

ome_zarr/reader.py
```diff
@@ -315,6 +328,30 @@ def array(self, resolution: str, version: str) -> da.core.Array:
...
+        multiscales = self.lookup("multiscales", [])
+        version = multiscales[0].get("version", "0.1")
+        print("Validating Multiscales spec at", self.zarr)
+        print("Using Multiscales schema version", version)
+        image_schema = get_schema(version)
+
+        # Always do a validation with the MUST rules
+        # Will throw ValidationException if it fails
+        json_data = self.zarr.root_attrs
+        validate(instance=json_data, schema=image_schema)
```
sbesson, line 341: Minor comment: this is confusing to read as this is called within a `validate` method. Potentially `jsonschema.validate` would clear any ambiguity on the scope.

sbesson (changes requested):
Overall this is a substantial step forward in the implementation of a first validation API consuming the newly introduced JSON schemas.  The strategy of using an `allOf` schema allows to validate recommended attributes in addition to the mandatory ones (covered by the default JSON schema). Interested to hear alternate implementation suggestons but from my side, this feels like a solid approach to encode and validate the standard MUST/SHOULD/MAY semantics.

Several suggestions/comments made inline on the implementation/packaging/API. The `.DS_Store` files also need to be removed (and added to `.gitignore`)

Probably beyond the scope of this PR, capturing a few thoughts about next steps:
- should minimal validation be opt-in by default (leaving strict validation as an option)? Assuming the main drawback to doing so would be a performance impact, I suspect this will need to be profiled e.g. using  complex datasets (HCS) stored remotely and look into the time spent validating vs performing other logic. 
- should validation also be added to the writing code?

schemas/0.1/schemas/json_schema/image.schema
```diff
@@ -0,0 +1,112 @@
+{
```
joshmoore, line 1: Would be good to discuss _not_ including these in this repo.

sbesson: I almost commented on whether `ngff_schemas` could be pip installable but since you've opened the door... ;)

ome_zarr/cli.py
```diff
@@ -99,6 +106,12 @@ def main(args: List[str] = None) -> None:
     parser_info.add_argument("--stats", action="store_true")
     parser_info.set_defaults(func=info)
 
+    # validate
+    parser_validate = subparsers.add_parser("validate")
+    parser_validate.add_argument("path")
+    parser_validate.add_argument("--warnings", action="store_true")
```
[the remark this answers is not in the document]
will-moore: yes. We have various terminology: There are Spec rules that `SHOULD` be followed (so it's `recommended` that you do). These are covered by a `strict schema` and if you don't then this gives you `warnings` (if you use that flag)!
It would be nice to settle on ONE term that could cover all these cases?

ome_zarr/reader.py
```diff
@@ -315,6 +328,30 @@ def array(self, resolution: str, version: str) -> da.core.Array:
         # data.shape is (t, c, z, y, x) by convention
         return self.zarr.load(resolution)
 
+    def validate(self, warnings: bool = False) -> None:
+        multiscales = self.lookup("multiscales", [])
+        version = multiscales[0].get("version", "0.1")
```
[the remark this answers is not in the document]
will-moore: Fixed in f0683a6

ome_zarr/utils.py
```diff
@@ -46,6 +46,26 @@ def info(path: str, stats: bool = False) -> Iterator[Node]:
...
+    assert zarr, f"not a zarr: {zarr}"
+    reader = Reader(zarr)
+    for node in reader():
+        if not node.specs:
+            print(f"not an ome-zarr node: {node}")
+            continue
+
+        if hasattr(node, "validate"):
+            node.validate(warnings)
+        yield node
```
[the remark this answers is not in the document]
will-moore: So, if you just do this: does it `report` by default (and validate)?
```
$ ome_zarr info image.zarr
```
Then you can disable both with: ?
```
$ ome_zarr info image.zarr --no-validate --no-report
```

schemas/0.1/schemas/json_schema/image.schema
will-moore: Sorry, yes - the intention was not to include them, but just put them there to get this working.
Do we just add a download_schema command somewhere that is run on the fly if the schemas are missing, and can also be run when we package for release?

ome_zarr/reader.py
will-moore: Fixed in b388a3a

ome_zarr/schemas.py
```diff
@@ -0,0 +1,43 @@
...
+    curr_dir = os.path.dirname(__file__)
+    # The paths here match the paths in the ngff repo (and public schemas)
+    path = os.path.join(
+        curr_dir, f"../schemas/{version}/schemas/json_schema/image.schema"
+    )
+    path = os.path.normpath(path)
+    return load_json(path)
+
+
+def get_strict_schema(version: str) -> Dict:
```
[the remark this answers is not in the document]
will-moore: Fixed in 4c28024

ome_zarr/schemas.py
```diff
@@ -0,0 +1,43 @@
...
+import os
+from typing import Dict
+
+from jsonschema import RefResolver
+
+
+class LocalRefResolver(RefResolver):
+    def resolve_remote(self, url: str) -> Dict:
+        # Use remote URL to generate local path
+        rel_path = url.replace("https://ngff.openmicroscopy.org/", "../schemas/")
```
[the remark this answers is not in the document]
will-moore: Fixed in 33298bd

schemas/0.1/schemas/json_schema/image.schema
will-moore: I don't think `ngff_schemas` should be pip installable, unless we make `ngff` pip installable since it would be nice to have them in the same repo (update the spec at the same time as the schemas).

ome_zarr/cli.py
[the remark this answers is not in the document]
sbesson: Agreed. Trying to look into the terminology used by others, I came back to SHACL which we have not decided to adopt at this stage but we might want to come back to in the near future.

A relevant concept is the idea of severity defined [here](https://www.w3.org/TR/shacl/#results-severity) and used by the [validation report](https://book.validatingrdf.com/bookHtml011.html#sec127). In particular, looking at the [pySHACL](https://github.com/RDFLib/pySHACL) library, the `--allow-warnings` option effectively controls the severity level beyond which the result will be considered as invalid.

schemas/0.1/schemas/json_schema/image.schema
will-moore: Making ngff pip-installable at https://github.com/ome/ngff/pull/77 and using that below in this PR...

will-moore:
This will fail checks until https://github.com/ome/ngff/pull/77 is merged and released on pypi so that `ngff` is pip installable (now a requirement of this PR).

ome_zarr/reader.py
```diff
@@ -8,7 +8,11 @@
 import dask.array as da
 import numpy as np
 from dask import delayed
+from jsonschema import validate as jsonschema_validate
+from jsonschema.validators import validator_for
+from ngff.schemas import LocalRefResolver, get_schema
```
sbesson, line 13: As mentioned earlier today, if we go for an upstream packages containing the schemas as a resource, my vote goes towards prefixing the package/module with `ome` i.e. `ome_ngff.schemas`

setup.py
```diff
@@ -23,6 +23,8 @@ def read(fname):
 install_requires += (["requests"],)
 install_requires += (["scikit-image"],)
 install_requires += (["toolz"],)
+install_requires += (["jsonschema"],)
+install_requires += (["ngff"],)
```
sbesson, line 27: Same as above, I think our project should be prefixed with `ome` i.e.` ome-ngff` inline with the differentiation between NGFFs and OME-NGFF that we settled on in the paper

sbesson (commented):
Two inline comments regarding the prefixing of the package/module.

I am generally in favor of being able to consume schemas from a package rather than copying/pasting and this approach will become increasingly needed as the number of consumer repositories increases and this will help with propagating updates.

Probably my biggest concern is related to the usage of github.com/ome/ngff as a single repository for the specification + the schemas + the samples. Assuming we start covering more languages, we will run into similar issues that we are already facing with github.com/ome/ome-model. That being said, I have not found a good working example of an alternate structure. Happy to hear other feedback including from the community but immediately I see gaining experience with the proposal of https://github.com/ome/ngff/pull/77 while being aware that things might be split in the future.

Although we are still exploring and not 100% clear on the best approach, I assume this strategy can then be easily adapated for other languages (Java/JAR, JS/npm)
- 
- if we start increasing our language coverage

ome_zarr/utils.py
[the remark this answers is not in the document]
constantinpape: Just my 2cents here: I think the cleanest design would be to implement `def visit(path: str, func: callable)` that evaluates `func` for each node. This could then be used to implement both `info` and `validate` without much code duplication, i.e. both `info` and `validate` would just use `visit` with an appropriate function.

This has the added advantage that it allows further customization by users. If you go for this functionality I would also adopt the "usual" syntax for visit in python: If `func` returns something other than `None`, `visit` terminates and returns it.

constantinpape (commented):
I think it would be really valuable to have this in, @will-moore. I guess this is pending https://github.com/ome/ngff/pull/77. Anything else blocking here?

will-moore:
Updated to use the latest naming change to `ome_ngff` from https://github.com/ome/ngff/pull/77

```
$ ome_zarr validate https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.3/idr0079A/9836998.zarr --warnings
WARNING:ome_zarr.io:version mismatch: detected:FormatV03, requested:FormatV04
WARNING:ome_zarr.io:version mismatch: detected:FormatV04, requested:FormatV03
WARNING:ome_zarr.io:version mismatch: detected:FormatV03, requested:FormatV04
WARNING:ome_zarr.io:version mismatch: detected:FormatV04, requested:FormatV03
WARNING:ome_zarr.reader:'metadata' is a required property
WARNING:ome_zarr.reader:'type' is a required property
WARNING:ome_zarr.reader:'name' is a required property
WARNING:ome_zarr.reader:'metadata' is a required property
WARNING:ome_zarr.reader:'type' is a required property
WARNING:ome_zarr.reader:'name' is a required property
```

will-moore:
Added `visit()` as suggested by @constantinpape in [c37e116](https://github.com/ome/ome-zarr-py/pull/142/commits/c37e116f4979ba29f197d3d4ca77e997dd894146)

joshmoore:
@satra mentioned that it would be useful to add validation of Zarrs into https://github.com/dandi/dandi-cli (After I pointed out [this](https://ome.github.io/ome-ngff-validator/?source=https://dandiarchive.s3.amazonaws.com/zarr/3d538281-06d4-48ea-9e8b-72dfe08981a1) :smile:)

Thoughts on getting this into a release?

ome_zarr/reader.py
```diff
@@ -324,6 +337,30 @@ def array(self, resolution: str, version: str) -> da.core.Array:
...
+        # we want to iterate all errors and show as "Warnings"
+        if warnings:
+            strict_schema = get_schema(version, strict=True)
+            cls = validator_for(strict_schema)
+            cls.check_schema(strict_schema)
+            # Use our local resolver subclass to resolve local documents
+            localResolver = LocalRefResolver.from_schema(strict_schema)
+            validator = cls(strict_schema, resolver=localResolver)
+            for error in validator.iter_errors(json_data):
+                LOGGER.warn(error.message)
```
yarikoptic, line 362: sorry, not familiar with the code/design pattern here, can't see if there is some [MVC like pattern](https://en.wikipedia.org/wiki/Model%E2%80%93view%E2%80%93controller) where the warnings&errors (the Model) are collected/returned by some `validate` (the Controller), and then displayed (by the Viewer).
For me, as a 3rd party user intending to use this library to validate it, would be useful to have clear separation of the Model/Controller from the Viewer which would be "mine".

Here https://github.com/dandi/dandi-cli/issues/943#issue-1175862331 we are to "converge" on some data structure which we would use to cover validation reports across multiple validators possibly used for any specific file/dataset at hands.  So it would be important for us to be able to collect errors/warnings/hints uniformly first and then report them via our own "viewer".

will-moore: Thanks for your feedback. This work is at quite an early stage, so we haven't considered the usage of this tool outside of a cli validation. In fact, with changes to the schemas at https://github.com/ome/ngff/tree/main/0.4/schemas (include regular and "strict" schemas), there should be less logic required to validate.
You may find that you can choose the JSON schema validation tool of your choice, and don't need anything from `ome-zarr-py`.
In any case, this PR needs to be reviewed... 
cc @sbesson

will-moore:
@sbesson It looks like https://ngff.openmicroscopy.org/0.4/schemas/plate.schema is `404`, but this looks like the correct URL? Using https://raw.githubusercontent.com/ome/ngff/main/0.4/schemas/plate.schema for now...

will-moore:
With that last commit, renamed `--warnings` to `--strict` and added support for validating Wells and Plates.

```
$ ome_zarr validate https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0056B/7361.zarr --strict
Downloading https://raw.githubusercontent.com/ome/ngff/main/0… ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00 4.2/4.2 kB
Downloading https://raw.githubusercontent.com/ome/ngff/main/0… ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00 568/568 bytes
WARNING:ome_zarr.reader:'name' is a required property
WARNING:ome_zarr.reader:'maximumfieldcount' is a required property
```

Well:
```
$ ome_zarr validate https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0056B/7361.zarr/A/1/ --strict
Downloading https://raw.githubusercontent.com/ome/ngff/main/0… ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00 1.2/1.2 kB
Downloading https://raw.githubusercontent.com/ome/ngff/main/0… ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 100% 0:00:00 309/309 bytes
WARNING:ome_zarr.reader:'version' is a required property
```

sbesson (commented):
Thanks for updating the PR and pushing this important functionality forward.

A few questions:

- the usage of `cached_path` looks like an interesting alternative in the absence of a clear path forward for packaging/distributing thef JSON schemas. What is the library behavior when a live JSON schema is amended? Is the local cached version invalidated and the new version refetched?
- looking at the `get_schema` implementations for each specification, it only needs two elements: 1- the version of the specification, 2- the mapping between `Spec` and schema name. Could this be simplified by populating the version as part of the `__init__` method since the lookup is already made there and storing the schema name as a class constant?
- is there a way to extend the unit tests to add some coverage of the new validation infrastructure?

will-moore:
Thanks: currently, `Multiscales` spec has
```
            self.version = multiscales[0].get(
                "version", "0.1"
```
whereas elsewhere I've been using `CurrentFormat().version` as the default instead of `"0.1"`. OK to update that behaviour for Multiscales to use current version as default?

ome_zarr/reader.py
```diff
@@ -177,6 +204,41 @@ def __init__(self, node: Node) -> None:
...
+            # we only need this store to allow use of cached schemas
+            # (and potential off-line use)
+            schema_store = {
+                schema["$id"]: schema,
+                strict_schema["$id"]: strict_schema,
+            }
+            resolver = RefResolver.from_schema(strict_schema, store=schema_store)
+            validator = Validator(strict_schema, resolver=resolver)
+            for error in validator.iter_errors(json_data):
+                LOGGER.warn(error.message)
```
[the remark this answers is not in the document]
will-moore: As in https://github.com/ome/ome-zarr-py/pull/142#discussion_r905076199 above, I'm thinking of separating the reporting of errors from the validation itself. So, the cli `validate` command can log or print errors, but the `Spec.validate()` and `utils.validate()` would only *return* errors / messages (and maybe log at INFO level)?

It could get tricky to give the user an accurate picture of what failed: eg. when you validate an Image with Labels, the `image` schema is used for both, so you can get an output like this and it's hard to know which object was missing "name":

```
$ ome_zarr validate 6001247.zarr/ --strict --clear_cache
WARNING:ome_zarr.reader:'metadata' is a required property
WARNING:ome_zarr.reader:'type' is a required property
WARNING:ome_zarr.reader:'name' is a required property
WARNING:ome_zarr.reader:'metadata' is a required property
WARNING:ome_zarr.reader:'type' is a required property
```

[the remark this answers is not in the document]
sbesson: In the spirit of `jsonschema.validate` vs `jsonschema.iter_errors`, I concur with the idea of having an API that returns an iterable of all exceptions.

Will the contract be the same for the default and the strict schemas? In the current implementation, the base schema raises an exception but the strict schema does not.
Also I assume since the strict schema extends the base schema, it is not necessary to have 2 validation calls?

will-moore:
@sbesson I added a `--clear_cache` option that removes the `cached_path` directory and forces a fresh download.

yarikoptic:
Is there replacement for this PR or some issue to get announcement whenever validation would become possible?

will-moore:
There is some validation code (using pydantic) at https://github.com/czbiohub/iohub/blob/main/iohub/ngff_meta.py but I haven't tried it myself yet.
It was mentioned in the discussion at https://github.com/ome/ome-zarr-py/issues/258.

You can also use the JSON schemas to validate JSON, e.g.

```
$ pip install check-jsonschema
$ check-jsonschema --schemafile https://raw.githubusercontent.com/ome/ngff/main/0.4/schemas/image.schema 6001240.zarr/.zattrs
ok -- validation done
```

yarikoptic:
Thank you for the pointers -- I will check them out!    Since this PR was closed without any comment left to describe decision to close, a quick follow up question
- no built-in validation interface is being planned/aimed to get?

I am asking since IMHO relying on downstream projects to reimplement standard "specification" for validation is "suboptimal", and validation should go beyond validating just .zattrs.

PR #148, opened 2021-12-10
Fix remaining assumptions on 5D dimensions

sbesson:
Fixes #145 

A few specifications especially in the HCS domain still rely on the expectation that the underlying image will be 5D. As this constraint has been lifted with v0.3, these changes update various places in the code to dynamically identify the x, y and c axis before performing array manipulation.

Without this PR, both `ome_zarr info` and `napari` should fail to read https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.3/idr0094A/7751.zarr is a suitable example for testing this change (plate with XY dimensions). With this PR included, both commands should work successfully.

The last commits also disable `PlateLabels` to fix the HCS display in `napari`.  #177 as well as https://github.com/ome/ome-zarr-py/issues/65#issuecomment-1010074621 might be fixed by this PR

PR #149, opened 2021-12-16
Add new API to write multiscales metadata

sbesson:
While investigating the testing requirements for https://github.com/ome/ome-zarr-py/pull/148, I found that my primary need are API helper methods allowing to write NGFF metadata e.g. `write_plate_metadata`, `write_well_metadata` and generate test data

As a first step towards these APIs, this PR refactors the existing multiscales generation code and extract the `multiscales` metadata addition logic into a separate API `write_multiscales_metadata`.
A direct application of this method is the scenario where a library would have its own implementation of the data writing but would still like to write the metadata using the core API. An example is `omero-cli-zarr` where https://github.com/ome/omero-cli-zarr/blob/7589cc47920b912480b3e6daa282c3d773c2bfdf/src/omero_zarr/raw_pixels.py#L276-L289 could be updated to consume this API.

In addition to the new API, the existing internal axes validation method is split into two, separating the validation of the axes against the array dimensions and the inference of axes values for 2D/5D cases from the strict validation of the axes using the allowed values.

A set of unit tests are also added to cover the different scenarios.

### DIFF
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -103,9 +103,39 @@ def write_multiscale(
     for path, dataset in enumerate(pyramid):
         # TODO: chunks here could be different per layer
         group.create_dataset(str(path), data=dataset, chunks=chunks)
-        paths.append({"path": str(path)})
+        paths.append(str(path))
+    write_multiscales_metadata(group, paths, fmt, axes)
 
-    multiscales = [{"version": fmt.version, "datasets": paths}]
+
+def write_multiscales_metadata(
+    group: zarr.Group,
+    paths: List[str],
+    fmt: Format = CurrentFormat(),
+    axes: Union[str, List[str]] = None,
+) -> None:
+    """
+    Write the multiscales metadata in the group metadata
+
+    Parameters
+    ----------
+    group: zarr.Group
+      the group within the zarr store to store the data in
+    paths: list of str,
+      The list of path to the datasets for this multiscale image
+    fmt: Format
+      The format of the ome_zarr data which should be used.
+      Defaults to the most current.
+    axes: str or list of str
+      the names of the axes. e.g. "tczyx". Not needed for v0.1 or v0.2
+      or for v0.3 if 2D or 5D. Otherwise this must be provided
+    """
+
+    multiscales = [
+        {
+            "version": fmt.version,
+            "datasets": [{"path": str(p)} for p in paths],
+        }
+    ]
     if axes is not None:
         multiscales[0]["axes"] = axes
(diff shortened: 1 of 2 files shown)


will-moore:
Looks good. I don't know if I'd guess the different behaviour of `_validate_axes()` and `_validate_axes_names()` from their names, but naming things is hard.
All of this will change soon for v0.4 anyway. I imagined that we might use `_validate_axes(axes: List[Dict])` to validate axes objects (in v0.4) whereas validation of their names with `_validate_axes_names(axes: List[str])`, but I don't know if it will make sense to separate the logic in the same way for v0.4...?

sbesson:
@will-moore thanks for picking that up. Yes the name is not the greatest and I was also pondering how this would evolve with the upcoming axes specification.

As we decided to keep this API private, we can certainly refine names etc. At least for 0.3, I see two different behaviors:

- validate the value of the axes (i.e. is a subset of TCZYX)
- validate the axes in relationship to some data i.e. are the dimensions consistent
- infer unspecified axes in some scenarios e.g. for 2D or 5D

I think the spirit of https://github.com/ome/ome-zarr-py/pull/149/commits/0c71fbe7f3375bc393c8c9a77afbc15f7256d7ae is to separate these different functionalities.

How do you expect the changes in 0.4 will affect this API? One thought is that the axis guessing based on the dimensionality might become irrelevant. But are the other behaviors that different? Or is it only the input type which will primarily vary?

will-moore:
Yes, as this is a private API, it's OK to go with what we have here and update for 0.4.
I think for v0.4 we could still guess axes for 2D, maybe not for 5D.
The input type will certainly vary, at least for `_validate_axes()` since axes will be Dicts.

joshmoore (approved):

sbesson:
Thanks both. Unless you can feel a particular need for a patch release, I'd propose to get this into the main development branch and bump the version to `0.3.0.dev0`.

I have started working on similar APIs for plate/well metadata. My goal is to help to construct the unit tests allowing to test https://github.com/ome/ome-zarr-py/pull/148 and build the framework for the support of https://github.com/ome/ngff/pull/24.

Said otherwise, I am starting to see a `0.3.0` roadmap emerging with a Jan 2022 timeline including:
1. new metadata writing API (multiscales, plate, well)
2. the fix for HCS with <5D data
3. anything else?

jburel:
Since the ome_zarr_test_suite is triggered on merge, we should wait for the test suite to finish before tagging in case there is an issue. 
It is green see https://github.com/ome/ome_zarr_test_suite/actions/runs/1594165903 but we never know

Issue #145, opened 2021-12-07
HCS stitching of Wells to Plate broken for data less than 5D


will-moore:
The concatenation of Wells into a Plate assumes 5D images.
This is not always the case with v0.3, since this version supports 2D -> 5D

e.g. https://uk1s3.embassy.ebi.ac.uk/0.3_samples/7751.zarr 
See this [plate in vizarr](https://hms-dbmi.github.io/vizarr/?source=https://uk1s3.embassy.ebi.ac.uk/0.3_samples/7751.zarr)
cc @sbesson

sbesson:
Transferred the issue from `napari-ome-zarr` as running  `ome_zarr info` on the plates reveals the same issue

PR #153, opened 2022-01-05
Add API for writing HCS metadata

sbesson:
Follow-up of #149, this adds two new top-level APIs to the `ome_zarr.writer` module: `write_plate_metadata` and `write_well_metadata` for creating [plate](https://ngff.openmicroscopy.org/latest/#plate-md) and [well](https://ngff.openmicroscopy.org/latest/#well-md) specifications.

Unit tests are added to cover the combinations of parameters both in writing mode (`test_writer`) as well as in writing/reading mode (`test_node/test_reader`). This should significantly increase the coverage of the HCS codepath especially in `reader.py`.

The `test_plate_2D5D` unit test reproduces the issue raised in #145 and is currently marked as `xfail`. As the bug should be addressed in #148, I would propose to prioritize the review and integration of this PR first and remove the test marker in #148 together with the fix.

Thinking of the upcoming sepcification changes, the biggest potential implication for this new API are the changes proposed in https://github.com/ome/ngff/pull/24 to the `wells` attribute of the `plate` specification with the addition of `rowIndex` and `columnIndex` attributes.  A proposal would be to preemptively change type of the `wells` argument to `List[dict]` and add an internal API similar to `_validate_plate_acquisitions` to inspect the required fields conditionally to the format version.

### DIFF
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -2,7 +2,7 @@
 
 """
 import logging
-from typing import Any, List, Tuple, Union
+from typing import Any, Dict, List, Tuple, Union
 
 import numpy as np
 import zarr
@@ -68,6 +68,53 @@ def _validate_axes(axes: List[str], fmt: Format = CurrentFormat()) -> None:
     else:
         if val_axes != ("t", "c", "z", "y", "x"):
             raise ValueError("5D data must have axes ('t', 'c', 'z', 'y', 'x')")
+
+
+def _validate_well_images(images: List, fmt: Format = CurrentFormat()) -> None:
+
+    VALID_KEYS = [
+        "acquisition",
+        "path",
+    ]
+    for index, image in enumerate(images):
+        if isinstance(image, str):
+            images[index] = {"path": str(image)}
+        elif isinstance(image, dict):
+            if not all(e in VALID_KEYS for e in image.keys()):
+                raise ValueError(f"{image} contains invalid keys")
+            if "path" not in image:
+                raise ValueError(f"{image} must contain a path key")
+            if not isinstance(image["path"], str):
+                raise ValueError(f"{image} path must be of string type")
+            if "acquisition" in image and not isinstance(image["acquisition"], int):
+                raise ValueError(f"{image} acquisition must be of int type")
+        else:
+            raise ValueError(f"Unrecognized type for {image}")
+
+
+def _validate_plate_acquisitions(
+    acquisitions: List[Dict], fmt: Format = CurrentFormat()
+) -> None:
+
+    VALID_KEYS = [
+        "id",
+        "name",
+        "maximumfieldcount",
+        "description",
+        "starttime",
+        "endtime",
+    ]
+    if acquisitions is None:
+        return
+    for acquisition in acquisitions:
+        if not isinstance(acquisition, dict):
+            raise ValueError(f"{acquisition} must be a dictionary")
+        if not all(e in VALID_KEYS for e in acquisition.keys()):
+            raise ValueError(f"{acquisition} contains invalid keys")
+        if "id" not in acquisition:
+            raise ValueError(f"{acquisition} must contain an id key")
+        if not isinstance(acquisition["id"], int):
+            raise ValueError(f"{acquisition} id must be of int type")
 
 
 def write_multiscale(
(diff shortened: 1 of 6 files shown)


sbesson:
After some quick investigation of https://github.com/ome/ome-zarr-py/runs/4714215369?check_suite_focus=true, the failure is unrelated to the Zarr development branch but the tests were simply failing on Ubuntu.

Reproducing them in a Docker environment, I found the issue comes from the key normalization which leads to the creation of Zarr group `a` for row `A` causing an obvious mismatch between the hierarchy and the `.zattrs` metadata. f468a0a, which I also found to be necessary in https://github.com/ome/ome-zarr-py/pull/148, addresses this issue. 
d1311aa also re-enables Ubuntu tests without the Zarr development branch.

ome_zarr/writer.py
```diff
@@ -70,6 +70,53 @@ def _validate_axes(axes: List[str], fmt: Format = CurrentFormat()) -> None:
...
+    VALID_KEYS = [
+        "acquisition",
+        "path",
+    ]
+    for index, image in enumerate(images):
+        if isinstance(image, str):
+            images[index] = {"path": str(image)}
+        elif isinstance(image, dict):
+            if not all(e in VALID_KEYS for e in image.keys()):
+                raise ValueError(f"{image} contains invalid keys")
```
joshmoore, line 84: The spec has been updated with a "MUST NOT" have other keys? Eventually, that could be problematic. Perhaps the prefix mechanism that @will-moore found in json-schema that leaves them unchecked could eventually be introduced here.

joshmoore (approved):
:+1: but one "future warning" so to speak.

ome_zarr/writer.py
sbesson: Thanks for picking up on this. I have not found a clear statement regarding the validity of additional keys in the ngff spec. This strict implementation is probably derived from my reading of `multiscales` where I assume `metadata` is the single key that tried to capture extra arguments.

That being said, I totally see this MUST NOT interpretation is and will be limiting both for the extension of the specification itself as well as for supporting external metadata not covered by the spec. 

The `well` metadata is a perfect example as https://github.com/ome/ngff/pull/24 defines more keys. Assuming someone wanted to write a version 0.3 with these new keys populated, is it legit and desired to let the writer implementation write this metadata? If so, is it worth a logging statement at WARN, INFO or DEBUG level? Or should this writer only care about the keys defined in the spec and ignore anything extra?

will-moore: I think it was `salad schema` where you could use a `prefix:` to avoid validation, but I *think* for json-schema you don't need to do that and it won't fail for unrecognised attributes.

sbesson:
Pushed 12b341d in response to the concerns expressed in https://github.com/ome/ome-zarr-py/pull/153#discussion_r780679873 about unspecified keys. With this commit, the writer handles leniently unspecified keys are present in lists of dictionaries (acquisitions, images). The unit tests are adjusted to ensure the group creation is successful with all keys been written.

If we agree with this approach, I think it makes sense to implement immediately the proposal made at the end of this PR description i.e. update the `write_plate_metadata` signature to support `wells` either as `Union[List[str], List[dict]]` or as `List[dict]` and handle unspecified keys similarly to `acquisitions` and `images`.

will-moore:
@sbesson That commit 12b341d replaces `not all` with `not any` in one case and `any` in the other. Is that right?

sbesson:
Good catch. ff197a0 should unify both behaviors. Since the behavior is a simple logging statement, that's not something that the unit tests picked up

will-moore:
@sbesson Thanks PR looks good. Updating `write_plate_metadata` to support `List[Dict]` makes sense. In this PR?

sbesson:
> In this PR?

Originally, that's what I thought but with different fires coming up, I feel this will get pushed. If you and @joshmoore are happy with the state of the latest proposed API, I'd propose to get this merged and have the new tests as well as the re-activated Ubuntu tests merged into the mainline. I'll handle the new wells type handling in a follow-up PR which should be self-contained and easier to review

ome_zarr/writer.py
```diff
@@ -147,6 +194,85 @@ def write_multiscales_metadata(
     group.attrs["multiscales"] = multiscales
 
 
+def write_plate_metadata(
+    group: zarr.Group,
+    rows: List[str],
+    columns: List[str],
+    wells: List[str],
+    fmt: Format = CurrentFormat(),
```
jburel, line 202: Why not use ``format`` since name is explicit for other fields?

sbesson: Primarily because `fmt` is the nomenclature used in all the existing APIs in the library. Do you want to capture the rename as an issue?

jburel: https://github.com/ome/ome-zarr-py/issues/156

PR #157, opened 2022-01-13
Add support for passing wells as List[dict] in write_plate_metadata

sbesson:
Follow-up of https://github.com/ome/ome-zarr-py/pull/153, this adds support for wells specified as lists of dictionaries rather than lists of string. The primary use case are:

1- support additional custom metadata e.g.

```
wells = [
    {"path": "A/1", "concentration": "1"},
    {"path": "A/2", "concentration": "2"},
    {"path": "B/1", "concentration": "5"},
]
write_plate_metadata(self.root, ["A", "B"], ["1", "2"], wells)
```

2- support the upcoming breaking changes proposed in https://github.com/ome/ngff/pull/24 i.e.

```
wells = [
    {"path": "A/1", "rowIndex": 0, "columnIndex": 0},
    {"path": "A/2", "rowIndex": 0, "columnIndex": 1},
    {"path": "B/1", "rowIndex": 1, "columnIndex": 0},
]
write_plate_metadata(self.root, ["A", "B"], ["1", "2"], wells)
```

Unit tests are updated accordingly to cover all variants. While working on the tests, I found that the usage of the list mutability by the `_validate*` API was creating unnecessary confusion when the input argument is re-used. 347da426df98db2a9653d9bdbf17e6b8bbc00377 updates the logic to create a copy of the input list when the content might be modified by the validation function.

Since this API will likely be released with 0.4 support and given the changes proposed in https://github.com/ome/ngff/pull/24, it is likely the former support for wells as `List[str]` will:
 1- either need to be removed as  a list of paths will not be sufficient to generate valid `wells`
 2- or be updated to include the assumptions that the wells are following the standard letter/number convention and auto-generate the underlying `rowIndex`/`columnIndex` indices.
I left this discussion outside the scope of this PR which only focuses on supporting the `List[dict]` form.

### DIFF
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -104,8 +104,7 @@ def _validate_plate_acquisitions(
         "starttime",
         "endtime",
     ]
-    if acquisitions is None:
-        return
+
     for acquisition in acquisitions:
         if not isinstance(acquisition, dict):
             raise ValueError(f"{acquisition} must be a dictionary")
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -126,7 +126,7 @@ def _validate_plate_wells(wells: List, fmt: Format = CurrentFormat()) -> List[di
         "path",
     ]
     validated_wells = []
-    if wells is None:
+    if wells is None or len(wells) == 0:
         raise ValueError("Empty wells list")
     for well in wells:
         if isinstance(well, str):
@@ -138,7 +138,7 @@ def _validate_plate_wells(wells: List, fmt: Format = CurrentFormat()) -> List[di
                 raise ValueError(f"{well} must contain an path key")
             if not isinstance(well["path"], str):
                 raise ValueError(f"{well} path must be of str type")
-            validated_wells.append({"path": str(well)})
+            validated_wells.append(well)
         else:
             raise ValueError(f"Unrecognized type for {well}")
     return validated_wells
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -120,6 +120,30 @@ def _validate_plate_acquisitions(
     return acquisitions
 
 
+def _validate_plate_wells(wells: List, fmt: Format = CurrentFormat()) -> List[dict]:
+
+    VALID_KEYS = [
+        "path",
+    ]
+    validated_wells = []
+    if wells is None:
+        raise ValueError("Empty wells list")
+    for well in wells:
+        if isinstance(well, str):
+            validated_wells.append({"path": str(well)})
+        elif isinstance(well, dict):
+            if any(e not in VALID_KEYS for e in well.keys()):
+                LOGGER.debug("f{well} contains unspecified keys")
+            if "path" not in well:
+                raise ValueError(f"{well} must contain an path key")
+            if not isinstance(well["path"], str):
+                raise ValueError(f"{well} path must be of str type")
+            validated_wells.append({"path": str(well)})
+        else:
+            raise ValueError(f"Unrecognized type for {well}")
+    return validated_wells
+
+
 def write_multiscale(
     pyramid: List,
     group: zarr.Group,
@@ -201,7 +225,7 @@ def write_plate_metadata(
     group: zarr.Group,
     rows: List[str],
     columns: List[str],
-    wells: List[str],
+    wells: Union[List[str], List[dict]],
     fmt: Format = CurrentFormat(),
     acquisitions: List[dict] = None,
     field_count: int = None,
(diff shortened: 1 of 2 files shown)


ome_zarr/writer.py
```diff
@@ -88,13 +89,15 @@ def _validate_well_images(images: List, fmt: Format = CurrentFormat()) -> None:
                 raise ValueError(f"{image} path must be of string type")
             if "acquisition" in image and not isinstance(image["acquisition"], int):
                 raise ValueError(f"{image} acquisition must be of int type")
+            validated_images.append(image)
```
joshmoore, line 92: Not really related to this PR, but this PR did make me notice that here we throw on the first error, rather than finding all validation errors and listing them.

ome_zarr/writer.py
```diff
@@ -198,7 +225,7 @@ def write_plate_metadata(
     group: zarr.Group,
     rows: List[str],
     columns: List[str],
-    wells: List[str],
+    wells: Union[List[str], List[dict]],
```
joshmoore, line 228: This could also be `List[Union[str, dict]]` eh?

joshmoore (commented):
Barring the same discussion around possible prefixes for unparsed keys, :+1:

ome_zarr/writer.py
sbesson: Heh pedantic but actually correct and this reflects more the current implementation which support of list of mixed type - pushed 0ed8b0b  to unify

ome_zarr/writer.py
sbesson: Agreed, this is a fail-fast approach. Worth capturing as an RFE?

joshmoore: https://github.com/ome/ome-zarr-py/issues/158

ome_zarr/writer.py
```diff
@@ -115,6 +119,33 @@ def _validate_plate_acquisitions(
...
+    if wells is None or len(wells) == 0:
+        raise ValueError("Empty wells list")
+    for well in wells:
+        if isinstance(well, str):
+            validated_wells.append({"path": str(well)})
+        elif isinstance(well, dict):
+            if any(e not in VALID_KEYS for e in well.keys()):
+                LOGGER.debug("f{well} contains unspecified keys")
+            if "path" not in well:
+                raise ValueError(f"{well} must contain an path key")
```
will-moore, line 142: typo `an` -> `a`

will-moore:
Looks good.
With the breaking spec change, I think the API support for list of strings should be removed, since the logic for guessing the correct row/col indices will be tricky and not easy to document well so users will understand (potential for getting it wrong).

sbesson:
Re list of strings, my consideration was to limit the scope of supporting conventions and only consider the single standard scenario where row == letter, column == 1-based index i.e. support `wells = ["A/3", "B/2", "B/5"]` and compute the indices. Probably the biggest decision in that case would be to agree on what happens if len(rows) > 26 - see https://www.openmicroscopy.org/Schemas/Documentation/Generated/OME-2016-06/ome_xsd.html#Plate_RowNamingConvention.

But I am okay with leaving it the caller responsibility to generate the row/column indices first if we believe this is unnecessary. We can always add convenience APIs later

will-moore:
I guess that is kinda convenient. And we will also have the `rows` and `columns` names, so we could simply split on `/` and see if they match:
```
r, c = well.split("/")
rowIndex = rowNames.index(r)
colIndex = colNames.index(c)
```
and raise `ValueError` if rowIndex or colIndex is -1. This will also work if `len(rows) > 26`.
If we're happy with this approach, then we don't have to make a breaking change?

sbesson:
Very good point about `rows` and `columns`, it certainly feels more robust than what I was suggesting i.e.
- support `wells` as `List[str]`assuming the well paths can be computed from the `rows` and `columns`
- generically support wells as `List[dict]` in which case the name of the sub-groups can be decoupled from the names of the `rows` and `columns` e.g. could be using UUIDs

Either way, I don't think there is any breaking change involved as the HCS API has not been released yet.

will-moore:
@sbesson do you want to add support for calculating rowIndex and columnIndex like in https://github.com/ome/ome-zarr-py/pull/157#issuecomment-1013183903 in this PR?
Just thinking about using this from `omero-cli-zarr` for export

sbesson:
Yes, as per the discussion above, my intention is to support the following APIs

```
write_plate_metadata(plate_group, ["A", "B"], ["1", "2"], wells=["A/1", "A/2", "B/1", "B/2"])
write_plate_metadata(plate_group, ["A", "B"], ["1", "2"], wells=[{"path": "A/1", "rowIndex": 0, "columnIndex": 0}, ...)
```

I would add support in a follow-up PR as I'd need the `FormatV04` class introduced in #124

PR #159, opened 2022-01-18
Add support for HCS 0.4 specification

sbesson:
Implements the specification changes proposed in https://github.com/ome/ngff/pull/24

- `rows` and `columns` must have unique names and be alphanumerical
- `wells` elements must include `path` (composed of exactly 2 groups), `rowIndex` and `colIndex`

As discussed in #157, the internal logic is adjusted to support the simple case of passing wells as list of strings and compute the row/column indices assuming the paths match the `rows` and `columns` lists.

Unit tests are updated to cover both the new and the legacy behavior as well as various invalid cases. The reader is not adjusted to consume the new logic. I would propose to complete the work on #148 first.

### DIFF
--- a/ome_zarr/format.py
+++ b/ome_zarr/format.py
@@ -72,11 +72,21 @@ class Format(ABC):
     def __eq__(self, other: object) -> bool:
         return self.__class__ == other.__class__
 
+    @abstractmethod
+    def generate_well_dict(self, well: str) -> dict:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def validate_well_dict(self, well: dict) -> None:
+        raise NotImplementedError()
+
 
 class FormatV01(Format):
     """
     Initial format. (2020)
     """
+
+    REQUIRED_PLATE_WELL_KEYS = {"path": str}
 
     @property
     def version(self) -> str:
@@ -92,8 +102,20 @@ class FormatV01(Format):
         LOGGER.debug(f"Created legacy flat FSStore({path}, {mode})")
         return store
 
+    def generate_well_dict(self, well: str) -> dict:
+        return {"path": str(well)}
 
-class FormatV02(Format):
+    def validate_well_dict(self, well: dict) -> None:
+        if any(e not in self.REQUIRED_PLATE_WELL_KEYS for e in well.keys()):
+            LOGGER.debug("f{well} contains unspecified keys")
+        for key, key_type in self.REQUIRED_PLATE_WELL_KEYS.items():
+            if key not in well:
+                raise ValueError(f"{well} must contain a {key} key of type {key_type}")
+            if not isinstance(well[key], key_type):
+                raise ValueError(f"{well} path must be of {key_type} type")
+
+
+class FormatV02(FormatV01):
     """
     Changelog: move to nested storage (April 2021)
     """
--- a/ome_zarr/format.py
+++ b/ome_zarr/format.py
@@ -117,8 +117,6 @@ class FormatV01(Format):
                 raise ValueError(f"{well} must contain a {key} key of type {key_type}")
             if not isinstance(well[key], key_type):
                 raise ValueError(f"{well} path must be of {key_type} type")
-        if len(well["path"].split("/")) != 2:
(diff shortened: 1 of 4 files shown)


ome_zarr/format.py
```diff
@@ -151,9 +179,23 @@ class FormatV04(FormatV03):
...
 
+    def generate_well_dict(
+        self, well: str, rows: List[str], columns: List[str]
+    ) -> dict:
+        row, column = well.split("/")
+        if row not in rows:
+            raise ValueError(f"{row} is not defined in the list of rows")
+        rowIndex = rows.index(row)
+        if column not in columns:
+            raise ValueError(f"{row} is not defined in the list of rows")
```
will-moore, line 196: Error needs to use `{column}` and "columns"

ome_zarr/format.py
```diff
@@ -151,9 +179,23 @@ class FormatV04(FormatV03):
...
+        colIndex = columns.index(column)
+        return {"path": str(well), "rowIndex": rowIndex, "colIndex": colIndex}
```
will-moore, line 198: In the spec this is `columnIndex` not `colIndex`.

ome_zarr/writer.py
```diff
@@ -107,26 +107,38 @@ def _validate_plate_acquisitions(
...
+            fmt.validate_well_dict(well_dict)
+            validated_wells.append(well_dict)
         elif isinstance(well, dict):
-            if any(e not in VALID_KEYS for e in well.keys()):
-                LOGGER.debug("f{well} contains unspecified keys")
-            if "path" not in well:
-                raise ValueError(f"{well} must contain a path key")
-            if not isinstance(well["path"], str):
-                raise ValueError(f"{well} path must be of str type")
+            fmt.validate_well_dict(well)
```
will-moore, line 141: I wonder if you want to validate that `well` against the rows and columns?
E.g. this is invalid, but I think wouldn't fail:
```
_validate_plate_wells([{"path": "Z/10"}], ["A"], ["1"])
```

sbesson: This is probably a point up for discussion. As  per the current specification proposal, what you suggest is _not_ invalid i.e. there is no enforced relationship between the Zarr group names and the rows/columns names. Instead the relationship is purely expressed via the row and column indices.

In the typical usage, I also expect these groups and names should match and this is what the row/column index computation logic assumes.

I'll push another unit test to cover this use case shortly

sbesson: See 16986d0

will-moore: The spec PR says "The `path` MUST consist of a `name` in the `rows` list, a file separator (`/`),
      and a `name` from the `columns` list, in that order."

sbesson: Yes, I just read that and I commented on the spec PR itself - https://github.com/ome/ngff/pull/24#discussion_r597699471.

sbesson:
The last commits add extra validation checks ensuring the consistency of the well `path` against `rows`, `columns`, `rowIndex` and `columnIndex`.

will-moore:
Looks good, thanks.

will-moore:
@seb - I've started using `write_plate_metadata()` in https://github.com/ome/omero-cli-zarr/pull/93/commits/cc44688116add04e1cebe3cbb4cdbf374a4c629e

tests/test_writer.py
```diff
@@ -453,25 +495,88 @@ def test_invalid_well_list(self, wells):
             [{"path": 0}],
             [{"id": "test"}],
             [{"path": "A/1"}, {"path": None}],
+            [{"path": "A/1", "rowIndex": 0}],
+            [{"path": "A/1", "columnIndex": 0}],
+            [{"path": "A/1", "rowIndex": "0", "columnIndex": 0}],
+            [{"path": "A/1", "rowIndex": 0, "columnIndex": "0"}],
+            [{"path": "A1", "rowIndex": 0, "columnIndex": "0"}],
+            [{"path": "plate/A/1", "rowIndex": 0, "columnIndex": 0}],
```
jburel, line 503: should a ``plate/A1`` case be tested too?

PR #161, opened 2022-01-24
add compressor option to write image

### LINKED ISSUE #160: add compressor to write_image
i couldn't see an exposed interface for compressing the chunks. is this intentional? if not, i will send a  PR to alter `write_image` to take a compressor keyword and passing it on to zarr array creation.

satra:
this proposes a change to the write_image api to allow compressors.  this is now more general to pass any options to the create dataset function and should handle multilevel options.

closes #160

- [x] Write some tests

### DIFF
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -186,7 +186,9 @@ def write_multiscale(
     paths = []
     for path, dataset in enumerate(pyramid):
         # TODO: chunks here could be different per layer
-        group.create_dataset(str(path), data=dataset, chunks=chunks, compressor=compressor)
+        group.create_dataset(
+            str(path), data=dataset, chunks=chunks, compressor=compressor
+        )
         paths.append(str(path))
     write_multiscales_metadata(group, paths, fmt, axes, transformations)
 
@@ -388,12 +390,13 @@ def write_image(
         image = [image]
 
     write_multiscale(
-        image, group, 
-        chunks=chunks, 
-        fmt=fmt, 
-        axes=axes, 
-        transformations=transformations, 
-        compressor=compressor
+        image,
+        group,
+        chunks=chunks,
+        fmt=fmt,
+        axes=axes,
+        transformations=transformations,
+        compressor=compressor,
     )
     group.attrs.update(metadata)
 
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -214,13 +214,15 @@ def write_multiscale(
 
     datasets: List[dict] = []
     for path, data in enumerate(pyramid):
-        options = (
-            storage_options
-            if not isinstance(storage_options, list)
-            else storage_options[path]
-        )
-        if "chunks" not in options:
-            options["chunks"] = chunks
+        options = {}
+        if storage_options:
+            options = (
+                storage_options
+                if not isinstance(storage_options, list)
+                else storage_options[path]
+            )
+            if "chunks" not in options:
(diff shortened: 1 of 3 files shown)


joshmoore:
Thanks for this, @satra. Could you sign and return the OME CLA as described in https://ome-contributing.readthedocs.io/en/latest/cla.html?

joshmoore (commented):
Implementation seems simple enough, though we should probably get it integrated into a test.

As a side note, I do remember, however, not being completely happy with the `chunks` argument since conceivably one might want different settings per pyramid level. The same might apply here.

I don't know if you, @satra, or anyone else at this point has a clear alternative that would keep this method cleaner. Options that come to mind include:
 * passing all arguments as key words similar to fsspec's `storage_options`
 * and/or passing a callback function which review is given the path, the dataset, as well as the resolution level

Happy to keep pondering those after this PR though.

sbesson:
Brainstorming on the API questions raised in https://github.com/ome/ome-zarr-py/pull/161#pullrequestreview-862102250, a solution along the lines of `storage_options` feels  appealing to me as this would allow any of the options available in [create_dataset](https://zarr.readthedocs.io/en/stable/api/hierarchy.html#zarr.hierarchy.Group.create_dataset) to be specified via this library.

Do we expect the `dataset` level to be the finer level of granularity for these storage options? In this case, I assume an option would be either to support a simple dictionary (and use it for each dataset) or a list of dictionaries of size matching the size of `pyramid` which would allowing to specify options at the individual dataset level.

satra:
@sbesson - updated this to use `storage_options`.

ome_zarr/writer.py
```diff
@@ -202,15 +203,27 @@ def write_multiscale(
...
-        group.create_dataset(str(path), data=data, chunks=chunks)
+        options = {}
+        if storage_options:
+            options = (
+                storage_options
+                if not isinstance(storage_options, list)
+                else storage_options[path]
+            )
+        if "chunks" not in options:
+            options["chunks"] = chunks
```
sbesson, line 225: This raises the question of whether we would like to deprecate the explicit `chunks` argument and only rely on the `storage_options` dictionary moving forward.

sbesson (changes requested):
Thanks @satra. We are making progress towards `ome-zarr 0.3.0` which should include OME-NGFF 0.4 support. It would be great to make this new storage options available at the same time.

At minimum, could a unit test be added that covers the compression scenario? As a bonus, it looks like your `storage_options` also cover the per-resolution `chunks` feature mentioned in https://github.com/ome/ome-zarr-py/pull/161#pullrequestreview-862102250. I suspect we will want some tests covering this scenario but this can happen as a follow-up PR depending on time.
Also could you return the CLA as mentioned in https://github.com/ome/ome-zarr-py/pull/161#issuecomment-1021047464.

ome_zarr/writer.py
satra: if it's deprecated, i would suggest giving it a couple of cycles and perhaps add a deprecation warning with the next release. but leave this up to you.

sbesson: Fair enough, we can gain knowledge about the `storage_options` in the upcoming 0.3.x and schedule the deprecation of `chunks` for 0.4. I'll create an issue

sbesson (approved):
Thanks for the test @satra. All looks good from my side. Merging and we will aim to get this as a public release in the upcoming days

PR #162, opened 2022-01-25
write_multiscales_metadata no_transformations

will-moore:
As discussed at https://github.com/ome/omero-cli-zarr/pull/93#discussion_r791685798
this allows the 'paths` argument of `write_multiscales_metadata()` to take a list of `datasets` dicts.

EDIT: Also updated to validate latest 0.4 spec with `coordinateTransformations` of `scale` and `translation`.

### DIFF
--- a/ome_zarr/format.py
+++ b/ome_zarr/format.py
@@ -85,10 +85,16 @@ class Format(ABC):
         raise NotImplementedError()
 
     @abstractmethod
+    def generate_coordinate_transformations(
+        self, shapes: List[tuple]
+    ) -> Optional[List[List[Dict[str, Any]]]]:  # pragma: no cover
+        raise NotImplementedError()
+
+    @abstractmethod
     def validate_coordinate_transformations(
         self,
         shapes: List[tuple],
-        coordinateTransformations: List[List[Dict[str, Any]]] = None,
+        coordinate_transformations: List[List[Dict[str, Any]]] = None,
     ) -> Optional[List[List[Dict[str, Any]]]]:  # pragma: no cover
         raise NotImplementedError()
 
@@ -130,11 +136,16 @@ class FormatV01(Format):
             if not isinstance(well[key], key_type):
                 raise ValueError(f"{well} path must be of {key_type} type")
 
+    def generate_coordinate_transformations(
+        self, shapes: List[tuple]
+    ) -> Optional[List[List[Dict[str, Any]]]]:
+        return None
+
     def validate_coordinate_transformations(
         self,
         shapes: List[tuple],
-        coordinateTransformations: List[List[Dict[str, Any]]] = None,
-    ) -> Optional[List[List[Dict[str, Any]]]]:
+        coordinate_transformations: List[List[Dict[str, Any]]] = None,
+    ) -> None:
         return None
 
 
@@ -193,7 +204,7 @@ class FormatV03(FormatV02):  # inherits from V02 to avoid code duplication
 class FormatV04(FormatV03):
     """
     Changelog: axes is list of dicts,
-    introduce coordinateTransformations in multiscales (Nov 2021)
+    introduce coordinate_transformations in multiscales (Nov 2021)
     """
 
     REQUIRED_PLATE_WELL_KEYS = {"path": str, "rowIndex": int, "columnIndex": int}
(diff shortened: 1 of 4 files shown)


sbesson (changes requested):
Overall looks good and inline with the `plate.wells` handling. Can a few unit tests be added to `test_writer` to cover `datasets` as dictionaries as well as the `None,[]` scenarios?

I suspect the next step will be to add some code to validate optional `transformations` elements on write. This can be captured as an issue for now.

sbesson:
With transformations being proposed to be renamed as `coordinateTransforms` and to become a requirement at the `datasets` level, I think supporting `datasets` to be passed as dictionaries is becoming extremely timely.

will-moore:
@sbesson The `test_writer()` does `write_image(...transformations)` and then reads the result to check transformations are there. The test didn't change in this PR yet because the API for `write_image()` didn't change.

But I can also add more tests for `write_multiscales_metadata` directly.
I think I'll include the `coordinateTransforms` renaming in this PR. It will be slightly out of step with omero-cli-zarr but it is probably simpler to keep the number of PRs smaller?

ome_zarr/writer.py
```diff
@@ -172,28 +189,32 @@ def write_multiscale(
...
     axes = _get_valid_axes(dims, axes, fmt)
 
-    paths = []
+    datasets: List[Union[str, dict]] = []
     for path, dataset in enumerate(pyramid):
         # TODO: chunks here could be different per layer
         group.create_dataset(str(path), data=dataset, chunks=chunks)
-        paths.append(str(path))
-    write_multiscales_metadata(group, paths, fmt, axes, transformations)
+        datasets.append({"path": str(path)})
```
sbesson, line 204: Should a default `scale` transformation be computed and stored under `"coordinateTransformations"` at this stage to comply with the latest specification proposal?

If so, what should be the behavior if an additional `coordinateTransformations` dictionary is passed?

ome_zarr/writer.py
```diff
@@ -172,28 +189,32 @@ def write_multiscale(
...
+
+    if coordinateTransformations is not None:
+        for dataset, transform in zip(datasets, coordinateTransformations):
+            dataset["coordinateTransformations"] = transform
```
sbesson, line 208: See also discussion in https://github.com/ome/ngff/pull/85/files#r794486756 alongside the ability to define global vs dataset/level coordinate transformations. If both forms are allowed, I suspect we will need to have a way for the API to handle both variants?

ome_zarr/writer.py
```diff
@@ -122,6 +122,23 @@ def _validate_plate_rows_columns(
     return validated_list
 
 
+def _validate_datasets(datasets: List[Union[str, dict]]) -> List[Dict]:
```
sbesson, line 125: With the proposed changes in 0.4 making `coordinateTransformations` mandatory, this API should likely be made `fmt` dependent with version-specific implementations delegated to the `Format` instances?

sbesson (changes requested):
A few API questions primarily related to the questions around defining/validating transformations

ome_zarr/writer.py
will-moore: If a `scale` is provided for each level, then we just use that. If ONLY a `translation` is provided, then I guess we can append that to the default `scale`

ome_zarr/writer.py
```diff
@@ -122,6 +122,23 @@ def _validate_plate_rows_columns(
...
+
+    validated_datasets = []
+    if datasets is None or len(datasets) == 0:
+        raise ValueError("Empty datasets list")
+    for dataset in datasets:
+        if isinstance(dataset, str):
+            validated_datasets.append({"path": dataset})
```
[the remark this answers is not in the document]
will-moore: In `omero-cli-zarr` I use this for generating a set of transformations for a pyramid:
```
def marshal_transformations(image, levels=1, multiscales_zoom=2.0)
```
But I realise that this is quite `omero-cli-zarr` specific (e.g. no down-sampling in Z). So yes, I guess this is required now.

ome_zarr/format.py
```diff
@@ -215,5 +241,60 @@ def validate_well_dict(
...
+        for shape in shapes:
+            assert len(shape) == len(data_shape)
+            scale = [full / level for full, level in zip(data_shape, shape)]
+            coordinate_transformations.append([{"type": "scale", "scale": scale}])
+
+        return coordinate_transformations
+
+    def validate_coordinate_transformations(
+        self,
+        shapes: List[tuple],
```
sbesson, line 260: Feels strange to have this API depending on shapes especially as only the length is used so far. Given the known assumptions, I wonder if `len(axes)` would be sufficient ?

ome_zarr/format.py
```diff
@@ -215,5 +241,60 @@ def validate_well_dict(
...
+            )
+        for shape, transformations in zip(shapes, coordinate_transformations):
+            ndim = len(shape)
+            assert isinstance(transformations, list)
+            scale_transfs = [
+                trans for trans in transformations if trans["type"] == "scale"
+            ]
+            if len(scale_transfs) != 1:
+                raise ValueError(
+                    "Must supply 1 'scale' item in coordinate_transformations"
```
sbesson, line 285: Two other potential elements to validate here:
- there should be at most one `translation`
- the `scale` transformation should be the first

ome_zarr/writer.py
```diff
@@ -122,6 +122,23 @@ def _validate_plate_rows_columns(
     return validated_list
 
 
+def _validate_datasets(datasets: List[dict]) -> List[Dict]:
+
+    validated_datasets = []
+    if datasets is None or len(datasets) == 0:
+        raise ValueError("Empty datasets list")
+    for dataset in datasets:
+        if isinstance(dataset, str):
```
sbesson, line 131: I assume this block can be completely removed now and `datasets == str` should be handled by the `ValueError` below

ome_zarr/writer.py
```diff
@@ -122,6 +122,23 @@ def _validate_plate_rows_columns(
...
+            validated_datasets.append({"path": dataset})
+        elif isinstance(dataset, dict):
+            if not dataset.get("path"):
+                raise ValueError("no 'path' in dataset")
+            validated_datasets.append(dataset)
```
sbesson, line 136: Should this also validate `coordinateTransformations`? If so, shouldn't the `fmt.validate_coordinate_transformations()` API be migrated here?

sbesson (commented):
A few questions

ome_zarr/writer.py
will-moore: That didn't quite work because we still support `List[str]` list of paths for `< 0.4`. Lots of tests failed when I removed it.

sbesson: Given that this API was introduced in  https://github.com/ome/ome-zarr-py/pull/149 and never made available in a public release, I would be personally fine with completely dropping this form and updating all the failing tests to use the new datasets form. The alternative would be to add some form of check on the format version here and fail hard if `fmt.version` is not 0.3 or lower.

I think the biggest issue I have reading the code is that it looks like there is a path for using this API and creating a `0.4` version of the `multiscales` with `datasets` only composed of `path` which is invalid. I can look into that in a follow-up PR.

ome_zarr/writer.py
```diff
@@ -172,28 +198,39 @@ def write_multiscale(
...
-        group.create_dataset(str(path), data=dataset, chunks=chunks)
-        paths.append(str(path))
-    write_multiscales_metadata(group, paths, fmt, axes, transformations)
+        group.create_dataset(str(path), data=data, chunks=chunks)
+        datasets.append({"path": str(path)})
+
+    shapes = [data.shape for data in pyramid]
+    if coordinate_transformations is None:
+        coordinate_transformations = fmt.generate_coordinate_transformations(shapes)
+    fmt.validate_coordinate_transformations(
```
sbesson, line 219: Although this does not hurt, is that call now redundant with the one that is called later during the validation of the `datasets` element?

sbesson (approved):
Overall looks good. One minor question around the redundant call to the transformations validation.

I think the unit tests should cover most of the validation logic for the transforms. I propose to review the code coverage as well as https://github.com/ome/ome-zarr-py/pull/162#discussion_r798623248 in a follow-up PR so that we can start testing this API.

Two general API thoughts which are probably outside the scope of this PR:
- I can foresee the need to add support for writing of bot `datasets:coordinateTransformations` and `multiscales:coordinateTransformations`. A fairly crude idea would be to use the current API but support `coordinate_transformations` lists of dimensions `len(datasets) + 1` and use the first or last item as the shared transformation
- the additional logic in `write_multiscales_metadata` points at  the fact that `axes` and `datasets` are now more tightly coupled to each other. Moving forward, I wonder whether `_validate_datasets` should simply become `_validate_multiscales` and internalize all the internal checks between fields. Longer-term, I would hope this API cab=n be simply replaced by the usage of generic validation framework/constraints.

In the interim, proposing to merge this so that we can test it together with `omero-cli-zarr`. Is it worth cutting a new pre-release?

ome_zarr/writer.py
will-moore: Not quite! If you pass in a list of transformations that is longer than the number of pyramid levels, this will fail validation at this point. BUT, if we remove the validation here it *won't* fail later because the transformations are zipped onto the Datasets, so the lengths will match (and the extra transformations will be ignored).
It is probably better to fail than to ignore the mismatch.

sbesson: 👍 Semi-related I discovered that Python 3.10 introduced a `strict` argument to [zip](https://docs.python.org/3.10/library/functions.html#zip) that will eventually allow similar code to fail on mismatching list lengths rather than truncating. Let's keep things as they are for now but maybe add an inline comment explaining this logic?

will-moore: Done in [b081342](https://github.com/ome/ome-zarr-py/pull/162/commits/b08134238d9552b2c46b821ff6effb4cd958a9e0)

sbesson:
@will-moore  https://pypi.org/project/ome-zarr/0.3a2/ is now available as a pre-release with this API. Do you want to bump the requirement in `napari-ome-zarr` and `omero-cli-zarr`?

PR #165, opened 2022-02-04
Multiscales metadata API: only support datasets as lists of dictionaries

sbesson:
Follow-up of the discussion in #162, this PR primarily the `write_multiscales_metadata` API introduced in #149 to only work on `datasets` passed as list of dictionaries. The writer unit tests are all adjusted for different version of the OME-NGFF format.

Additional unit tests are also added testing various combinations of valid and invalid `coordinateTransformations`. The transformations validation method is updated to add a few extra checks on the presence of mandatory keys.

### DIFF
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -126,17 +126,13 @@ def _validate_datasets(
     datasets: List[dict], dims: int, fmt: Format = CurrentFormat()
 ) -> List[Dict]:
 
-    validated_datasets = []
     if datasets is None or len(datasets) == 0:
         raise ValueError("Empty datasets list")
     transformations = []
     for dataset in datasets:
-        if isinstance(dataset, str):
-            validated_datasets.append({"path": dataset})
-        elif isinstance(dataset, dict):
+        if isinstance(dataset, dict):
             if not dataset.get("path"):
                 raise ValueError("no 'path' in dataset")
-            validated_datasets.append(dataset)
             transformation = dataset.get("coordinateTransformations")
             # transformation may be None for < 0.4 - validated below
             if transformation is not None:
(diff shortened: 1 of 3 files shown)


ome_zarr/writer.py
```diff
@@ -126,17 +126,13 @@ def _validate_datasets(
...
 
-    validated_datasets = []
     if datasets is None or len(datasets) == 0:
         raise ValueError("Empty datasets list")
     transformations = []
     for dataset in datasets:
-        if isinstance(dataset, str):
-            validated_datasets.append({"path": dataset})
-        elif isinstance(dataset, dict):
+        if isinstance(dataset, dict):
```
joshmoore, line 133: This might have been simpler as `assert isinstance(dataset, dict)`

joshmoore: Now I'm pondering if there's some generic `assertion` or `validation` module that one could use to skip for loops where something isn't valid without just throwing a `ValueError`.

joshmoore (commented):
I missed most of the overarching discussion, but barring one minor comment, lgtm.

ome_zarr/writer.py
sbesson: No strong opinion on using `assert` vs conditional and a custom error like `ValueError`. If we collectively agree on the former version for now, I can quickly go over the codebase after this PR.

My mid-term hope is to replace this internal validation code by a generic validation framework based on schemas/constraints. I expect this should also bring typical features like being able to collect all validation errors like the `Validator.iter_errors()` API of the `jsonschema` module.

joshmoore: Exactly. No strong opinion on `assert isinstance` especially if we'll be reviewing soon _anyway_.

PR #174, opened 2022-03-02
Add Implicit spec to loop over metadata-less "collections"

joshmoore:
Part of the investigation of metadata in https://github.com/ome/ngff/issues/104. This "implicit" group is the cheapest form of collection imaginable.

Currently, only _groups_ within the given group (and not arrays or explicit files) will be further parsed.

### DIFF
--- a/ome_zarr/reader.py
+++ b/ome_zarr/reader.py
@@ -7,6 +7,7 @@ from typing import Any, Dict, Iterator, List, Optional, Type, Union, cast, overl
 
 import dask.array as da
 import numpy as np
+import zarr
 from dask import delayed
 
 from .axes import Axes
@@ -45,21 +46,32 @@ class Node:
         self.post_nodes: List[Node] = []
 
         # TODO: this should be some form of plugin infra over subclasses
+        found: List[Spec] = []
         if Labels.matches(zarr):
-            self.specs.append(Labels(self))
+            found.append(Labels(self))
+            self.specs.append(found[-1])
         if Label.matches(zarr):
-            self.specs.append(Label(self))
+            found.append(Label(self))
+            self.specs.append(found[-1])
         if Multiscales.matches(zarr):
-            self.specs.append(Multiscales(self))
+            found.append(Multiscales(self))
+            self.specs.append(found[-1])
         if OMERO.matches(zarr):
-            self.specs.append(OMERO(self))
+            found.append(OMERO(self))
+            self.specs.append(found[-1])
         if plate_labels:
-            self.specs.append(PlateLabels(self))
+            found.append(PlateLabels(self))
+            self.specs.append(found[-1])
         elif Plate.matches(zarr):
-            self.specs.append(Plate(self))
+            found.append(Plate(self))
+            self.specs.append(found[-1])
             # self.add(zarr, plate_labels=True)
         if Well.matches(zarr):
-            self.specs.append(Well(self))
+            found.append(Well(self))
+            self.specs.append(found[-1])
+
+        if not found:
+            self.specs.append(Implicit(self))
 
     @overload
     def first(self, spectype: Type["Well"]) -> Optional["Well"]:
@@ -176,6 +188,25 @@ class Spec(ABC):
 
     def lookup(self, key: str, default: Any) -> Any:
         return self.zarr.root_attrs.get(key, default)
+
+
+class Implicit(Spec):
+    """
+    A spec-type which simply iterates over available zgroups.
+    """
+
+    @staticmethod
+    def matches(zarr: ZarrLocation) -> bool:
+        """Always return true"""
+        return True
+
+    def __init__(self, node: Node) -> None:
+        super().__init__(node)
+
+        for name in zarr.group(self.zarr.store).group_keys():
+            child_zarr = self.zarr.create(name)
+            if child_zarr.exists():
+                node.add(child_zarr)
 
 
 class Labels(Spec):


joshmoore:
See https://github.com/ome/ome-zarr-metadata/releases/tag/0.1.0 for an example of an entrypoint. After creating a fake .zgroup under the output of `bioformats2raw a.fake /tmp/a.ome.zarr`

```
$ ome_zarr info /tmp/a.ome.zarr/0/test/
/private/tmp/a.ome.zarr/0/test [zgroup]
 - metadata
   - Implicit (1)
   - Leaf (2)
 - data
/private/tmp/a.ome.zarr/0 [zgroup]
 - metadata
   - Multiscales
   - Leaf (2)
 - data
   - (1, 1, 1, 512, 512)
   - (1, 1, 1, 256, 256)
/private/tmp/a.ome.zarr [zgroup]
 - metadata
   - bioformats2raw (3)
   - Root (2)
 - data
```

Notice:
 1. the `Implicit` spec scans groups that have no other metadata
 2. `Leaf/Root` work their way up and back down a hierarchy
 3. `bioformats2raw` reads `OME/METADATA.ome.xml`

ome_zarr/reader.py
```diff
@@ -45,21 +47,44 @@ def __init__(
...
-            self.specs.append(Plate(self))
+            found.append(Plate(self))
+            self.specs.append(found[-1])
             # self.add(zarr, plate_labels=True)
         if Well.matches(zarr):
-            self.specs.append(Well(self))
+            found.append(Well(self))
+            self.specs.append(found[-1])
+
+        for key, value in entrypoints.get_group_named("ome_zarr.spec").items():
```
will-moore, line 74: What does this do? I see no file named `"ome_zarr.spec"` and I've not used `entrypoints` before. I don't get much idea from https://github.com/takluyver/entrypoints as to what it does, only that "This package is in maintenance-only mode. New code should use the `importlib.metadata module`".

will-moore:
For an Image in a Plate (12 Wells A-C, 1-4, all wells with labels), without this PR I get:

```
$ ome_zarr info 251.zarr/A/1/0/
/Users/wmoore/Desktop/ZARR/data/v4_omero/plates/251.zarr/A/1/0 [zgroup]
 - metadata
   - Multiscales
   - OMERO
 - data
   - (3, 1024, 1344)
   - (3, 512, 672)
   - (3, 256, 336)
   - (3, 128, 168)
   - (3, 64, 84)
/Users/wmoore/Desktop/ZARR/data/v4_omero/plates/251.zarr/A/1/0/labels [zgroup] (hidden)
 - metadata
   - Labels
 - data
/Users/wmoore/Desktop/ZARR/data/v4_omero/plates/251.zarr/A/1/0/labels/0 [zgroup] (hidden)
 - metadata
   - Label
   - Multiscales
 - data
   - (1, 1024, 1344)
   - (1, 512, 672)
   - (1, 256, 336)
   - (1, 128, 168)
   - (1, 64, 84)
   - (1, 32, 42)
```

and with this PR I get all the sibling `A` Wells A2, A3, A4, but not B1-B4 or C1-C4. And I don't get labels for those Wells.
```
$ ome_zarr info 251.zarr/A/1/0/
/Users/wmoore/Desktop/ZARR/data/v4_omero/plates/251.zarr/A/1/0 [zgroup]
 - metadata
   - Multiscales
   - OMERO
   - Leaf
 - data
   - (3, 1024, 1344)
   - (3, 512, 672)
   - (3, 256, 336)
   - (3, 128, 168)
   - (3, 64, 84)
/Users/wmoore/Desktop/ZARR/data/v4_omero/plates/251.zarr/A/1/0/labels [zgroup] (hidden)
 - metadata
   - Labels
   - Leaf
 - data
/Users/wmoore/Desktop/ZARR/data/v4_omero/plates/251.zarr/A/1/0/labels/0 [zgroup] (hidden)
 - metadata
   - Label
   - Multiscales
   - Leaf
 - data
   - (1, 1024, 1344)
   - (1, 512, 672)
   - (1, 256, 336)
   - (1, 128, 168)
   - (1, 64, 84)
   - (1, 32, 42)
/Users/wmoore/Desktop/ZARR/data/v4_omero/plates/251.zarr/A/1 [zgroup]
 - metadata
   - Well
   - Leaf
 - data
   - (3, 1024, 1344)
/Users/wmoore/Desktop/ZARR/data/v4_omero/plates/251.zarr/A [zgroup]
 - metadata
   - Implicit
   - Leaf
 - data
/Users/wmoore/Desktop/ZARR/data/v4_omero/plates/251.zarr/A/2 [zgroup]
 - metadata
   - Well
   - Leaf
 - data
   - (3, 1024, 1344)
/Users/wmoore/Desktop/ZARR/data/v4_omero/plates/251.zarr/A/3 [zgroup]
 - metadata
   - Well
   - Leaf
 - data
   - (3, 1024, 1344)
/Users/wmoore/Desktop/ZARR/data/v4_omero/plates/251.zarr/A/4 [zgroup]
 - metadata
   - Well
   - Leaf
 - data
   - (3, 1024, 1344)
/Users/wmoore/Desktop/ZARR/data/v4_omero/plates/251.zarr [zgroup]
 - metadata
   - Plate
   - Root
 - data
   - (3, 768, 1344)
```

Without this PR, `napari 251.zarr/A/1/0/` gives me just the 1 image + labels:

Screenshot 2022-03-23 at 10 47 04

With this PR, I get everything as for 'info' above: All the `A` wells, but only labels for A1:
Screenshot 2022-03-23 at 10 46 07

joshmoore:
> and with this PR I get all the sibling `A` Wells A2, A3, A4, but not B1-B4 or C1-C4. And I don't get labels for those Wells.

@will-moore, you approve of getting the siblings? If so, I'll look why it's only for the one row.

joshmoore:
@will-moore, I've reverted the upwards parsing. It seemed like a good strategy but there are currently too many edge cases. I don't have labels on plates for testing at the moment, but I think with the current state along with https://github.com/ome/ome-zarr-metadata/commit/08e12f784e085adbf4ca6d384720443108f96cb6#diff-0bb17e0ecb4ac83835ee3800a1af71a12f644b0ce782c623ba97f8917916250eR54 all the following should be true:

|         | non-bf2raw | bf2raw               |
|---------|------------|----------------------|
| HCS     | unchanged  |      unchanged       |
| non-HCS | unchanged  | now loads all images |

The only other change I can think of is if you pass a group that previously did nothing, it will likely try to load the contents.

joshmoore:
In discussing today with @dgault, @sbesson, @jburel and @melissalinkert, there was a case made for at least adding the flag (`Leaf`) to make it possible for clients to detect that there is more information that needs loading. Additional methods or parameters should then allow that loading.

joshmoore:
To improve the codecov results, see https://github.com/zarr-developers/numcodecs/pull/300/files#diff-bc37cd9860eec1facdc18a47798e8a1a2c0ef5dabd999deee049de4a48a5d35fR1 for an option of in-repo testing of entrypoints.

joshmoore:
see a quick use of this functionality:
 - https://github.com/ome/ome-zarr-metadata/pull/1
 - https://github.com/ome/napari-ome-zarr/pull/47

ome_zarr/bioformats2raw.py
```diff
@@ -0,0 +1,99 @@
...
+            if data.plates:
+                _logger.info("Plates detected. Skipping implicit loading")
+            else:
+                for idx, image in enumerate(data.images):
+                    series = node.zarr.create(str(idx))
+                    assert series.exists(), f"{series} is missing"
+                    _logger.info("found %s", series)
+                    subnode = node.add(series)
+                    if subnode:
+                        subnode.metadata["ome-xml:index"] = idx
```
joshmoore, line 44: This is probably the biggest lingering question I have since this basically becomes public API. Thoughts welcome. cc: @will-moore @sbesson _et al._

will-moore: I wonder where this public API is documented?
This new module should be added to https://github.com/ome/ome-zarr-py/blob/master/docs/source/api.rst but I think these changes in parsing also need to be described elsewhere, maybe at https://ome-zarr.readthedocs.io/en/stable/python.html#reading-ome-ngff-images

ome_zarr/bioformats2raw.py
```diff
@@ -0,0 +1,98 @@
...
+                        subnode.metadata["ome-xml:image"] = image
```
will-moore, line 44: If you add the following here, then you don't need https://github.com/ome/napari-ome-zarr/pull/47/
```
subnode.metadata["name"] = image.name
```

joshmoore: But then plugins would/could overwrite the main metadata, right? Which means it becomes a question of the order that the plugins run in. Would:

```
subnode.plugins["bioformats2raw"].metadata["name"]
```
be better?

sbesson: As probably transpired from https://github.com/ome/ome-zarr-py/pull/140, I am also unclear on the contract and expectation for the `node.metadata` field. My impression was that this API has been largely driven by the napari use-case.Reading this, it looks like another goal is to introduce some form of metadata consolidation so that a consumer does not have to go up and down the hierarchy to assemble pieces of metadata, is that correct?

If so, I agree what is missing is some form of namespace to differentiate the metadata injected at different levels in the hierarchy. Taking advantage of the fact this is a Python dictonary, another proposal would be to group all the metadata under a top-level key e.g. `subnode.metadata["bioformats2raw"] = {"index": idx, "image": image", "name", image.name}`

joshmoore: > .Reading this, it looks like another goal is to introduce some form of metadata consolidation so that a consumer does not have to go up and down the hierarchy to assemble pieces of metadata, is that correct?

Maybe. But now that we have the json-schema, I could also see just attaching objects at the right node level to prevent re-reading the file. That's essentially what would happen if the ome-types `OME` object becomes part of the bioformats2raw plugin's public API.

will-moore: In my mind I might expect something like Josh's original idea or `subnode.metadata["ome.xml"] = {"index": idx, "image": image", "name", image.name}` since this is metadata coming from the `ome.xml`?

The reason I like `subnode.metadata["name"] = image.name` is that any consumer doesn't have to know about `bioformats2raw` or xml etc to have the correct image name.
But if it's just `napari-ome-zarr` and we already have https://github.com/ome/napari-ome-zarr/pull/47 then 👍

joshmoore:
Should we also discuss the name of the module itself?

ome_zarr/bioformats2raw.py
joshmoore: > is that any consumer doesn't have to know about bioformats2raw or xml etc to have the correct image name.

This possibly gets back to the question of whether or not this bf2raw parsing is core or not. If it is, then you're probably right that we should just encode the rules for what "wins" directly. If this is a plugin, though, then how would we handle misbehavior in the plugins?

will-moore:
We added an `omero` block of channel & rendering metadata to the `multiscale .zattrs` (because it came from omero) but we actually want other tools to read and write this metadata, which may be discouraged by the naming.
In the same way, `bioformats2raw.layout` is a spec that just happens to be produced originally by `bioformats2raw`, but it's really a spec that ALL tools should read/write.
I don't know if it's too late to think about a different name there, or if the name has already stuck?

ome_zarr/bioformats2raw.py
will-moore: What is a plugin here? If it's a plugin, then it's not installed by default and you get to choose if you want to add it?

sbesson: > This possibly gets back to the question of whether or not this bf2raw parsing is core or not

Trying to answer this, this is where I would draw the line: 

- if a specification is published in https://ngff.openmicroscopy.org/, I consider it as `core` (even if `transitional`) and my expectation is that this library should support it
- a contrario, any specification _not_ defined in the upstream specification should be seen as an extension and should  be handled by some plugin/third-party mechanism. This might drive the discussions on the extensibility of this library which is a good thing.

joshmoore:
Other than the string `bioformats2raw.layout` we're pretty free to change things here. (I'd say we definitely *don't* want to reproduce what we did with `omero` and we actually need to think about how to make that "transitional" as well)

will-moore:
Ah - yes, too late to change the `"bioformats2raw.layout"` key because data generated with this already exists.

Issue #171, opened 2022-02-28
Is writing "image-label" supported?


constantinpape:
Is there support to write "image-label" datasets (https://ngff.openmicroscopy.org/latest/#label-md) in ome-zarr-py already?
If yes, how do I use it?

will-moore:
Threre's an example at https://github.com/ome/ome-zarr-py/blob/0ea5bf61ef754069abf4f223897d55630c953fbf/README.rst
which is part of a long-standing PR https://github.com/ome/ome-zarr-py/pull/121
Actually, I have subsequently removed the labels from the PR to make it more concise for the README

constantinpape:
Thanks @will-moore; the example was very helpful and I could update my scripts to write ome.zarr with labels.

sbesson:
Looking at the example, I could imagine wrapping the succession of low-level calls into utilities dealing with the group/metadata creation/update e.g.

```
write_image(image=data, group=image_root)
label_root = add_label(image_root, label_name)
write_image(image=label_data, group=label_root)
```

or 

```
write_image(image=data, group=image_root)
write_label(image=image_root, group=image_root, name=label_name)
```

It would be interesting to know you would see the value of introducing such API based on your usage @constantinpape.

constantinpape:
> It would be interesting to know you would see the value of introducing such API based on your usage @constantinpape.

Yes, absolutely! `write_label` should also have support for (optional? depending on https://github.com/ome/ngff/issues/105) arguments to pass the label/property values.

sbesson:
I have updated the title of this issue to reflect the new scope.

So current preference would be rather a single API that would first create/update the intermediate groups (depending on the presence of labels) and then write the label image, is that correct? In terms of behavior, I agree the API should also support both mandatory as well as optional `image-label` attributes.

For the metadata, my current preference would be to extend `write_multiscale` to support an extra metadata dictionary similarly `storage_options`. This would allow to set additional metadata, either specified (`omero`, `image-label`...) or custom, to be stored in the attributes of the multiscales group.

Thinking of the sequence of a `write_label_image` API, the biggest unknown to me at this point is whether this API should call `write_image` or `write_multiscales`. The former is effectively a wrapper around the latter but I do not know whether it should always be seen as the single entry-point @joshmoore @will-moore ?

constantinpape:
Implemented in #178

PR #178, opened 2022-03-17
Add image-label convenience functions

### LINKED ISSUE #176: write_multiscales does not expose additional metadata
[write_multiscales](https://github.com/ome/ome-zarr-py/blob/master/ome_zarr/writer.py#L170) does not expose additional `metadata` as a argument, which could be used to e.g. add the `name` field.
Note that this is supported by [write_image](https://github.com/ome/ome-zarr-py/blob/master/ome_zarr/writer.py#L364). I think that `write_multiscales` should be updated s.t. it also supports this.

constantinpape:
Fixes #176 and #171

This PR:
- unifies `**metadata` across `write_image`, `write_multiscale` and `write_multiscales_metadata`
- introduces new convenience function `write_multiscale_image_labels` to create `image-label` data. 

For a usage example see https://github.com/ome/ome-ngff-prototypes/blob/update-ome-zarr-py/workflows/spatial-transcriptomics-example/convert_transcriptomics_data_to_ngff.py.

Still TODO:
- [x] ~~Add `write_image_labels` (analogue to `write_image`) that uses `write_multiscale_image_labels` internally~~ (leave this for a follow-up todo and also clean up `write_image` ...)
- [ ] Add tests

I will address the TODOs if we decide to go forward with these changes.
Side-note: some confusing / unclear things I noticed in the `image-labels` spec while working on this.

### DIFF
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -175,6 +175,7 @@ def write_multiscale(
     axes: Union[str, List[str], List[Dict[str, str]]] = None,
     coordinate_transformations: List[List[Dict[str, Any]]] = None,
     storage_options: Union[JSONDict, List[JSONDict]] = None,
+    **metadata: JSONDict,
 ) -> None:
     """
     Write a pyramid with multiscale metadata to disk.
@@ -232,7 +233,7 @@ def write_multiscale(
         for dataset, transform in zip(datasets, coordinate_transformations):
             dataset["coordinateTransformations"] = transform
 
-    write_multiscales_metadata(group, datasets, fmt, axes)
+    write_multiscales_metadata(group, datasets, fmt, axes, **metadata)
 
 
 def write_multiscales_metadata(
@@ -240,6 +241,7 @@ def write_multiscales_metadata(
     datasets: List[dict],
     fmt: Format = CurrentFormat(),
     axes: Union[str, List[str], List[Dict[str, str]]] = None,
+    **metadata: JSONDict,
 ) -> None:
     """
     Write the multiscales metadata in the group.
@@ -272,6 +274,7 @@ def write_multiscales_metadata(
         {
             "version": fmt.version,
             "datasets": _validate_datasets(datasets, ndim, fmt),
+            **metadata
         }
     ]
     if axes is not None:
@@ -432,8 +435,8 @@ def write_image(
         axes=axes,
         coordinate_transformations=coordinate_transformations,
         storage_options=storage_options,
+        **metadata
     )
-    group.attrs.update(metadata)
 
 
 def _retuple(
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -175,7 +175,7 @@ def write_multiscale(
     axes: Union[str, List[str], List[Dict[str, str]]] = None,
     coordinate_transformations: List[List[Dict[str, Any]]] = None,
     storage_options: Union[JSONDict, List[JSONDict]] = None,
-    **metadata: JSONDict,
+    **metadata: Union[str, JSONDict, List[JSONDict]],
 ) -> None:
     """
     Write a pyramid with multiscale metadata to disk.
@@ -241,7 +241,7 @@ def write_multiscales_metadata(
     datasets: List[dict],
     fmt: Format = CurrentFormat(),
     axes: Union[str, List[str], List[Dict[str, str]]] = None,
-    **metadata: JSONDict,
+    **metadata: Union[str, JSONDict, List[JSONDict]],
 ) -> None:
     """
     Write the multiscales metadata in the group.
@@ -369,7 +369,7 @@ def write_image(
     axes: Union[str, List[str], List[Dict[str, str]]] = None,
     coordinate_transformations: List[List[Dict[str, Any]]] = None,
     storage_options: Union[JSONDict, List[JSONDict]] = None,
-    **metadata: JSONDict,
+    **metadata: Union[str, JSONDict, List[JSONDict]],
 ) -> None:
     """Writes an image to the zarr store according to ome-zarr specification
 
@@ -425,13 +425,13 @@ def write_image(
                 "Can't downsample if size of x or y dimension is 1. "
                 "Shape: %s" % (image.shape,)
             )
-        image = scaler.nearest(image)
+        mip = scaler.nearest(image)
     else:
         LOGGER.debug("disabling pyramid")
-        image = [image]
+        mip = [image]
 
     write_multiscale(
-        image,
+        mip,
         group,
         chunks=chunks,
         fmt=fmt,
@@ -449,7 +449,7 @@ def write_label_metadata(
     name: str,
     colors: List[JSONDict] = None,
     properties: List[JSONDict] = None,
-    **metadata: JSONDict,
+    **metadata: Union[List[JSONDict], JSONDict, str],
 ):
     """
     Write image-label metadata to the group.
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -442,8 +442,6 @@ def write_image(
     )
 
 
-# how do we actually validate this?
-# do we have a schema for the image-label metadata already?
 def write_label_metadata(
     group: zarr.Group,
     name: str,
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -270,12 +270,11 @@ def write_multiscales_metadata(
             if axes is not None:
                 ndim = len(axes)
 
+    # note: we construct the multiscale metadata via 'dict(...)' rather than `{...}` in order
+    # to avoid duplication of protected keys like 'version' in **metadata
+    # (for {} this would silently over-write it, with dict() it explicitly fails)
     multiscales = [
-        {
-            "version": fmt.version,
-            "datasets": _validate_datasets(datasets, ndim, fmt),
-            **metadata
-        }
+        dict(version=fmt.version, datasets=_validate_datasets(datasets, ndim, fmt), **metadata)
     ]
     if axes is not None:
         multiscales[0]["axes"] = axes
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -439,6 +439,75 @@ def write_image(
     )
 
 
+def write_multiscale_image_labels(
+    pyramid: List,
+    group: zarr.Group,
+    name: str,
+    chunks: Union[Tuple[Any, ...], int] = None,
+    fmt: Format = CurrentFormat(),
+    axes: Union[str, List[str], List[Dict[str, str]]] = None,
+    coordinate_transformations: List[List[Dict[str, Any]]] = None,
+    colors: List[JSONDict] = None,
+    properties: List[JSONDict] = None,
+    storage_options: Union[JSONDict, List[JSONDict]] = None,
+    **metadata: JSONDict,
+) -> None:
+    """
+    Write image labels in pyramidal format with multiscale and image-label metadata to disk.
+
+    Creates the label data in the sub-group "labels/{name}"
+
+    pyramid: List of np.ndarray
+      the image label data to save. Largest level first
+      All image arrays MUST be up to 5-dimensional with dimensions
+      ordered (t, c, z, y, x)
+    group: zarr.Group
+      the group within the zarr store to store the data in
+    name: str
+      the name of this labale data
+    chunks: int or tuple of ints,
+      size of the saved chunks to store the image
+    fmt: Format
+      The format of the ome_zarr data which should be used.
+      Defaults to the most current.
+    axes: str or list of str or list of dict
+      List of axes dicts, or names. Not needed for v0.1 or v0.2
+      or if 2D. Otherwise this must be provided
+    coordinate_transformations: 2Dlist of dict
+      For each path, we have a List of transformation Dicts.
+      Each list of dicts are added to each datasets in order
+      and must include a 'scale' transform.
+    colors: list of dict
+      Fixed colors for (a subset of) the label values.
+      Each dict specifies the color for one label and must contain the fields
+      "label-value" and "rgba".
+    properties: list of dict
+      Additional properties for (a subset of) the label values.
+      Each dict specifies additional properties for one label.
+      It must contain the field "label-value" and may contain arbitrary additional properties.
+    storage_options: dict or list of dict
+      Options to be passed on to the storage backend. A list would need to match
+      the number of datasets in a multiresolution pyramid. One can provide
+      different chunk size for each level of a pyramind using this option.
+    """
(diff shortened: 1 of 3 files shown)


ome_zarr/writer.py
```diff
@@ -432,8 +435,85 @@ def write_image(
...
+        fmt,
+        axes,
+        coordinate_transformations,
+        storage_options,
+        name=name,
+        **metadata,
     )
-    group.attrs.update(metadata)
+
+    image_label_metadata = {}
```
sbesson, line 506: I assume this dictionary could also be assembled, combined with `**metadata` and passed to the `write_multiscale` call above?

ome_zarr/writer.py
```diff
@@ -432,8 +435,85 @@ def write_image(
...
+      The format of the ome_zarr data which should be used.
+      Defaults to the most current.
+    axes: str or list of str or list of dict
+      List of axes dicts, or names. Not needed for v0.1 or v0.2
+      or if 2D. Otherwise this must be provided
+    coordinate_transformations: 2Dlist of dict
+      For each path, we have a List of transformation Dicts.
+      Each list of dicts are added to each datasets in order
+      and must include a 'scale' transform.
+    colors: list of dict
```
sbesson, line 480: The primary downside of these explicit declaration is that it prevents someone to use this wrapper to store additional `image-label` metadata.
Another option would be to have pass the entire dictionary for `image-label` as a single argument?

ome_zarr/writer.py
```diff
@@ -432,8 +435,85 @@ def write_image(
...
+    properties: list of dict
+      Additional properties for (a subset of) the label values.
+      Each dict specifies additional properties for one label.
+      It must contain the field "label-value" and may contain arbitrary additional properties.
+    storage_options: dict or list of dict
+      Options to be passed on to the storage backend. A list would need to match
+      the number of datasets in a multiresolution pyramid. One can provide
+      different chunk size for each level of a pyramind using this option.
+    """
+    sub_group = group.require_group(f"labels/{name}")
```
sbesson, line 493: Should there be any check that the current group is actually a `multiscales` itself?

ome_zarr/writer.py
```diff
@@ -432,8 +435,85 @@ def write_image(
...
-    group.attrs.update(metadata)
+
+    image_label_metadata = {}
+    if colors is not None:
+        image_label_metadata["colors"] = colors
+    if properties is not None:
+        image_label_metadata["properties"] = properties
+    sub_group.attrs["image-label"] = image_label_metadata
+
+    label_group = group["labels"]
```
sbesson, line 513: Is the intermediate group created by the `require_group` above?

sbesson (changes requested):
Thanks. That's a great starting point to introducing some functionalities allowing to write label images. A couple of inline design questions mostly about the new label writing. 

Thinking of other use case, @will-moore we also have some code in `omero-cli-zarr` writing label images from OMERO masks. Would the new API be flexible enough to e.g. delegate some of the label creation logic to `ome_zarr.writer`?

As suggested by the codecov report, I think we want to introduce some unit tests covering the semantics of the new API and testing the metadata is correctly stored.

ome_zarr/writer.py
will-moore: I'm not sure that would be very useful. I have previously wanted to export labels (without image data) using `omero-cli-zarr` and I think I was prevented. The labels without image might not be recognised as labels, but they are still valid zarr data.

ome_zarr/writer.py
constantinpape: No that wouldn't work, since `**metadata` is written to `multiscales`. Whereas `image_label_metadata` is written to `image-label`, which is at the root of the json:
```yaml
{
"multiscales": [
  {
    ...  # this is where metadata goes
  }
],
"image-label": {
  ...  #  this is where image_label_metadata goes
}
}
```

ome_zarr/writer.py
constantinpape: You're right, it's probably better to remove `colors` and `properties` in favor of a dict argument `image_label_metadata` (or similar name).

ome_zarr/writer.py
```diff
@@ -432,8 +435,85 @@ def write_image(
         axes=axes,
         coordinate_transformations=coordinate_transformations,
         storage_options=storage_options,
+        **metadata,
+    )
+
+
+def write_multiscale_image_labels(
```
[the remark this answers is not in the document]
constantinpape: I wouldn't do that, because I think we should mirror the functionality for normal multiscale data and have
- `write_multiscale_image_labels`, corresponding to `write_multiscale` (which I have implemented here already)
- `write_image_label`, corresponding to `write_image` (not implemented yet, since I wanted to wait till we agree on the signature for `write_multiscale_image_labels`; this would use `write_multiscale_image_labels` internally).

If you want to simplify the names we can rename `write_multiscale_image_labels` -> `write_multiscale_labels` and `write_image_labels` -> `write_labels` though; that would be fine with me.

ome_zarr/writer.py
constantinpape: yes

ome_zarr/writer.py
constantinpape: I would like to keep this close to the spec, which is pretty explicit about the intermediate `labels` hierarchy. If we allow more freedom here it will be confusing for users, who should not need to worry about the internal file structure to much.
For more customization advanced users can just use `write_multiscale` or `write_image` and add the necessary metadata themselves. We could think about adding `write_(image)_label_metadata` for this use-case to again mirror the functionality for normal multiscales metadata.

ome_zarr/writer.py
sbesson: Ah I missed you changed the behavior of the API to have `**metadata` now appended  to the `multiscales` key level (as opposed to the top-level previously).

Is there another option for someone to add top-level keys e.g. `omero` and `image-label` as two specified examples?

constantinpape: > Ah I missed you changed the behavior of the API to have **metadata now appended to the multiscales key level (as opposed to the top-level previously).

Yes, it's changed but I think the previous behavior was **very** unfortunate, since it would just write into top level without any connection to the multiscales metadata. So to add a property to multiscales you would need to do some very weird code gymnastics and it might even lead to incorrect metadata if you over-write a reserved key with something invalid.

constantinpape: > Is there another option for someone to add top-level keys e.g. omero and image-label as two specified examples?

Not in the current function. But I think that for this use case just doing e.g. `group.attrs["omero"] = {...}` is acceptable.
Or if you want to be more explicit we could add explicit functions for this like `write_image_label_metadata` (could be added in this PR) and `write_omero_metadata` (this should be a different PR since it's unrelated). This would also mirror the current `write_multiscale_metadata` function.

ome_zarr/writer.py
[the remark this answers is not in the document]
sbesson: Understood. For me, the most confusing part is that a label image must implement the multiscales specification so the `_multiscales` bit feels redundant.
That being said, there is already a (confusing) distinction between `write_image` and `write_multiscales` in the current API.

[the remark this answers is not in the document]
constantinpape: > Understood. For me, the most confusing part is that a label image must implement the multiscales specification so the `_multiscales` bit feels redundant.

But we need to keep the `_multiscales` to keep the distinction between passing a single array or the MIP. 

> That being said, there is already a (confusing) distinction between `write_image` and `write_multiscales` in the current API.

Yes, that's why we need the `_multiscales` here.

(I will go ahead and simplify a bit by removing the "image" as layed out above.)

ome_zarr/writer.py
sbesson: Following up, my initial expectation was that the current API could append multiscales metadata when passing a dictionary using a `multiscales` key. However running `write_image(data, self.group, axes="zyx", multiscales={"type": "gaussian"})` made me realize that the current implementation completely overwrites the `multiscales` metadata.

So :+1: for fixing this bug and defining an API with a set of tests allowing to store optional `multiscales` metadata esp. since several keys are defined in the spec. A few questions:
- what is the behavior if a core key (`axes`, `datasets`) is also passed as keywords?
- is there still room for an API allowing to defining other top-level keys or is the recommendation for consumers to invoke `group.attrs.update()`?

constantinpape:
@sbesson I have changed a few things based on your review:
- rename `write_multiscale_image_labels` -> `write_multiscale_labels`
- change its function signature so that image-label metadata can be passed as a dict
- implement `write_label_metadata`
- small changes to `write_multiscales_metadata` to prevent over-writing protected keys

I decided not to implement `write_labels`, which would correspond to `write_image`, here: the signature of `write_image` is outdated, e.g. `byte_order` is not used anymore and some other issues. I would tackle this in a follow-up PR and then implement `write_labels` once this is done.

What's left to do:
- add tests
- fix the pre-commit (I have no idea what that thing is doing, can you maybe check and let me know how to fix it?)

One more question:
do we have a schema for `image-labels` already that can be used for validation in the tests?

constantinpape:
Follow up for `write_image`:
- remove `byte_order`
- remove `chunks` (should be part of `storage_options`)
- do [this](https://github.com/ome/ome-zarr-py/blob/master/ome_zarr/writer.py#L413-L414) in `write_multiscale_labels`, to have the same functionality for both

sbesson:
>  do we have a schema for image-labels already that can be used for validation in the tests?

Not at the moment. The steps to unlock this would be:

- add a new JSON schema and examples (migrate the snippet in the spec) under https://github.com/ome/ngff/tree/main/0.4
- make the decisions on the distribution mechanism @will-moore was working this in https://github.com/ome/ngff/pull/77 and https://github.com/ome/ome-zarr-py/pull/142 using the standard Python packaging appproach

will-moore:
Looks good. I've started to consume this with `omero-cli-zarr` in https://github.com/ome/omero-cli-zarr/pull/107, which is working for a single Image. Not tried it for plates yet.

constantinpape:
I have added a test now and fixed the pre-commit. This is good to go from my side. Please let me now if you see the need for any more changes, @sbesson @will-moore.
Once this is in I will follow up with a PR for cleaning up the `write_image` function signature and implementing `write_labels`.

> > do we have a schema for image-labels already that can be used for validation in the tests?
> 
> Not at the moment. The steps to unlock this would be:
>     * add a new JSON schema and examples (migrate the snippet in the spec) under https://github.com/ome/ngff/tree/main/0.4
>     * make the decisions on the distribution mechanism @will-moore was working this in

Thanks for the pointers, @sbesson. I had a look at the two referenced issues and left comments. It would be great to get this in (and then use it in the tests or directly in the writers), but since the PRs seem to be stalled a bit it's probably better to go ahead and merge this issue first.

sbesson:
Thanks for adding tests and fixing the linting issue. I agree not tie the API addition to this work and focus on getting it released so that we can gain experience consuming. I'll review the tests and the API more thoroughly tomorrow.

tests/test_writer.py
```diff
@@ -823,3 +824,115 @@ def test_unspecified_images_keys(self):
...
+                node.metadata["coordinateTransformations"], transformations
+            ):
+                assert transf == expected
+            assert len(node.metadata["coordinateTransformations"]) == len(node.data)
+        assert np.allclose(label_data, node.data[0][...].compute())
+
+        # Verify label metadata
+        label_root = zarr.open(f"{self.path}/labels", "r")
+        assert "labels" in label_root.attrs
+        assert label_name in label_root.attrs["labels"]
```
sbesson, line 935: Should we also ensure that `label_name` is correctly added to the `multiscales` metadata ?

sbesson (changes requested):
A few questions regarding the inferred types of arguments. Otherwise, I think the new API makes sense and allows to support the writing of extended metadata. The new API should already populate the recommended `name` attribute. We can add additional tests covering more advanced scenarios such as examples compliant with the recommended specification in a follow-up PR.

The default test should cover the basic scenario of creating an image and an associated label image. Could we add another minimal scenario that creates two labels and checks that the attributes are correctly created e.g. the `labels` attributes register both labels?
Similarly as above, follow-up PRs can add coverage for advanced `image-label` properties such as `colors` and `properties`.

constantinpape:
Thanks for the review, @sbesson.

> The new API should already populate the recommended name attribute.

I see two options for this: (i) add `name` as a required argument to `write_image`, `write_multiscale` and `write_multiscales_metadata`. (ii) check if it's in `**metadata` in `write_multiscales_metadata` and if not choose a sensible default. Which option do you prefer? (I would vote for (ii) because it's less invasive, but I don't have a strong opinion here).

>  Could we add another minimal scenario that creates two labels and checks that the attributes are correctly created e.g. the labels attributes register both labels?

Sounds good, I will add it now.

For some reason I can't comment under the code comments you made (seems to be some github issue), so I will respond to the relevant ones here:

> The addition of Union[str is a bit confusing. I know it was flagged by mypy but it would be good to know whether there is a configuration where a single string can be passed of it's a bug or a red herring

Adding `str` is necessary and we should probably also add `int` and `float`. Note that mypy evaluates a the individual **value** passed as kwarg and not the whole kwargs object. See https://adamj.eu/tech/2021/05/11/python-type-hints-args-and-kwargs/ for details.

> Would label_metadata: JSONDict = {}, potentially reduce the conditional block in https://github.com/ome/ome-zarr-py/pull/178/files#diff-d51ad4cb28d657d5e09c58f40674b95273611983353ab5d7bb2049bbc6303f7bR545

This is considered bad practice in python: https://docs.quantifiedcode.com/python-anti-patterns/correctness/mutable_default_value_as_argument.html

joshmoore:
@constantinpape: I found https://github.com/codecov/codecov-action/issues/557 which led to https://github.com/marketplace/actions/codecov

> ⚠️ Deprecration of v1
On February 1, 2022, this version will be fully sunset and no longer function

https://github.com/marketplace/actions/codecov#migration-from-v1-to-v2

sbesson (approved):
All looks good to me and I am inclined to have this new functionality available sooner rather than later so that we can start consuming it and potentially amending it based on real use cases.
Should we aim to release this immediately as `ome-zarr 0.4.0`?

constantinpape:
> All looks good to me and I am inclined to have this new functionality available sooner rather than later so that we can start consuming it and potentially amending it based on real use cases.
> Should we aim to release this immediately as `ome-zarr 0.4.0`?

Thanks! I agree that we should have this out as soon as possible. But I think it would be a good idea to have a look at fixing and simplifying the `write_image` API first. So I suggest the following:
- merge this PR now
- I make a follow-up PR with API clean-ups (this will be fast, I can do it as soon as this one is merged)
- You have a look at the follow up and we see if the changes are controversial or not. If they are, we release 0.4.0 without it, if they are not we can still include it before the new release.

PR #121, opened 2021-10-21
Add create and write_multiscale() usage to README

will-moore:
When compared to other OME-NGFF writing tools
(see https://forum.image.sc/t/ndtiff-to-ome-zarr-ngff/58856/2) I notice that we don't provide good docs on how to write OME-NGFF.

This adds code example to the README.
It also adds the `ome_zarr create` command to the README.

To test, try running the code and opening in `napari` as described.
Should get this:

Screenshot 2021-10-21 at 09 54 33

PR #181, opened 2022-03-24
Add all ome_zarr modules to the generated API documentation

sbesson:
Builds on top of #121 and exposes all the modules as single pages


sbesson:
See https://ome-zarr--181.org.readthedocs.build/en/181/api.html for the documentation built from this PR

will-moore:
Looks good!
The `@param` are not formatted correctly at https://ome-zarr--181.org.readthedocs.build/en/181/api/csv.html#ome_zarr.csv.csv_to_zarr

sbesson:
:+1: updated all remaining usages of `@param` to `:param` - see https://ome-zarr--181.org.readthedocs.build/en/181/api/csv.html. The writer API uses a standard dictionary style for the parameters and could also use this markup but I would propose not to update these before #178 is merged as new methods are introduced

will-moore:
Thanks. The `writer API` style looks good too, and formats well with the longer param descriptions, so hopefully it'll work OK with `:param` too.

will-moore (approved):
LGTM 👍

PR #184, opened 2022-03-24
Update write_image signature

constantinpape:
This PR updates the signatures of `write_image`, `write_multiscale` and `write_multiscale_labels`:
- Removes `byte_order` from `write_image`: this was not used at all
- Removes `chunks` from all of them to avoid duplication with `storage_options`. 
- Moves `retuple` from `write_image` to `write_multiscale`, which is where the "actual" writing of data happens.

If we agree on these changes I will also add `write_labels` here (corresponding to `write_image`, but for image label data).

cc @will-moore @sbesson @joshmoore

### DIFF
--- a/docs/source/python.rst
+++ b/docs/source/python.rst
@@ -35,7 +35,7 @@ The following code creates a 3D Image in OME-Zarr with labels::
     # write the image data
     store = parse_url(path, mode="w").store
     root = zarr.group(store=store)
-    write_image(image=data, group=root, chunks=(1, size_xy, size_xy), axes="zyx")
+    write_image(image=data, group=root, iaxes="zyx", storage_options=dict(chunks=(1, size_xy, size_xy)))
     # optional rendering settings
     root.attrs["omero"] = {
         "channels": [{
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+# 0.4.0 (Mar 2022)
+
+- Add API for writing data according to the `image-label` spec: `write_labels`, `write_multiscale_labels` & `write_label_metadata` ([#178](https://github.com/ome/ome-zarr-py/pull/178), [#184](https://github.com/ome/ome-zarr-py/pull/184))
+- Remove `byte_order` from `write_image` signature ([#184](https://github.com/ome/ome-zarr-py/pull/184))
+- Deprecate `chunks` argument in favor of `storage_options` ([#184](https://github.com/ome/ome-zarr-py/pull/184))
+
 # 0.3.0 (Feb 2022)
 
 ## Specification changes
--- a/ome_zarr/data.py
+++ b/ome_zarr/data.py
@@ -135,7 +135,8 @@ def create_zarr(
             if zct + 2 < len(chunks):
                 chunks[zct] = 1
 
-    write_multiscale(pyramid, grp, chunks=tuple(chunks), axes=axes)
+    storage_options = dict(chunks=tuple(chunks))
+    write_multiscale(pyramid, grp, axes=axes, storage_options=storage_options)
 
     if size_c == 1:
         image_data = {
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -363,7 +363,6 @@ def write_image(
     image: np.ndarray,
     group: zarr.Group,
     chunks: Union[Tuple[Any, ...], int] = None,
-    byte_order: Union[str, List[str]] = "tczyx",
     scaler: Scaler = Scaler(),
     fmt: Format = CurrentFormat(),
     axes: Union[str, List[str], List[Dict[str, str]]] = None,
@@ -382,9 +381,6 @@ def write_image(
       the group within the zarr store to store the data in
     chunks: int or tuple of ints,
       size of the saved chunks to store the image
-    byte_order: str or list of str, default "tczyx"
-      combination of the letters defining the order
-      in which the dimensions are saved
     scaler: Scaler
       Scaler implementation for downsampling the image argument. If None,
       no downsampling will be performed.
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -414,31 +414,7 @@ def write_image(
       the number of datasets in a multiresolution pyramid. One can provide
       different chunk size for each level of a pyramind using this option.
     """
-
-    if image.ndim > 5:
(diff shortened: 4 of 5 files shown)


constantinpape:
Re deprecation warning: would you be ok with leaving the argument, but not using it at all and raise a warning that states that `storage_options` should be used? I would be on-board with that, but I would rather avoid implementing any consolidation of `storage_options` and chunks.

sbesson:
> Re deprecation warning: would you be ok with leaving the argument, but not using it at all and raise a warning that states that storage_options should be used?

Tricky. When deprecating,I am definitely on board with a loud warning to tell people to upgrade their code, ideally with an idea of when the feature will be dropped. 
In all cases, I would be very cautious about making an existing argument a no-op as that would effectively mean change the behavior of existing code using `chunks` and potentially invalidate data generation.
Actually due to the `**metadata` argument, I am correctly assuming that in the current state of this proposal

```
write_image(image=data, group=root, chunks=chunks, axes="zyx")
```

would use the default chunks but write `chunks` as a key in the `multiscales` metadata?

constantinpape:
Hmm, my problem with this is that the current code is only semi-functional. For example passing chunks via `storage_options` will not `retuple` it and thus likely fail when writing `0.1` and `0.2`.

sbesson:
Definitely worth fixing the retupling of `chunks` via `storage_options` as this is our preferred API moving forward.
Trying to lay down the different scenarios and at least my preference in terms of behavior


```
write_image(image=data, group=root)
```

Expected behavior: unchanged

```
write_image(image=data, group=root, chunks=chunks)
```

Expected behavior: deprecation warning, retuple `chunks` and pass it to `create_dataset`


```
write_image(image=data, group=root, storage_options={"chunks": chunks})
```

Expected behavior: retuple `stoage_options["chunks"]`  and pass it to `create_dataset`


```
write_image(image=data, group=root, chunks=chunks, storage_options={"chunks": chunks})
```

Expected behavior: deprecation warning, ignore `chunks`, retuple `stoage_options["chunks"]` and pass it to `create_dataset`

I would think the only change with your implementation is that https://github.com/ome/ome-zarr-py/pull/184/files#diff-d51ad4cb28d657d5e09c58f40674b95273611983353ab5d7bb2049bbc6303f7bR219 would not unconditionally default to `None` but to the retupled version of `chunks` if it is passed as an explicit keyword argument.

Also happy to hear @joshmoore or @will-moore's opinions on the above

constantinpape:
Ok, I added back the chunk argument in a way that should specify your requirements, @sbesson.

constantinpape:
@sbesson I have added `write_labels` now (and refactored code a bit to reuse the mip creation from `write_image`).
There's also a test for it.

This is good to go from my side, pending the decision on whether we indeed want to remove `chunks` in 0.5; if not the deprecation warning has to be updated.

docs/source/python.rst
```diff
@@ -35,7 +35,7 @@ The following code creates a 3D Image in OME-Zarr with labels::
     # write the image data
     store = parse_url(path, mode="w").store
     root = zarr.group(store=store)
-    write_image(image=data, group=root, chunks=(1, size_xy, size_xy), axes="zyx")
+    write_image(image=data, group=root, iaxes="zyx", storage_options=dict(chunks=(1, size_xy, size_xy)))
```
sbesson, line 38:
```suggestion
    write_image(image=data, group=root, axes="zyx", storage_options=dict(chunks=(1, size_xy, size_xy)))
```

docs/source/python.rst
```diff
@@ -35,7 +35,7 @@ The following code creates a 3D Image in OME-Zarr with labels::
     # write the image data
     store = parse_url(path, mode="w").store
     root = zarr.group(store=store)
-    write_image(image=data, group=root, iaxes="zyx", storage_options=dict(chunks=(1, size_xy, size_xy)))
+    write_image(image=data, group=root, axes="zyx", storage_options=dict(chunks=(1, size_xy, size_xy)))
```
constantinpape, line 38: Thanks for catching this

sbesson (approved):
Nice.

I like the extraction of the pyramid creation logic into a sub-function which clarifies a lot of the contract of `write_image` vs `write_multiscale` (at the code level at least). I assume at some point, `_create_mip` could be turned into a public API for more advanced workflows but that's outside the scope of this PR.

The deprecation of `chunks` makes sense (while keeping the implementation functional) and I think it's fair enough to anticipate and warn we are planning to remove it in the next `0.x.0` release.

No objection from my side to getting this in. I'll leave people mentioned in the description and anyone interested until EOB to express secondary options and merge if no significant objection has been raised. I might follow-up with a quick 

In terms of release, I am away until next Thursday. Anyone interested in carrying out the release early next week? I think it's as simple as updating the CHANGELOG, running `bump2version` and pushing the tag but we might need to sort out permissions and optionally review the instructions.

constantinpape:
> In terms of release, I am away until next Thursday. Anyone interested in carrying out the release early next week? I think it's as simple as updating the CHANGELOG, running `bump2version` and pushing the tag but we might need to sort out permissions and optionally review the instructions.

I could do it on Monday, but I currently don't have any permissions for this repo. But I am also happy to let @will-moore or @joshmoore go ahead with this. (But from my side it would be great to make the release available on Monday since I would like to finish the conversion scripts for Shila's data next week and having this version released would help a lot with this.)

will-moore:
Tested the latest code with https://github.com/ome/omero-cli-zarr/pull/107 and looks good.
Happy to release once this is merged.
@constantinpape could you update the CHANGELOG.md in this PR?

constantinpape:
> Tested the latest code with [ome/omero-cli-zarr#107](https://github.com/ome/omero-cli-zarr/pull/107) and looks good.

Thanks!

> Happy to release once this is merged.
> @constantinpape could you update the CHANGELOG.md in this PR?

I added the changes from my two PRs to the changelog now.

joshmoore:
> I could do it on Monday, but I currently don't have any permissions for this repo. But I am also happy to let @will-moore or @joshmoore go ahead with this.

Also added an ome/ome-zarr-py team with @constantinpape. :+1:

constantinpape:
Thanks for the reviews here and also for adding me to the repo :).
@will-moore: do you think you can make the release today?

will-moore:
Done! https://pypi.org/project/ome-zarr/

sbesson:
Thanks all for driving this

PR #185, opened 2022-03-26
Convert `ome_zarr.writer.write_multiscale` to use docstrings markup

sbesson:
This proposes to use the standard docstring markup (`param`, `type`) to unify the rendering in the published documentation. Also includes the `deprecated` markup for deprecated options.

Initially applied to `write_multiscales` for a first round of review, the other docstrings in the `writer` module can be unified if we all agree with this strategy.

Proposed tag:  this is no a blocker for the imminent release of `0.4.0`  and should be releasable as a `0.4.1`

### DIFF
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -181,30 +181,37 @@ def write_multiscale(
     """
     Write a pyramid with multiscale metadata to disk.
 
-    pyramid: List of np.ndarray
-      the image data to save. Largest level first
-      All image arrays MUST be up to 5-dimensional with dimensions
-      ordered (t, c, z, y, x)
-    group: zarr.Group
-      the group within the zarr store to store the data in
-    chunks: int or tuple of ints,
-      Size of the saved chunks to store the image.
-      This argument is deprecated and will be removed in a future version.
-      Use storage_options instead.
-    fmt: Format
-      The format of the ome_zarr data which should be used.
-      Defaults to the most current.
-    axes: str or list of str or list of dict
-      List of axes dicts, or names. Not needed for v0.1 or v0.2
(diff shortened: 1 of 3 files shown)


sbesson:
See https://ome-zarr--185.org.readthedocs.build/en/185/api/writer.html#ome_zarr.writer.write_multiscale for the rendering of this proposal. To be compared with https://ome-zarr.readthedocs.io/en/latest/api/writer.html#ome_zarr.writer.write_multiscale

constantinpape (approved):

sbesson:
The last commit reviews all docstrings of the `writer` module - see https://ome-zarr--185.org.readthedocs.build/en/185/api/writer.html.

Proposing to get this out as `0.4.1`

ome_zarr/writer.py
```diff
@@ -181,30 +181,38 @@ def write_multiscale(
...
-      or if 2D. Otherwise this must be provided
-    coordinate_transformations: 2Dlist of dict
-      For each path, we have a List of transformation Dicts.
-      Each list of dicts are added to each datasets in order
-      and must include a 'scale' transform.
-    storage_options: dict or list of dict
-      Options to be passed on to the storage backend. A list would need to match
-      the number of datasets in a multiresolution pyramid. One can provide
-      different chunk size for each level of a pyramind using this option.
+    :type pyramid: list of :class:`np.ndarray`
```
will-moore, line 184: The `np.ndarray` class here and elsewhere don't create links to the `numpy` docs, whereas `numpy.ndarray` probably would?

will-moore (approved):
Great, thx 👍

sbesson:
Now deployed to https://ome-zarr.readthedocs.io/en/stable/api/writer.html

PR #189, opened 2022-04-13
Relax version detection

joshmoore:
Some writers (like for the DANDI project) have stored
the versions as floating point values (e.g. `0.2`).
Though this won't be long-term sustainable, relaxing
the version identification at this point seems
unproblematic.

cc: @leekamentsky

### DIFF
--- a/ome_zarr/format.py
+++ b/ome_zarr/format.py
@@ -12,6 +12,11 @@ LOGGER = logging.getLogger("ome_zarr.format")
 def format_from_version(version: str) -> "Format":
 
     for fmt in format_implementations():
+
+        # Support floating-point versions like `0.2`
+        if isinstance(version, float):
+            version = str(version)
+
         if fmt.version == version:
             return fmt
     raise ValueError(f"Version {version} not recognized")


LeeKamentsky:
I've updated those files to "0.4" as a string. My bad. Hopefully I am the only one.

ome_zarr/format.py
```diff
@@ -12,6 +12,11 @@
 def format_from_version(version: str) -> "Format":
 
     for fmt in format_implementations():
+
+        # Support floating-point versions like `0.2`
+        if isinstance(version, float):
+            version = str(version)
```
sbesson, line 18: While this is primarily consumed at the reader level currently and hence quite downstream of the generation process, this raises the question of whether this library should also raise a `WARN` level logging statement indicating the mismatch.

joshmoore: Possibly, though I'm already annoyed by the format mismatch warnings. Perhaps we could differentiate between a more purely user (`view()`) and developer (`validate()`) context.

sbesson: I am quite annoyed by the format mismatch warning as well - happy to create an issue or even look into it 😄 

In this scenario, the the warning is about a core issue with the data validation rather than exposing the internal guessing logic of the library so assuming I had to pick between the two, I would personally only keep the former. The idea of differentiating the usage contexts is interesting and possibly brings us back to the validation flag that @will-moore introduced while working on https://github.com/ome/ome-zarr-py/pull/189.

joshmoore: I've found why it's duplicated and will open a PR. We'll still need to decide on the _first_ mismatch.

will-moore:
LGTM 👍

joshmoore:
Thanks, @will-moore. Anyone have an opinion on the warning? Or shall I just record the fact (in code or an issue) that we may want to warn here in the future when there's a clear "dev" situation?

sbesson:
As per the [JSON schema](https://github.com/ome/ngff/blob/caa7e8eb83291947e4168539d30a7afc967f6f4f/0.4/schemas/image.schema#L33-L37), my assumption is that these datasets should throw a warning with the `--validate` option proposed in #142?


If so, no objection to getting this in its current form and the warning should naturally be appear as part of the validation framework.

PR #190, opened 2022-04-13
Fix format warning duplication

joshmoore:
The call graph that led to duplicated WARNING messages of the form:

```
WARNING:ome_zarr.io:version mismatch: detected:FormatV02, requested:FormatV04
WARNING:ome_zarr.io:version mismatch: detected:FormatV04, requested:FormatV02
```
where caused by (roughly) this sequence of calls:
 * Multiscales created using the default CurrentFormat (V04)
 * Metadata of Multiscales is checked (V02) leading to the first warning
 * The call to `zarr.create("labels")` initializes a Labels object
 * It has no metadata and so defaults to the CurrentFormat (V04)
 * This then mismatches from the expected format (V02) leading to the second

We now simply require that a value is passed for when the object does not contain metadata.

### DIFF
--- a/ome_zarr/format.py
+++ b/ome_zarr/format.py
@@ -27,10 +27,10 @@ def format_implementations() -> Iterator["Format"]:
     yield FormatV01()
 
 
-def detect_format(metadata: dict) -> "Format":
+def detect_format(metadata: dict, default: "Format") -> "Format":
     """
     Give each format implementation a chance to take ownership of the
-    given metadata. If none matches, a CurrentFormat is returned.
+    given metadata. If none matches, the default value will be returned.
     """
 
     if metadata:
@@ -38,7 +38,7 @@ def detect_format(metadata: dict) -> "Format":
             if fmt.matches(metadata):
                 return fmt
 
-    return CurrentFormat()
+    return default
 
 
 class Format(ABC):
--- a/ome_zarr/io.py
+++ b/ome_zarr/io.py
@@ -46,7 +46,7 @@ class ZarrLocation:
         self.__store = loader.init_store(self.__path, mode)
 
         self.__init_metadata()
-        detected = detect_format(self.__metadata)
+        detected = detect_format(self.__metadata, loader)
         if detected != fmt:
             LOGGER.warning(f"version mismatch: detected:{detected}, requested:{fmt}")
             self.__fmt = detected


will-moore:
Before this PR:
```
$ ome_zarr info https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001247.zarr
WARNING:ome_zarr.io:version mismatch: detected:FormatV01, requested:FormatV04
WARNING:ome_zarr.io:version mismatch: detected:FormatV04, requested:FormatV01
WARNING:ome_zarr.io:version mismatch: detected:FormatV01, requested:FormatV04
WARNING:ome_zarr.io:version mismatch: detected:FormatV04, requested:FormatV01
https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001247.zarr/ [zgroup]
...
```

With this PR:
```
$ ome_zarr info https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001247.zarr
WARNING:ome_zarr.io:version mismatch: detected:FormatV01, requested:FormatV04
https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001247.zarr/ [zgroup]
...
```

LGTM 👍

sbesson (approved):
Sorry, forgot to comment but similar result as @will-moore and overall :+1: for fixing the annoying redudant format detection mismatch warning.

As demonstrated by Will's example, the remaining warning is related to the fact the default format is set to `CurrentFormat` (`v4` currently) so any sample with an earlier version will throw.

If we wanted to get rid of it completely at least in the context of the `info` command , an option would be

```diff
(base) sbesson@Sebastiens-MacBook-Pro ome-zarr-py % git diff
diff --git a/ome_zarr/io.py b/ome_zarr/io.py
index 563d31b..ad2b14b 100644
--- a/ome_zarr/io.py
+++ b/ome_zarr/io.py
@@ -48,7 +48,8 @@ class ZarrLocation:
         self.__init_metadata()
         detected = detect_format(self.__metadata, loader)
         if detected != fmt:
-            LOGGER.warning(f"version mismatch: detected:{detected}, requested:{fmt}")
+            if fmt is not None:
+                LOGGER.warning(f"version mismatch: detected:{detected}, requested:{fmt}")
             self.__fmt = detected
             self.__store = detected.init_store(self.__path, mode)
             self.__init_metadata()
diff --git a/ome_zarr/utils.py b/ome_zarr/utils.py
index 21aa92b..87d10c0 100644
--- a/ome_zarr/utils.py
+++ b/ome_zarr/utils.py
@@ -23,7 +23,7 @@ def info(path: str, stats: bool = False) -> Iterator[Node]:
     All :class:`Nodes <ome_utils.reader.Node>` that are found from the given path will
     be visited recursively.
     """
-    zarr = parse_url(path)
+    zarr = parse_url(path, fmt=None)
     assert zarr, f"not a zarr: {zarr}"
     reader = Reader(zarr)
     for node in reader():
```

i.e. explicitly set `fmt` to `None`. The loader will still be intialised using `CurrentFormat` and report a mismatch but the logging warning is skipped explicitly.

joshmoore:
Do we want to propagate that out to a user option?

sbesson:
Possibly although on this one, I am fairly opiniated that the current behavior is wrong. If I try to inspect a valid OME-NGFF 0.3 dataset using `ome_zarr info`, receiving a WARN-level statement telling me there is a mismatch wth a version I did not specify and cannot override feels like noise. On the other hand, a INFO-level statement indicating me which version was detected would be completely appropriate.

Since the proposed changes here were reviewed and agreed without modifications, I think it makes sense to merge this (and #189) and get it into a patch release and I can open a separate PR with the change above where we can start the discussion around options

Issue #187, opened 2022-04-02
Appropriate way of writing an larger-than-memory image? 


BioinfoTongLI:
Hello there!

I am having some troubles writing an large image using `write_image`.
If I understand correctly, as `write_image` takes numpy array as input.
It assumes that the image should always be loaded into the memory first, right?

Is there a way of writing images that are not possible to be loaded into the memory? i.e. writing chunks into zarr incrementally? Potentially using `dask.array` as input instread of `np.ndarray`?

Best,
Tong

will-moore:
There is an issue when using `dask.array` instead of `nd.array` to write to zarr. See https://github.com/zarr-developers/zarr-python/issues/962.

In `omero-cli-zarr`, when generating OME-NGFF from OMERO, we write a 2D-plane at a time to each level of a pyramid. See https://github.com/ome/omero-cli-zarr/blob/db58446bf4123bb02307f62748941d34abbbb358/src/omero_zarr/raw_pixels.py#L110

We only use `ome-zarr-py` for `write_multiscales_metadata()`.

BioinfoTongLI:
Thanks! @will-moore .
So, if I understand correctly, the ultimate solution is use `dask.array` as default, once that issues is solved? And the plane-by-plane in `omero-cli-zarr` is a temporary walk-around?

BioinfoTongLI:
That's good to know! Thanks for the explanation!
Then do you think adding a function to this package, as the one in `omero-cli-zarr`, a sensible thing to do?
Potentially that means to add a function that `append_plane/image` to an existing zarr store, given the metadata asscociated?

joshmoore:
Not having dug into the conversation above, :+1: for more hands adding to this library! 🙏🏽

toloudis:
I'm running into this too, writing a proof of concept OmeZarrWriter for aicsimageio (https://github.com/AllenCellModeling/aicsimageio/blob/feature/zarrwriter/aicsimageio/writers/ome_zarr_writer.py). I'm now at a point where I want to test on a large time series and I can't possibly load all of TCZYX into memory, but I _can_ load CZYX.   Just getting started here but thought I'd add my use case to the mix.

joshmoore:
[...]

PR #192, opened 2022-04-15
Feature/dask poc

toloudis:
Admittedly I am an extremely naive dask user but I'm taking a stab at this because I really want to pass a large dask array in to the ome zarr writer here.  

I wonder if anyone can chime in on how to make the resize work inside of Scaler - or indeed how to make this code more correct.  This is heavily WIP just to try to get anything working at first.  (For the array I'm passing in now, there's some size broadcast mismatch error when it gets to `compute` in the resizing stuff.)

I am loading some data using aicsimageio's get_image_dask_data and then basically forwarding it into this writer (through the WIP aicsimageio.writers.OmeZarrWriter)

Possibly addressing #187

### DIFF
--- a/ome_zarr/scale.py
+++ b/ome_zarr/scale.py
@@ -28,6 +28,12 @@ ListOfArrayLike = Union[List[dask.array.Array], List[np.ndarray]]
 ArrayLike = Union[dask.array.Array, np.ndarray]
 
 
+def resize_wrapper(array, **kwargs):
+    print("resize_wrapper...", array.shape, kwargs)
+    rsp = resize(array, **kwargs)
+    print("resize_wrapper rsp", rsp.shape)
+    return rsp
+
 @dataclass
 class Scaler:
     """Helper class for performing various types of downsampling.
@@ -148,9 +154,13 @@ class Scaler:
     def __nearest(self, plane: ArrayLike, sizeY: int, sizeX: int) -> ArrayLike:
         """Apply the 2-dimensional transformation."""
         if isinstance(plane, dask.array.Array):
+            # print("downscale_nearest...", plane.shape, plane.dtype)
+            # return downscale_nearest(plane, factors=(self.downscale, self.downscale))
+            print("resize...", self.downscale, sizeX, sizeY)
             outsize = (np.ceil(np.array([sizeY, sizeX]) / self.downscale)).astype(int)
+            print("outsize", outsize)
             resized = dask.array.map_blocks(
-                resize,
+                resize_wrapper,
                 plane,
                 output_shape=(outsize[0], outsize[1]),
                 order=0,
@@ -159,6 +169,7 @@ class Scaler:
                 dtype=plane.dtype,
                 chunks=(outsize[0], outsize[1]),
             )
+            print("__nearest resized", resized.shape)
             return resized
         else:
             return resize(
@@ -227,6 +238,9 @@ class Scaler:
             stack_to_scale = rv[-1]
             shape_5d = (*(1,) * (5 - stack_to_scale.ndim), *stack_to_scale.shape)
             T, C, Z, Y, X = shape_5d
+            print("_by_plane stack_to_scale.shape", stack_to_scale.shape)
+            print("_by_plane prev shape_5d", shape_5d)
+            print("_by_plane, Y, X", Y, X)
 
             # If our data is already 2D, simply resize and add to pyramid
             if stack_to_scale.ndim == 2:
--- a/ome_zarr/scale.py
+++ b/ome_zarr/scale.py
@@ -162,9 +162,7 @@ class Scaler:
         """
         return self._by_plane(base, self.__nearest)
 
-    def __nearest(
-        self, plane: Union[np.ndarray, da.Array], sizeY: int, sizeX: int
-    ) -> np.ndarray:
+    def __nearest(self, plane: ArrayLike, sizeY: int, sizeX: int) -> np.ndarray:
         """Apply the 2-dimensional transformation."""
         if isinstance(plane, da.Array):
 
--- a/ome_zarr/scale.py
+++ b/ome_zarr/scale.py
@@ -148,14 +148,17 @@ class Scaler:
     def __nearest(self, plane: ArrayLike, sizeY: int, sizeX: int) -> ArrayLike:
         """Apply the 2-dimensional transformation."""
         if isinstance(plane, dask.array.Array):
-            outsize = (np.ceil(np.array([sizeY, sizeX])/self.downscale)).astype(int)
-            resized = dask.array.map_blocks(resize, plane,
+            outsize = (np.ceil(np.array([sizeY, sizeX]) / self.downscale)).astype(int)
+            resized = dask.array.map_blocks(
+                resize,
+                plane,
                 output_shape=(outsize[0], outsize[1]),
                 order=0,
                 preserve_range=True,
                 anti_aliasing=False,
                 dtype=plane.dtype,
-                chunks=(outsize[0], outsize[1]))
+                chunks=(outsize[0], outsize[1]),
+            )
             return resized
         else:
             return resize(
--- a/ome_zarr/scale.py
+++ b/ome_zarr/scale.py
@@ -221,7 +221,7 @@ class Scaler:
         print(base.shape)
         for i in range(self.max_layer):
             print(i, self.downscale)
-            rv.append(zoom(base, self.downscale ** i))
+            rv.append(zoom(base, self.downscale**i))
             print(rv[-1].shape)
         return list(reversed(rv))
 
--- a/ome_zarr/scale.py
+++ b/ome_zarr/scale.py
@@ -162,9 +162,21 @@ class Scaler:
         """
         return self._by_plane(base, self.__nearest)
 
-    def __nearest(self, plane: np.ndarray, sizeY: int, sizeX: int) -> np.ndarray:
+    def __nearest(
+        self, plane: Union[np.ndarray, da.Array], sizeY: int, sizeX: int
+    ) -> np.ndarray:
         """Apply the 2-dimensional transformation."""
-        return resize(
+        if isinstance(plane, da.Array):
+
+            def _resize(
+                image: ArrayLike, output_shape: Tuple, **kwargs: Any
+            ) -> ArrayLike:
+                return dask_resize(image, output_shape, **kwargs)
+
+        else:
+            _resize = resize
+
+        return _resize(
             plane,
             output_shape=(sizeY // self.downscale, sizeX // self.downscale),
             order=0,
@@ -209,7 +221,7 @@ class Scaler:
         print(base.shape)
         for i in range(self.max_layer):
             print(i, self.downscale)
-            rv.append(zoom(base, self.downscale**i))
+            rv.append(zoom(base, self.downscale ** i))
             print(rv[-1].shape)
         return list(reversed(rv))
 
--- a/ome_zarr/scale.py
+++ b/ome_zarr/scale.py
@@ -13,15 +13,10 @@ import dask.array as da
 import numpy as np
 import zarr
 from scipy.ndimage import zoom
-from skimage.transform import (
-    downscale_local_mean,
-    pyramid_gaussian,
-    pyramid_laplacian
-)
-
+from skimage.transform import downscale_local_mean, pyramid_gaussian, pyramid_laplacian
 from skimage.transform import resize as skimage_resize
-from .dask_utils import resize as dask_resize
 
+from .dask_utils import resize as dask_resize
 from .io import parse_url
 
 LOGGER = logging.getLogger("ome_zarr.scale")
--- a/ome_zarr/scale.py
+++ b/ome_zarr/scale.py
@@ -147,12 +147,14 @@ class Scaler:
     def __nearest(self, plane: ArrayLike, sizeY: int, sizeX: int) -> ArrayLike:
         """Apply the 2-dimensional transformation."""
         if isinstance(plane, dask.array.Array):
+            outsize = (np.ceil(np.array([sizeY, sizeX])/self.downscale)).astype(int)
             resized = dask.array.map_blocks(resize, plane,
-                output_shape=(sizeY // self.downscale, sizeX // self.downscale),
+                output_shape=(outsize[0], outsize[1]),
                 order=0,
                 preserve_range=True,
                 anti_aliasing=False,
-                dtype=plane.dtype)
+                dtype=plane.dtype,
+                chunks=(outsize[0], outsize[1]))
             # resized_astype = dask.delayed(dask.array.Array.astype)(resized, plane.dtype)
             return resized
         else:
--- a/ome_zarr/scale.py
+++ b/ome_zarr/scale.py
@@ -27,6 +27,7 @@ LOGGER = logging.getLogger("ome_zarr.scale")
 ListOfArrayLike = Union[List[dask.array.Array], List[np.ndarray]]
 ArrayLike = Union[dask.array.Array, np.ndarray]
 
+
 @dataclass
 class Scaler:
     """Helper class for performing various types of downsampling.
@@ -72,7 +73,7 @@ class Scaler:
                 continue
             yield name
 
-    def scale_array(self, input_array: ArrayLike ) -> ListOfArrayLike:
+    def scale_array(self, input_array: ArrayLike) -> ListOfArrayLike:
         """Perform downsampling to memory."""
         func = getattr(self, self.method, None)
         if not func:
@@ -155,7 +156,6 @@ class Scaler:
                 anti_aliasing=False,
                 dtype=plane.dtype,
                 chunks=(outsize[0], outsize[1]))
-            # resized_astype = dask.delayed(dask.array.Array.astype)(resized, plane.dtype)
             return resized
         else:
             return resize(
@@ -245,8 +245,10 @@ class Scaler:
                             zct_dims = shape_5d[:-2]
                             shape_dims = zct_dims[-stack_dims:]
                             if isinstance(out, dask.array.Array):
-                                new_stack = dask.array.zeros((*shape_dims, out.shape[0], out.shape[1]),
-                                    dtype=base.dtype,)
+                                new_stack = dask.array.zeros(
+                                    (*shape_dims, out.shape[0], out.shape[1]),
+                                    dtype=base.dtype,
+                                )
                             else:
                                 new_stack = np.zeros(
                                     (*shape_dims, out.shape[0], out.shape[1]),
--- a/ome_zarr/scale.py
+++ b/ome_zarr/scale.py
@@ -7,8 +7,9 @@ import logging
 import os
 from collections.abc import MutableMapping
 from dataclasses import dataclass
-from typing import Callable, Iterator, List
+from typing import Callable, Iterator, List, Union
 
+import dask
 import numpy as np
 import zarr
 from scipy.ndimage import zoom
@@ -23,6 +24,8 @@ from .io import parse_url
 
 LOGGER = logging.getLogger("ome_zarr.scale")
 
+ListOfArrayLike = Union[List[dask.array.Array], List[np.ndarray]]
+ArrayLike = Union[dask.array.Array, np.ndarray]
 
 @dataclass
 class Scaler:
@@ -68,6 +71,14 @@ class Scaler:
             if name.startswith("_"):
                 continue
             yield name
+
+    def scale_array(self, input_array: ArrayLike ) -> ListOfArrayLike:
+        """Perform downsampling to memory."""
+        func = getattr(self, self.method, None)
+        if not func:
+            raise Exception
+        ret = func(input_array)
+        return ret
 
     def scale(self, input_array: str, output_directory: str) -> None:
         """Perform downsampling to disk."""
@@ -125,7 +136,7 @@ class Scaler:
             series.append({"path": path})
         return grp
 
-    def nearest(self, base: np.ndarray) -> List[np.ndarray]:
+    def nearest(self, base: ArrayLike) -> ListOfArrayLike:
         """
         Downsample using :func:`skimage.transform.resize`.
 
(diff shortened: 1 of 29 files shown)


ome_zarr/scale.py
```diff
@@ -220,10 +245,16 @@ def _by_plane(
                         if new_stack is None:
                             zct_dims = shape_5d[:-2]
                             shape_dims = zct_dims[-stack_dims:]
-                            new_stack = np.zeros(
-                                (*shape_dims, out.shape[0], out.shape[1]),
-                                dtype=base.dtype,
-                            )
+                            if isinstance(out, dask.array.Array):
+                                new_stack = dask.array.zeros(
+                                    (*shape_dims, out.shape[0], out.shape[1]),
```
toloudis, line 250: the use of out.shape here is questionable because the result of resize was delayed

toloudis:
after the last couple of commits, the writer seems to be working in my simple test case (only the "nearest" method has been modified to use dask in the Scaler)

ome_zarr/scale.py
toloudis: this is now resolved because the dask chunks were set appropriately

toloudis:
calling compute_mip before trying to store anything out is potentially problematic for very large data even in dask land.

joshmoore:
cc: @jakirkham here just in case he has any https://data-apis.org/ insight.

toloudis:
the delayed create_mip is definitely slow with dask and is one of those steps where it appears like nothing is happening in between writing out levels.  something seems inefficient, like maybe we could process all levels for each plane while we have it in memory, but then the write pattern changes completely.  on the other hand, maybe this is as good as it gets?

will-moore:
I tested a simple case using sample code at https://ome-zarr.readthedocs.io/en/stable/python.html#writing-ome-ngff-images
but with `data = da.from_array(data)` and this seems to work fine.

I tried a more 'real' example, loading data from "https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001247.zarr" (see below). This takes a while and I end up with the `6001247.zarr` dir, including `.zattrs` and resolutions dirs `/0` - `/4`, each containing only the `.zarray` but no chunks.

```
import dask.array as da
import zarr
from zarr.storage import FSStore
import os

url = "https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001247.zarr"
kwargs = {
    "dimension_separator": "/",
    "normalize_keys": False,
}

store = FSStore(url, mode='r', **kwargs)
data = da.from_zarr(store, '0')
print('data', data)

path = "6001247.zarr"
os.mkdir(path)

write_store = parse_url(path, mode="w").store
root = zarr.group(store=write_store)
write_image(image=data, group=root, axes="tczyx")
```

I haven't looked at any resize issues yet.
I'll try downloading some sample data...

@toloudis do you want to share your code you're using to test this?

toloudis:
The code I am using to test is inside of aicsimageio's feature/zarrwriter branch
https://github.com/AllenCellModeling/aicsimageio/tree/feature/zarrwriter
The key files are `aicsimageio/writers/ome_zarr_writer.py` and `scripts/makezarr.ipynb` which loads some data using aicsimageio into a dask array, and then calls the ome_zarr_writer to save it to zarr.

joshmoore:
cc: @aeisenbarth if there's anything from the https://github.com/aeisenbarth/ngff-writer front

joshmoore:
> ```
> url = "https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.1/6001247.zarr"
> kwargs = {
>     "dimension_separator": "/",
>     "normalize_keys": False,
> }
> ```

@will-moore, I think you're missing data is due to the fact that you're testing with an v0.1 dataset that uses `.` separators. I can get the code working with a freshly converted dataset.

aeisenbarth:
- **`dask.array.to_zarr`**: In my experience, computing pyramid levels with dask and writing them immediately is inefficient (`dask.array.to_zarr(…, compute=True)`). The next pyramid level will not receive the computed array from the previous level, only the delayed object. The next step computes 2 times downscaling, then the next step 3 times etc. Instead, we get the delayed objects for `to_zarr` and let dask [compute them all at once](https://github.com/aeisenbarth/ngff-writer/blob/74bd10f169a8953ac560136785dc02c41db6997a/ngff_writer/writer_utils.py#L311). This gives a single dask graph with several sinks instead of several independent dask graphs.

- **`_by_plane`**: The changes look good to me, also performance-wise. We use a somewhat different approach which is generalized for arbitrary [axis order](https://github.com/aeisenbarth/ngff-writer/blob/74bd10f169a8953ac560136785dc02c41db6997a/ngff_writer/array_utils.py#L103): We move/reshape axes so that we have blocks of to-be-resized axes (spatial 2D/or 3D), then iterate and resize them and restore the desired shape. Using the array protocol the same code works for Numpy and Dask arrays.

will-moore:
Thanks to @joshmoore for pointing out the errors in my script above (wrong `dimension_separator` for "v0.1" data).
I have fixed that, and also added a "re-writing" logic, so that the first time you run the script, it downloads the zarr from remote URL and the 2nd time you run it, it instead reads from the downloaded zarr and writes a copy (to test the speed of writing dask without network delays).

That script is here (it takes ~4 minutes to rewrite the zarr, using this PR):

<details><summary>write_dask_to_omezarr.py</summary>

```
from importlib.resources import path
import zarr
import os
from datetime import datetime

url = "https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.3/idr0095B/11511419.zarr"
target = "11511419.zarr"

# if we've already downloaded, try local re-write (perf not due to network speed)
if os.path.exists(target):
    url = target
    target = "out_" + target
    print("re-writing to", target)
else:
    print("downloading from", url)

reader = parse_url(url, mode="r")
# read full-size resolution
data = reader.load('0')

print('data', data)
print('attrs', reader.root_attrs)

ndims = len(data.shape)
default_axes = "tczyx"[5-ndims:]

axes = reader.root_attrs["multiscales"][0].get("axes", default_axes)
print("axes", axes)

os.mkdir(target)

start_time = datetime.now()
print("start_time", start_time)

# write the image data
write_store = parse_url(target, mode="w").store
root = zarr.group(store=write_store)
# Scaler will downsample dask data in X and Y per plane
write_image(image=data, group=root, axes=axes)

print("done. Duration:", datetime.now() - start_time)
```
</details>

To compare to `ngff-writer`, here's another script that does the same thing, using `ngff-writer`.
It performs the re-write in less that 0.01 seconds (Thanks to @aeisenbarth for explaining the logic of this):

<details><summary>ngff-writer-from-url.py</summary>

```
from ngff_writer.writer import open_ngff_zarr
import os
from datetime import datetime

url = "https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.3/idr0095B/11511419.zarr"
target = "11511419.zarr"

output_dir = "output"
collection_name = "my_images"

target_path = os.path.join(output_dir, collection_name, target)

# if we've already downloaded, try local re-write (perf not due to network speed)
if os.path.exists(target_path):
    url = target_path
    target = "rewrite_" + target
    print("reading from ", target_path)
    print("re-writing to", target)
else:
    print("downloading from", url)

reader = parse_url(url, mode="r")
multiscale = reader.root_attrs["multiscales"][0]
# read full-size resolution
data = reader.load(multiscale["datasets"][0]["path"])
axes = multiscale.get("axes")
print("axes", axes)
axes_names = None
if axes is not None:
    axes_names = axes
print("axes_names", axes_names)

start_time = datetime.now()
print("start_time", start_time)

with open_ngff_zarr(
    store=output_dir,
    dimension_separator="/",
    overwrite=True,
) as f:
    collection = f.add_collection(name=collection_name)

    image = collection.add_image(
        image_name=target,
        dimension_axes = axes_names,
        array=data
    )

print("done. Duration:", datetime.now() - start_time)
```
</details>

It might take a bit of work to adopt the approach that @aeisenbarth uses, but it's certainly what's needed!
@toloudis Do you want to look into that, or I can have a go if you want?

will-moore:
Looking into this, I noticed that `ngff-writer` is using a faster way of down-sampling: by slicing the image (https://github.com/aeisenbarth/ngff-writer/blob/74bd10f169a8953ac560136785dc02c41db6997a/ngff_writer/dask_utils.py#L43)
Using that same function in `ome_zarr.scale.nearest()` to simply `return downscale_nearest(plane, factors=(self.downscale, self.downscale))` fixes the performance issue in this PR (for the small sample image in my scripts above).

will-moore:
However, if I also tell `ngff-writer` to use `resize` instead of slicing (see diff) then it is still very fast.

<details><summary>diff</summary>

```
$ git diff
diff --git a/ngff_writer/array_utils.py b/ngff_writer/array_utils.py
index a07f571..63a796f 100644
--- a/ngff_writer/array_utils.py
+++ b/ngff_writer/array_utils.py
@@ -9,7 +9,7 @@ import zarr
 from dask import array as da
 
 from ngff_writer.constants import DIMENSION_AXES, SPATIAL_DIMENSIONS
-from ngff_writer.dask_utils import downscale_nearest
+from ngff_writer.dask_utils import downscale_nearest, resize
 from ngff_writer.typing import DimensionAxisType
 from ngff_writer.validation import validate_axes_names
 
@@ -74,7 +74,7 @@ def ngff_spatially_rescale(
             min(dim, round(1 / scale)) if name in SPATIAL_DIMENSIONS else 1
             for dim, name in zip(image.shape, axes_names)
         )
-        _resize = lambda image, out_shape, **kwargs: downscale_nearest(image, factors)
+        _resize = lambda image, out_shape, **kwargs: resize(image, out_shape)
         _smooth = dask_image.ndfilters.gaussian_filter
     else:
         _resize = skimage.transform.resize
```
</details>

will-moore:
Another issue with the current writing to disk is that the `dype` seems to get written as `float64` (`"dtype": "<f8"`) even though the dask array dtype was `uint16`. I don't really know what's causing this.

toloudis:
> It might take a bit of work to adopt the approach that @aeisenbarth uses, but it's certainly what's needed! @toloudis Do you want to look into that, or I can have a go if you want?

I am unable to work on this over the next couple of weeks.  I will return to it when I can but please feel free to move forward.

joshmoore:
Hi @toloudis. Big thanks for hammering on with this. If pre-commit becomes unruly, feel free to yell for help. Similarly, not to distract while you and @will-moore are in the thick of it, but we'll probably need to take another look at the final API changes when the dust has settled.

will-moore:
I am certain that the build was green on this PR for the last commit previously (as it is with the same changes at https://github.com/ome/ome-zarr-py/pull/199). Not sure what changed (no new commits)?

will-moore:
After running the test script above, you should be able to view in `napari`:

```
$ napari --plugin napari-ome-zarr 9836998.zarr_download.zarr
```

toloudis:
@will-moore Sorry I didn't do this sooner - I've added you.

I am currently running a long write process (a 250GB zeiss czi file) with the latest version of the codebase to re-test things.  I am also trying to use an explicit dask local cluster to try to get some multithreading to have an impact.

toloudis:
Good news from my testing with large-ish files using this branch in aicsimageio with my very simple OmeZarrWriter.  
```
from distributed import LocalCluster, Client
cluster = LocalCluster(processes=True, memory_limit="7GB")
client = Client(cluster)
# important to not include SpatialZ in chunks!!!!
chunk_dims= [
    DimensionNames.SpatialY,
    DimensionNames.SpatialX,
    DimensionNames.Samples,
]
img = AICSImage(filepath, chunk_dims=chunk_dims)
data = img.get_image_dask_data("TCZYX")
writer = OmeZarrWriter(mypath)

writer.write_image(
    image_data=data,  # must be 5D TCZYX
    image_name="Image_0",
    physical_pixel_sizes=img.physical_pixel_sizes,
    channel_names=img.channel_names,
    scale_num_levels=2,
    scale_factor=4.0,
)
```
This completed in a couple of hours, down from 12+ hours for a ~250GB czi file.  If I had included SpatialZ in the dask chunking for the aicsimageio reader, then memory was overwhelmed, and dask workers failed completely.

toloudis:
Side note related to the Scaler and its usage.  One interesting thing that will come up for us is the ability to choose very specific downsampling rates for pyramid levels.  a simple recursive 2x may not be the most effective for the type of visualization we want to do.  For example, I might want to tile the downsampled z slices across a gpu texture that can be no more than 4096x4096 and so I would compute the biggest xy size that can tile in there without downsampling in Z.   We would also be interested in writing a 2d projection thumbnail image which would be downsampled in xy to a fixed value like 256x256 but only 1 z slice as it would be generated as a projection. I suppose one could argue that it doesn't fit into "multiscales" but it is certainly associated with, and derived from, the same full res image data.

JacksonMaxfield:
@toloudis just curious was this on a SLURM node with high memory or what was the machine like that you were setting up the `LocalCluster` on? Interesting you went with `processes=True`, I would have guessed that threads would be better than processes but I guess the downsample is compute so :shrug: 

Still _very_ good news. In our `OmeZarrWriter` I feel like we should have a parameter in the writer init or something that says "setup_cluster" and try to handle this all for the user but last time we did that it was :-1:

JacksonMaxfield:
Also, I would love to see a screenshot of the dask dashboard when it's writing out all that data :grinning:

JacksonMaxfield:
> I found processes a little easier to monitor using windows task manager, and had changed it when things started crashing but never bothered to switch back to threads.

Gotcha. I would install the `bokeh` so you get the dask dashboard as it has info per worker (threaded or not) but that's only if you want to rerun your multi-hour test :joy:

toloudis:
I'm now running this code stably from a aicsimageio wrapper with dask multiprocessing and direct writes to s3, even. I'd like to graduate it out of draft status now.

joshmoore:
> I'd like to graduate it out of draft status now.

:tada: Thanks, @toloudis!

ome_zarr/writer.py
```diff
@@ -435,17 +437,115 @@ def write_image(
...
+                component=str(Path(group.path, str(path))),
+                storage_options=options,
+                compute=False,
+            )
+        )
+        datasets.append({"path": str(path)})
+
+    da.compute(*delayed)
+
+    if coordinate_transformations is None:
```
sbesson, line 536: Several blocks of this new API are identical to `write_multiscale`. This likely points towards extracting the shared logic into internal functions in order to reduce the maintenance moving forward.

sbesson (commented):
Overall great to see support for `dask` being added. @will-moore and @joshmoore will likely provide independent testing and comments.

The biggest question from my sude is a whether the intent is to generalize the existing APIs  to support more array types than `np.array` or to selectively introduce a set of new public dask writing API that the caller would need to be invoke explicitly when working with dask arrays. At least, in its current form, the PR does not make a clear choice between the two options which I think leads to confusion for consumers.

A few  other questions related to the scope of this work:
-  will other write APIs like `write_labels` need to be updated to support dask arrays?
- `write_multiscale` is currently unmodified, is there a use case for supporting a list of `da.Array`?
- do we foresee other types of arrays e.g. `xarray.DataArray` being supported in the mid-term?

toloudis:
As I continue testing this on big production data,  I'm still having one major dask issue. 

I'm loading data of shape TCXYZ=(241, 2, 150, 1297, 2048) into a dask array with chunk size (1, 1, 1, 1297, 2048) using aicsimageio.  Then passing it to this writer, with 4 levels of downsampling requested and a new "output" chunk size that is different per pyramid level (selected to make the chunks under 16MB).  It spends several minutes initially constructing the big task graph and starts chugging away. Trying to write direct to s3.  Some amount of S3 writes happen (about 100GB worth) before my local workstation's 1TB disk fills up with dask "spill" data in a dask-worker-space directory.   I then get an out of space on device error and the whole thing dies. 
Apart from trial and error changing LocalCluster parameters (which I've done over the past couple of days with no luck) I'm not sure how to approach splitting up this work to get it to finish in "reasonable" time.  Might there be some way to spread out the da.compute calls to do it once per level, or something?

JacksonMaxfield:
> Apart from trial and error changing LocalCluster parameters (which I've done over the past couple of days with no luck) I'm not sure how to approach splitting up this work to get it to finish in "reasonable" time. Might there be some way to spread out the da.compute calls to do it once per level, or something?

You may want to ask in the napari zulip / image.sc someone may have input. Worst case, it may be weird but okay to ask on `dask-image` / `distributed` github issues.

joshmoore:
Anything that can be done to help on this front?

toloudis:
> Anything that can be done to help on this front?

I think the other functions that write data could be updated to allow dask arrays coming in.

I guess my feeling is that this PR should strive not to change the current public API.
I was looking at write_image vs write_multiscales and if they need to both be public, then they both need to have a dask treatment.  

I haven't studied the intended public api THAT carefully, but I was having this thought:
It also seems you could have one single entrypoint like `write_image` that accepts either a single 5D array (with optional Scaler) or a list of 5D arrays (precomputed pyramid levels), and any of the 5D arrays in either flavor could be dask or not.
```
# compute multiscales with scaler
write_image(tczyx, scaler)
# no multiscale, just write this single level
write_image(tczyx, scaler=None)
# if list is passed in, then assume they are precomputed multiscales
write_image([tcxyz0, tczyx1, ...], scaler=None)
write_image([tczyx0, tczyx1, ...], scaler) # error, inconsistent args or else ignore scaler in this case
```

will-moore:
The reason that I've previously avoided supporting Dask arrays for the `write_multiscales` method was that if you don't do the writing and resizing of a pyramid in the same loop, and `compute` everything at the end (as described at https://github.com/ome/ome-zarr-py/pull/192#issuecomment-1103859096) then the downsampling is repeated multiple times by dask and performance suffers quite a bit.

will-moore:
That's a fair point - there are certainly cases where the data may be delayed, but not necessarily suffers the performance hit we saw.

So we should probably have tests for dask in all those write multiscale/image/labels.
I added just the 1 test earlier, but it might be easier to extend coverage by adding a `use_dask` `parametrize` to an existing test(s) with:

```
if use_dask:
    data = da.from_array(data)`
```

aeisenbarth:
In my tests, I use a similar pattern for testing multiple array implementations:
```python
@pytest.mark.parametrize("array_constructor", [np.array, da.from_array, zarr.array])
def test_func(array_constructor):
    ...
    data = array_constructor(data)
```

joshmoore:
Re-launched workflows. @toloudis, anything outstanding from your side?

toloudis:
There is probably some work to be done to consolidate different methods inside of scaler (by_plane vs the dask codepath that does something different), and, possibly related, I still have performance concerns for non-`nearest` scaling with large data...  but I think functionally this is working and all tests passing.

ome_zarr/writer.py
```diff
@@ -302,6 +315,7 @@ def write_multiscales_metadata(
         dict(
             version=fmt.version,
             datasets=_validate_datasets(datasets, ndim, fmt),
+            name=name if name else group.name,
```
toloudis, line 318: this line might be too heavyhanded, as the spec says that name "SHOULD" be provided but not "MUST".  Is it ok for this code to write out group.name?  I believe there were tests that depended on name having been written.

toloudis:
The thing I said about refining the Scaler is probably good for a separate PR and not this one.  Merging this PR should not be disruptive to the non-dask workflow, and only enables dask arrays to be passed in.

jakirkham:
Is there anything else needed here before merging? Sounds like no, but wanted to check 🙂

ome_zarr/writer.py
```diff
@@ -634,18 +764,38 @@ def write_labels(
...
+            fmt=fmt,
+            axes=axes,
+            coordinate_transformations=coordinate_transformations,
+            storage_options=storage_options,
+            name=name,
+            **metadata,
+        )
+    else:
+        mip, axes = _create_mip(labels, fmt, scaler, axes)
+        write_multiscale(
```
sbesson, line 783: Does that effectively mean that `write_multiscales_label` should be deprecated or only that it applies to a subset of cases (and does not support dask)?

sbesson (changes requested):
A few changes requested inline but nothing major.

Coming back to the API questions raised in https://github.com/ome/ome-zarr-py/pull/192#pullrequestreview-987213535, I think the following commits addressed most of my questions but the resolution state is a bit unclear. Would it possible to have a short summary of the list of APIs which have been updated to support `dask` either as a comment or as an update of the PR description? This can also be used directly in the release notes.

A few additional question regarding integration & release:

- what is the status in terms of functional testing with the latest state of this proposed change? Note https://github.com/ome/ome-zarr-py/issues/215 might also provide a potential use case for the functionality
- are there any follow-up todos planned on top of this work before they can be included in an upcoming `0.x.0` release?
- as this is probably the biggest ongoing piece of work on this library, are we happy to merge once all outstanding comments have been cleared and possibly release as a alpha/beta/release candidate to gain some experience before making a final release?

ome_zarr/writer.py
toloudis: `write_multiscale_labels` was calling `write_multiscale` before this PR too.   So basically I'm just trying to preserve the pre-dask code path.

ome_zarr/scale.py
```diff
@@ -125,6 +130,30 @@ def __create_group(
             series.append({"path": path})
         return grp
 
+    def resize_image(self, image: ArrayLike) -> ArrayLike:
```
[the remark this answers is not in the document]
will-moore: I guess that none of these issues is a blocker - Improving the `Scaler` documentation and usage - Using the `scaler.resize_image()` instead of `scaler.nearest()` for dask AND non-dask data, possibly removing other methods from the class etc could come in a follow-up PR (even a follow-up release). cc @joshmoore?

Probably the most important fix should be handling `write_image(dask_data, group, scaler=None)` which I think would currently fail.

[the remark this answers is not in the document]
toloudis: I would agree that dask versions of gaussian, laplacian, etc could be done in later PRs.
I can try to do one more pass this afternoon to make sure `scaler=None` does not break with dask array. 
As I'm sure is true with many others, I am splitting my time among many different projects, so thanks for living with these intermittent pieces of work on this PR.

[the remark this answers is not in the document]
toloudis: Had a quick look and I am already handling `scaler=None` in `_write_dask_image`: (1) max_layer set to 0 if scaler is None, and (2) don't call scaler.resize_image if scaler is None

toloudis:
I'm wondering what the status is on this PR.
If this is still on the good path to being merged, what work is remaining?  Maybe time for a re-review?
(I'm eager to get an aicsimageio writer published using this code.)
Also-- if it would make life easier for additional contributions, I'm happy to just have all the changes moved to an internal branch or someone else's fork. I'm not sure what the thoughts are on merging and then doing additional refinements in followup PRs.

sbesson:
@toloudis thanks for re-upping the priority of this now that everyone is more or less back from holidays. The questions you bring up pretty much echo what I said in https://github.com/ome/ome-zarr-py/pull/192#pullrequestreview-1052440568 so I believe we are all on the same page although a bit looking at each other :)

As a preamble, I am transitioning to a new role where I may or may not remain actively involved in the code base. In this context, it feels sensible to let you guys drive the next set of additions to the library and I'll make sure that Will and the rest of the OME team have the necessary permissions to do so.

My personal opinion is that the dask support is demanded feature, this is a large contribution so there is interest in prioritizing its integration and planning its officual release. Both you and @will-moore have the best knowledge of the state of things. Assuming you are both happy with the current state, my inclination would be to get this merged in `master` and  immediately cut a pre-release so that various use cases can start consuming it and providing feedback and/or contributions.

This leaves the wider question of the roadmap/timeline towards a final release with dask support. If that's something that cannot be decided at this time, maybe it is worth creating a `0.6.0` milestone so that we can capture issues from this PR as well as future feedback as issues.

will-moore:
Just tested this again using the sample at https://github.com/ome/ome-zarr-py/issues/219 and this seems to be working.

We'll need some follow-up additions to the documentation, but this is looking good for a pre-release as @sbesson suggested.

jakirkham:
Thanks all! It is great to see this in 😄

ome_zarr/scale.py
```diff
@@ -19,10 +20,14 @@
     resize,
 )
 
+from .dask_utils import resize as dask_resize
 from .io import parse_url
 
 LOGGER = logging.getLogger("ome_zarr.scale")
 
+ListOfArrayLike = Union[List[da.Array], List[np.ndarray]]
+ArrayLike = Union[da.Array, np.ndarray]
```
jakirkham, line 29: Perhaps in the future these could be [some kind of general Array]( https://data-apis.org/array-api/latest/design_topics/static_typing.html ) (once that is more fleshed out).

PR #195, opened 2022-05-04
Plate pyramids and version

will-moore:
This fixes one bug with viewing older (v0.1) plates and resolves the long-standing limitation of only loading a single resolution level for plates. Now we load a dask pyramid.

- Bug-fix: When `Format` class checks for version, it only read the `multiscales` metadata and not the 'plate', so we defaulted to using the `CurrentFormat` for reading plates. This meant that the `dimension_separator` was `/` and so for `v0.1` plates which use `.` the data wasn't loaded. Also, the `matches()` check was duplicated in the `FormatV02` subclass that had different logging than the `FormatV01` subclass so that logging for a `plate` previously looked like this:

```
14:55:02 DEBUG 0.4 matches None?
14:55:02 DEBUG 0.3 matches None?
14:55:02 DEBUG 0.2 matches None?
14:55:02 DEBUG V01:None v. 0.1
```
This now looks like:
```
14:57:00 DEBUG 0.4 matches 0.1?
14:57:00 DEBUG 0.3 matches 0.1?
14:57:00 DEBUG 0.2 matches 0.1?
14:57:00 DEBUG 0.1 matches 0.1?
```

 - Plate pyramids, see https://github.com/ome/napari-ome-zarr/issues/45 That discussion prompted another go at loading a dask pyramid (instead of picking a single resolution). So now, you can zoom in to a region of a plate in `napari` and it loads higher resolution tiles!

To test:
This `v0.1` plate previously appeared as black in `napari`, but now it loads and when you zoom in you get higher resolution tiles (although they seem to be very aggressively loaded so it is quite slow!):

```
$ napari --plugin napari-ome-zarr https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.1/plates/422.zarr -vvv
```

Screenshot 2022-05-04 at 14 34 17

### DIFF
--- a/ome_zarr/format.py
+++ b/ome_zarr/format.py
@@ -64,14 +64,21 @@ class Format(ABC):
     def init_channels(self) -> None:  # pragma: no cover
         raise NotImplementedError()
 
-    def _get_multiscale_version(self, metadata: dict) -> Optional[str]:
+    def _get_metadata_version(self, metadata: dict) -> Optional[str]:
+        """
+        Checks the metadata dict for a version
+
+        Returns the version of the first object found in the metadata,
+        checking for 'multiscales', 'plate', 'well' etc
+        """
         multiscales = metadata.get("multiscales", [])
         if multiscales:
             dataset = multiscales[0]
             return dataset.get("version", None)
-        plate = metadata.get("plate", [])
-        if plate:
-            return plate.get("version", None)
+        for name in ["plate", "well", "image-label"]:
+            obj = metadata.get(name, None)
+            if obj:
+                return obj.get("version", None)
         return None
 
     def __repr__(self) -> str:
@@ -120,7 +127,7 @@ class FormatV01(Format):
         return "0.1"
 
     def matches(self, metadata: dict) -> bool:
-        version = self._get_multiscale_version(metadata)
+        version = self._get_metadata_version(metadata)
         LOGGER.debug(f"{self.version} matches {version}?")
         return version == self.version
 
--- a/ome_zarr/format.py
+++ b/ome_zarr/format.py
@@ -121,7 +121,7 @@ class FormatV01(Format):
 
     def matches(self, metadata: dict) -> bool:
         version = self._get_multiscale_version(metadata)
-        LOGGER.debug(f"V01:{version} v. {self.version}")
+        LOGGER.debug(f"{self.version} matches {version}?")
         return version == self.version
 
     def init_store(self, path: str, mode: str = "r") -> FSStore:
@@ -167,11 +167,6 @@ class FormatV02(FormatV01):
     @property
     def version(self) -> str:
         return "0.2"
-
-    def matches(self, metadata: dict) -> bool:
-        version = self._get_multiscale_version(metadata)
-        LOGGER.debug(f"{self.version} matches {version}?")
-        return version == self.version
 
     def init_store(self, path: str, mode: str = "r") -> FSStore:
         """
--- a/ome_zarr/format.py
+++ b/ome_zarr/format.py
@@ -69,6 +69,9 @@ class Format(ABC):
         if multiscales:
             dataset = multiscales[0]
             return dataset.get("version", None)
+        plate = metadata.get("plate", [])
+        if plate:
+            return plate.get("version", None)
         return None
 
     def __repr__(self) -> str:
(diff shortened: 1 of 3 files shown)


ome_zarr/format.py
```diff
@@ -69,6 +69,9 @@ def _get_multiscale_version(self, metadata: dict) -> Optional[str]:
         if multiscales:
             dataset = multiscales[0]
             return dataset.get("version", None)
+        plate = metadata.get("plate", [])
```
sbesson, line 72: Should the method be renamed to match the fact that it detects the version outside the scope of the `multiscales` specification or do you see a driver for keeping separate `_get_<spec>_version`?

Also will this method need to be extended to support every intermediate hierarchy level that would define a version, for instance `well` ?

will-moore: Good catch! Fixed in [d1b93bc](https://github.com/ome/ome-zarr-py/pull/195/commits/d1b93bc56cdc679073873e696da5fb45e3569c6f)
This now allows viewing a Well (which didn't work before) e.g.
```
$ napari --plugin napari-ome-zarr https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.1/plates/422.zarr/A/1 -vvv
```

joshmoore:
:+1: from my side on the code changes.

sbesson (approved):
Code changes look generally sensible and remove a  lot of logic (which I assume is now handled downstream).

Tested using the latest `napari 0.4.15` with https://github.com/ome/napari-ome-zarr/pull/42 included in the `napari-ome-zarr` plugin - see https://github.com/ome/napari-ome-zarr/pull/42#issuecomment-1134551267 with different plates/well, and two OME-NGFF plates (using the 0.1 and 0.4 versions of the specification)

Without this PR i.e. with

```
(napari) sbesson@Sebastiens-MacBook-Pro 2016-06 % pip freeze | grep zarr
napari-ome-zarr @ git+https://github.com/ome/napari-ome-zarr@f572c602da60646210f64583f407c9ac9e44bd00
ome-zarr==0.4.1
zarr==2.11.3
```

- `napari https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.1/plates/2551.zarr` displays black thumbnails
- `napari https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.1/plates/2551.zarr/A/1`  fails with `AttributeError: 'NoneType' object has no attribute 'plugin'`
- `napari https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0072B/9512.zarr` renders as expected but all tiles are low-resolution
  Screenshot 2022-05-23 at 12 33 29
- `napari https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0072B/9512.zarr/L/20 ` fails with `AttributeError: 'NoneType' object has no attribute 'plugin'`

With this PR i.e.
```
(napari) sbesson@Sebastiens-MacBook-Pro 2016-06 % pip freeze | grep zarr
napari-ome-zarr @ git+https://github.com/ome/

napari-ome-zarr@f572c602da60646210f64583f407c9ac9e44bd00
ome-zarr @ git+https://github.com/ome/ome-zarr-py@83ee1c1437b61ff5daeb0ae0095fd196a632fd12
zarr==2.11.3
```

- `napari https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.1/plates/2551.zarr`  now renders as expected (although slowly) and loads high-resolution tiles as you zoom in
 Screenshot 2022-05-23 at 12 36 54
 Screenshot 2022-05-23 at 12 37 24
 - `napari https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.1/plates/2551.zarr/A/1` still fails with `AttributeError: 'NoneType' object has no attribute 'plugin'`
 - `napari https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0072B/9512.zarr` renders as expected with higher resolution tiles loaded as the viewer zooms in

Screenshot 2022-05-23 at 12 40 04

- `napari https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0072B/9512.zarr/L/20 ` still fails with `AttributeError: 'NoneType' object has no attribute 'plugin'`

Since the well loading was not really part of the original contract, I'm inclined to get this in and released (alongside the napari-ome-zarr npe2 PR) to fix the behavior of napari with plates

sbesson:
After some investigation, it looks like the well-level access issue is related to the downstream `napari-ome-zarr` plugin - see https://github.com/ome/napari-ome-zarr/pull/42#discussion_r879374496

PR #197, opened 2022-05-11
bugfix in chunks when providing storage_options as dict

claesenm:
The current implementation breaks when providing a single dict for `storage_options`, because `options.pop("chunks",...)` on line 238 removes the key `chunks` from the original `storage_options` dict after the first iteration. This happens because `options` is an alias of `storage_options` instead of a deep copy. My change fixes this and provides the intended behavior (all pyramid levels follow the same chunking scheme if `storage_options` is provided as a dict.

### DIFF
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -228,7 +228,7 @@ Please use the 'storage_options' argument instead."""
         options = {}
         if storage_options:
             options = (
-                storage_options
+                storage_options.copy()
                 if not isinstance(storage_options, list)
                 else storage_options[path]
             )


sbesson (changes requested):
@claesenm thanks for the contribution. The rationale is clear and the proposed changes make complete sense as the `storage_options` object can be modified internally.

Looking at tests/test_writer.py, we have no coverage of calling `write_multiscale` with storage options specified either as simple dictionary or a list of dictionaries. Would it be possible to add unit tests matching the scenario addressed in this PR to prevent future regressions?

claesenm:
Will look into adding unit tests ASAP. Hopefully next week.

will-moore:
I recently added some testing of chunks in https://github.com/ome/ome-zarr-py/pull/192/commits/61a7440820a30adb900b4f3bf88f0a8a20be9a8c
Although it only checks the chunks for the first resolution of the pyramid.
I just updated my tests locally to test further pyramid levels, and indeed this fails without your fix:

```
diff --git a/tests/test_writer.py b/tests/test_writer.py
index 83ee40f..529ab6f 100644
--- a/tests/test_writer.py
+++ b/tests/test_writer.py
@@ -110,8 +110,10 @@ class TestWriter:
             ):
                 assert transf == expected
             assert len(node.metadata["coordinateTransformations"]) == len(node.data)
-        first_chunk = [c[0] for c in node.data[0].chunks]
-        assert tuple(first_chunk) == _retuple(chunks, node.data[0].shape)
+        # check chunks for first 2 resolutions (before shape gets smaller than chunk)
+        for nd_array in node.data[:2]:
+            first_chunk = [c[0] for c in nd_array.chunks]
+            assert tuple(first_chunk) == _retuple(chunks, nd_array.shape)
         assert np.allclose(data, node.data[0][...].compute())
```

You could take that and add it to this PR (then I can resolve merge conflicts later)

PR #202, opened 2022-05-24
Fix the Zarr development GH workflow following branch renaming

sbesson:
The upstream master branch has been merged into main and removed - see https://github.com/zarr-developers/zarr-python/pull/1033


sbesson:
So would it be better to completely remove `@<branch>` from your perspective i.e. should the default branch (as opposed to a named branch) be the one this workflow runs against?

joshmoore:
That was the action taken on (some of the) zarr repos. I don't have a strong feeling. :+1: either way.

PR #206, opened 2022-06-20
Write version attribute to image-label specification

sbesson:
Discovered during the testing of https://github.com/ome/omero-cli-zarr/pull/107

The `write*label*` APIs introduced in #178 do not store the OME-NGFF specification version in the `image-label` attributes.

The changes proposed in this PR

- add a `fmt` optional parameter to `write_label_metadata`
- add the logic to store the format version in the `image-label` attributes
- updates the label writing tests to test the version metadata

I would be inclined this as a bug fix and would suggest to get it out as a patch release `0.4.3`. But strictly speaking, this adds an optional parameter to a public API and could be released as a minor version increment.

### DIFF
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -453,6 +453,7 @@ def write_label_metadata(
     name: str,
     colors: List[JSONDict] = None,
     properties: List[JSONDict] = None,
+    fmt: Format = CurrentFormat(),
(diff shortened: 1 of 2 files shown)


will-moore (approved):
Code looks good, and using the comparison approach @sbesson described at https://github.com/ome/omero-cli-zarr/pull/107, this fixes the loss of version number caused by the change in that PR. 👍

PR #212, opened 2022-07-09
Disable auto_mkdir when path begins with `s3`

colobas:
When trying to use `ome-zarr-py` I was getting an error thrown that complained about `aiobotocore.session.AioSession`'s `__init__` not having an `auto_mkdir` parameter, which I believe was being incorrectly passed to it. This PR fixes that by adding a condition to exclude `auto_mkdir` from `kwargs` when the path starts with `s3`, similarly to what is already done when the path starts with `http`.

### DIFF
--- a/ome_zarr/format.py
+++ b/ome_zarr/format.py
@@ -187,7 +187,7 @@ class FormatV02(FormatV01):
         }
 
         mkdir = True
-        if "r" in mode or path.startswith("http"):
+        if "r" in mode or path.startswith("http") or path.startswith("s3"):
             # Could be simplified on the fsspec side
             mkdir = False
         if mkdir:


toloudis:
I had the same issue, and did an identical change on my local branch during my work with dask trying to write directly to s3.

sbesson (approved):
The change makes complete sense and treats the S3 protocol in the same way as HTTPS. I did not find anything in th Python libraries in the likes of `is_remote_protocol` that could be used here. Possibly some upstream libraries (like `fsspec`) have similar utilities. Immediately, I think there is no harm in getting this out as a patch release.

joshmoore:
cc: @martindurant in case there's something we're missing. Thanks all.

martindurant:
I believe it's fixed in s3fs main, but fine to fix it here too.

PR #214, opened 2022-07-22
Use multiscales name for node metadata['name']

will-moore:
Fixes https://github.com/ome/napari-ome-zarr/issues/36

This uses the `multiscales` "name" field to populate the `node.metadata["name"]` which is then used by `napari-ome-zarr` for the layer names.

To test, using the data provided on the issue above (https://oc.embl.de/index.php/s/yfwGLkFRHsopcug) and opened in `napari`, the name ("image") is used for each layer.

Screenshot 2022-07-22 at 15 04 25

### DIFF
--- a/ome_zarr/reader.py
+++ b/ome_zarr/reader.py
@@ -288,6 +288,8 @@ class Multiscales(Spec):
             # Raises ValueError if not valid
             axes_obj = Axes(axes, fmt)
             node.metadata["axes"] = axes_obj.to_list()
+            # This will get overwritten by 'omero' metadata if present
+            node.metadata["name"] = multiscales[0].get("name")
             paths = [d["path"] for d in datasets]
             self.datasets: List[str] = paths
             transformations = [d.get("coordinateTransformations") for d in datasets]


sbesson (approved):
As the `name` field already exists and is populated in the `OMERO` spec, it makes sense to also populate it in `Multiscales`. The overriding logic should maintain the current expectation if a node implements `Multiscales` and `OMERO`

Echoing some of the wonders in https://github.com/ome/ome-zarr-py/pull/140, this raises the question of whether we could define the expectations for the `metadata` field of each node. For instance, is there a reference list of the fields that should/must be populated?

ome_zarr/reader.py
```diff
@@ -288,6 +288,8 @@ def __init__(self, node: Node) -> None:
             # Raises ValueError if not valid
             axes_obj = Axes(axes, fmt)
             node.metadata["axes"] = axes_obj.to_list()
+            # This will get overwritten by 'omero' metadata if present
+            node.metadata["name"] = multiscales[0].get("name")
```
joshmoore, line 292: and/or OME-XML metadata down the road... :+1:

joshmoore:
Think it's safe to merge this and we will need to come back to all the open questions as we move forward.

Issue #219, opened 2022-08-22
Error when writing to disk a dask-backed xarray


LucaMarconato:
The following code example produces an error on the last line. 

The code does the following:
1) creates a 3D array (cyx axes) and writes it to zarr
2) loads it from zarr
3) writes to another file what has been loaded

```
##
import os.path
import numpy as np
import zarr
import shutil
from ome_zarr.writer import write_image
from ome_zarr.io import parse_url
from ome_zarr.io import ZarrLocation
from ome_zarr.reader import Multiscales, Reader
from ome_zarr.format import CurrentFormat
import dask.array.core

im = np.random.normal(size=(3, 100, 100))
fmt = CurrentFormat()

## write
def write_to_zarr(im, f):
    if os.path.isdir(f):
        shutil.rmtree(f)
    store = parse_url(f, mode="w").store
    group = zarr.group(store=store)
    write_image(im, group, axes=["c", "x", "y"], fmt=fmt)

write_to_zarr(im, "debug0.zarr")

## read
loc = ZarrLocation("debug0.zarr")
reader = Reader(loc)()
nodes = list(reader)
assert len(nodes) == 1
node = nodes[0]
im_read = node.load(Multiscales).array(resolution="0", version=fmt.version)

## write again (error)
write_to_zarr(im_read, "debug1.zarr")

##
import xarray as xr
xr.DataArray(im).chunks

```

The error is the following:
```
Traceback (most recent call last):
  File "/Users/macbook/miniconda3/envs/ome/lib/python3.9/site-packages/IPython/core/interactiveshell.py", line 3335, in run_ast_nodes
    code = compiler(mod, cell_name, mode)
  File "/Users/macbook/miniconda3/envs/ome/lib/python3.9/codeop.py", line 143, in __call__
    codeob = compile(source, filename, symbol, self.flags, True)
  File "<ipython-input-3-61b5ce34eaa1>", line 1
SyntaxError: 'return' outside function
Traceback (most recent call last):
  File "/Users/macbook/miniconda3/envs/ome/lib/python3.9/site-packages/zarr/hierarchy.py", line 800, in _write_op
    return f(*args, **kwargs)
  File "/Users/macbook/miniconda3/envs/ome/lib/python3.9/site-packages/zarr/hierarchy.py", line 965, in _create_dataset_nosync
    a = array(data, store=self._store, path=path, chunk_store=self._chunk_store,
[...]

PR #220, opened 2022-08-31
Use integer division for new_shape in resize_image

will-moore:
Fixes issue reported at https://github.com/ome/ome-zarr-py/issues/219#issuecomment-1231613397

The issue there was that the dask writing used `resize_image()` which calculated down-sizing using `ceil()` so the sizes of pyramids levels were `100, 50, 25, 13, 7` instead of `100, 50, 25, 12, 6` as for the non-dask data.

I updated `resize_image()` to use the same integer division for calculating new_shape as is used in the `def __nearest()` for non-dask data.

To test:
 - run the sample script on issue above (using ome-zarr-py installed from this branch)
 - check that the pyramid sizes are `100, 50, 25, 12, 6` for both the `debug0.zarr` (non-dask) and `debug1.zarr` (dask).
 - The `coordinateTransforms` should also be the same for both

### DIFF
--- a/ome_zarr/scale.py
+++ b/ome_zarr/scale.py
@@ -144,8 +144,8 @@ class Scaler:
 
         # only down-sample in X and Y dimensions for now...
         new_shape = list(image.shape)
-        new_shape[-1] = np.ceil(float(image.shape[-1]) / self.downscale)
-        new_shape[-2] = np.ceil(float(image.shape[-2]) / self.downscale)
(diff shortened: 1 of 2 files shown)


joshmoore:
The change & explanation make sense, @will-moore. I'd strongly suggest though that we get the test migrated over from the issue.

will-moore:
@joshmoore Added more tests, and with #221 they're green!

joshmoore:
Thanks, @will-moore.

PR #223, opened 2022-09-07
Dask docs update

will-moore:
Update docs to mention dask support.

Also add a code sample for reading data, since this is hard to figure out from the api docs alone.


joshmoore:
https://ome-zarr--223.org.readthedocs.build/en/223/python.html#reading-ome-ngff-images looks good with the latest commit. Waiting for the rest of the build.

PR #224, opened 2022-09-07
Add changelog 0.6.0

will-moore:
Add changelog 0.6.0.

Do we call this `0.6.0rc1`, or should we just go for `0.6.0` release?


sbesson:
> Do we call this 0.6.0rc1, or should we just go for 0.6.0 release?

No strong preference either way from my side. I just feel we need some form of release of PyPI that people can start consuming. Is https://github.com/ome/ome-zarr-py/pull/220 part of the roadmap?

will-moore:
#220 is an update to the dask PR so that dask-downsampling uses the same shape sizes as non-dask downsampling.
I didn't list it as it's not a fix to released code, so users won't be affected by it - it's really a tweak to the dask PR (considered part of the same new feature).

joshmoore (approved):
LGTM

pwalczysko (approved):

will-moore:
OK - I'll go direct for 0.6.0 release - thanks!

PR #235, opened 2022-11-11
Don't hard-code dimension_separator when reading

will-moore:
Fixes https://github.com/ome/napari-ome-zarr/issues/66

Allows reading of OME-NGFF that use `dimension_separator: "."`,
corresponding to the proposal to allow `.` as dimension_separator: https://github.com/ome/ngff/pull/161

When writing, the behaviour is unchanged and we still use `/`.

### DIFF
--- a/ome_zarr/format.py
+++ b/ome_zarr/format.py
@@ -183,10 +183,14 @@ class FormatV02(FormatV01):
         TODO: could also check dimension_separator
         """
 
-        kwargs = {
-            "dimension_separator": "/",
+        kwargs: Dict[str, Any] = {
             "normalize_keys": False,
         }
+
+        # if creating a store, use / by default. If reading, use the
+        # dimension_separator specified in the .zarray
+        if mode == "w":
+            kwargs["dimension_separator"] = "/"
 
         mkdir = True
         if "r" in mode or path.startswith("http") or path.startswith("s3"):
--- a/tests/data/v2/0/.zarray
+++ b/tests/data/v2/0/.zarray
@@ -13,6 +13,7 @@
         "id": "blosc",
         "shuffle": 1
     },
+    "dimension_separator": "/",
     "dtype": "|u1",
     "fill_value": 0,
     "filters": null,


will-moore:
This PR would fix the usage reported at #245, where dimension separator `.` is being used (unintentionally).

will-moore:
I've temporarily merged in #243 above to fix the tests. All passing now except the one Windows build that is failing on that PR.

ome_zarr/format.py
```diff
@@ -183,11 +183,15 @@ def init_store(self, path: str, mode: str = "r") -> FSStore:
...
 
-        kwargs = {
-            "dimension_separator": "/",
+        kwargs: Dict[str, Any] = {
             "normalize_keys": False,
         }
 
+        # if creating a store, use / by default. If reading, use the
+        # dimension_separator specified in the .zarray
+        if mode == "w":
```
joshmoore, line 192: There certainly _were_ datasets that had no `dimension_separator` in the .zarray and therefore this would make them unreadable. I'd suggest we do this with the bump of the spec that says we no longer support data written before dimension_separator was introduced.

will-moore: If datasets have no `dimension_separator` then they will be using `.` which is the zarr default, so that will be used and we will be able to read them.
Whereas, without this PR they won't be readable.
However, the spec for v0.2 - 0.4 says they should use `/`, so this PR is about adding support for "invalid" data, but we have seen plenty of examples where that exists.

joshmoore: > However, the spec for v0.2 - 0.4 says they should use /, so this PR is about adding support for "invalid" data, but we have seen plenty of examples where that exists.

Understood but there are (at least) "two" types of invalid data.

will-moore: Ah - OK, I understand the possible existence of e.g. v0.2 data that uses `/` dimension_separator but doesn't specify the `dimension_separator` in the `.zarray`.
But I'm not so sure about how the next release of NGFF spec (v0.5) should affect the support for `ome-zarr-py` to read that data, since the spec doesn't specify how libraries should support old versions of data?

joshmoore: The plan I've been working towards is dropping any requirement on the dimension separator in 0.5 other than to say the Zarr implementation used is responsible for _either_ using the default of `.` or of setting `dimension_separator` correctly. i.e. not our business. But the current specs <0.5 explicitly state the dimension separator to use and therefore I would suggest not changing those semantics.

cc: @normanrz @kevinyamauchi

PR #256, opened 2023-03-03
Read tables using anndata._io.read_zarr for local data

will-moore:
This creates a Node for each table listed under `image.zarr/tables/.zattrs` and adds a `node.table` AnnData object to the node. The `node.metadata` dict is populated with the `image.zarr/tables/my_table/.zattrs` data.

These tables can be used by e.g. `napari-ome-zarr` as needed to read "regions" or "tracking" data etc.

We read remote anndata using `anndata.experimental` methods so the anndata requirement is pinned to an exact version.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,5 +1,5 @@
 [settings]
-known_third_party = dask,numcodecs,numpy,pytest,scipy,setuptools,skimage,zarr
+known_third_party = anndata,dask,numcodecs,numpy,pytest,scipy,setuptools,skimage,zarr
 multi_line_output = 3
 include_trailing_comma = True
 force_grid_wrap = 0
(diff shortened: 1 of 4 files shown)


will-moore:
@joshmoore/@sbesson If this approach looks good to you, I'll start to add tests?
The only bit of metadata that isn't easily accessible by API is the "name" of the table, e.g. `image.zarr/tables/table_name/` which is available at the end of the node's `path`. I could add to this to the `node.metadata`, which would "pollute" the anndata metadata: e.g.

```
{
    "table-name": "table_name",
    "encoding-type": "anndata",
    "encoding-version": "0.1.0",
    "type": "ngff:points_table"
}
```
but I'll probably hold off for now, since it's not needed, and can be got from path.

will-moore:
Closing since tables spec is not changing just now...

PR #261, opened 2023-03-14
add omero metadata to writer

### LINKED ISSUE #260: writer for omero metadata
following discussion: https://github.com/ome/ngff/issues/155

I passed the omero metadata naively to `write_multiscale_metadata` but noticed that they are put in the `multiscale` level instead of `group` level. In the test dataset such metadata is simply assigned to the group-level metadata, without validation
https://github.com/giovp/ome-zarr-py/blob/8f265fb931f9ce9070943659b15bf14f3d9b5758/ome_zarr/data.py#L170

I think having a way to pass the omero metadata to write (and hence to validate before writing) would be useful. A minimal solution for this could be the following

```python
# pass omero metadata to multiscale
# to that this function is called

write_multiscales_metadata(..., **metadata)
```
and such function could simply first validate (e.g. can OMERO spec be instantiated?) and then write

```python
	....
    if "omero" in metadata:
		omero_metadata = metadata.get("omero")
		try:
				_ = Omero(omero_metadata)
				group.attrs["omero"] = omero_metadata
		except:
			raise KeyError(...)
    ...
    group.attrs["multiscales"] = multiscales
```

What do you think?

giovp:
close #260 .

Validation before write is minimal, feedbacks welcome.

### DIFF
--- a/ome_zarr/reader.py
+++ b/ome_zarr/reader.py
@@ -383,8 +383,7 @@ class OMERO(Spec):
                     elif contrast_limits is not None:
                         contrast_limits[idx] = [start, end]
 
-            # node.metadata["channel_names"] = names
-            node.metadata["names"] = names
+            node.metadata["channel_names"] = names
             node.metadata["visible"] = visibles
             node.metadata["contrast_limits"] = contrast_limits
             node.metadata["colormap"] = colormaps
--- a/ome_zarr/reader.py
+++ b/ome_zarr/reader.py
@@ -383,7 +383,8 @@ class OMERO(Spec):
                     elif contrast_limits is not None:
                         contrast_limits[idx] = [start, end]
 
-            node.metadata["name"] = names
+            # node.metadata["channel_names"] = names
+            node.metadata["names"] = names
             node.metadata["visible"] = visibles
             node.metadata["contrast_limits"] = contrast_limits
             node.metadata["colormap"] = colormaps
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -4,7 +4,7 @@
 import logging
 import warnings
 from pathlib import Path
-from typing import TYPE_CHECKING, Any, Dict, List, Optional, Tuple, Union
+from typing import Any, Dict, List, Optional, Tuple, Union
 
 import dask.array as da
 import numpy as np
@@ -311,9 +311,7 @@ def write_multiscales_metadata(
             if axes is not None:
                 ndim = len(axes)
 
-    if isinstance(metadata, dict) and "omero" in metadata["metadata"]:
-        if TYPE_CHECKING:
-            assert isinstance(metadata["metadata"], dict)
+    if isinstance(metadata["metadata"], dict) and "omero" in metadata["metadata"]:
         omero_metadata = metadata["metadata"].get("omero")
         if omero_metadata is None:
             raise KeyError("If `'omero'` is present, value cannot be `None`.")
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -318,8 +318,17 @@ def write_multiscales_metadata(
         for c in omero_metadata["channels"]:
             if "color" not in c:
                 raise KeyError("`'color'` not found in `'channels'`.")
+            if not isinstance(c["color"], str):
+                raise TypeError("`'color'` must be a string.")
             if "window" not in c:
                 raise KeyError("`'window'` not found in `'channels'`.")
+            if not isinstance(c["window"], dict):
+                raise TypeError("`'window'` must be a dict.")
+            for p in ["min", "max", "start", "end"]:
+                if p not in c["window"]:
+                    raise KeyError(f"`'{p}'` not found in `'window'`.")
+                if not isinstance(c["window"][p], int):
+                    raise TypeError(f"`'{p}'` must be an int.")
             group.attrs["omero"] = omero_metadata
 
     # note: we construct the multiscale metadata via dict(), rather than {}
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -332,19 +332,17 @@ def write_multiscales_metadata(
         if omero_metadata is None:
             raise KeyError("If `'omero'` is present, value cannot be `None`.")
         for c in omero_metadata["channels"]:
-            if "color" not in c:
-                raise KeyError("`'color'` not found in `'channels'`.")
-            if not isinstance(c["color"], str):
-                raise TypeError("`'color'` must be a string.")
-            if "window" not in c:
-                raise KeyError("`'window'` not found in `'channels'`.")
-            if not isinstance(c["window"], dict):
-                raise TypeError("`'window'` must be a dict.")
-            for p in ["min", "max", "start", "end"]:
-                if p not in c["window"]:
-                    raise KeyError(f"`'{p}'` not found in `'window'`.")
-                if not isinstance(c["window"][p], int):
-                    raise TypeError(f"`'{p}'` must be an int.")
+            if "color" in c:
+                if not isinstance(c["color"], str) and len(c["color"]) != 6:
+                    raise TypeError("`'color'` must be a hex code string.")
+            if "window" in c:
+                if not isinstance(c["window"], dict):
+                    raise TypeError("`'window'` must be a dict.")
+                for p in ["min", "max", "start", "end"]:
+                    if p not in c["window"]:
+                        raise KeyError(f"`'{p}'` not found in `'window'`.")
+                    if not isinstance(c["window"][p], int):
+                        raise TypeError(f"`'{p}'` must be an int.")
             group.attrs["omero"] = omero_metadata
 
     # note: we construct the multiscale metadata via dict(), rather than {}
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -321,7 +321,6 @@ def write_multiscales_metadata(
             axes = _get_valid_axes(axes=axes, fmt=fmt)
             if axes is not None:
                 ndim = len(axes)
-    print(metadata)
     if (
         isinstance(metadata, dict)
         and metadata.get("metadata")
@@ -341,7 +340,7 @@ def write_multiscales_metadata(
                 for p in ["min", "max", "start", "end"]:
                     if p not in c["window"]:
                         raise KeyError(f"`'{p}'` not found in `'window'`.")
-                    if not isinstance(c["window"][p], int):
+                    if not isinstance(c["window"][p], (int, float)):
                         raise TypeError(f"`'{p}'` must be an int.")
             group.attrs["omero"] = omero_metadata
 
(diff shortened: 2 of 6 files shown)


will-moore:
Thanks for the PR. I wonder if you would also want to extend the validation a bit, although some of this would be duplicating the [schema validation](https://github.com/ome/ngff/blob/15565cb8301e14fc9ba2fc63b6332a97534cee67/0.4/schemas/image.schema#L61) e.g:

 - check that the `window` has min, max, start, end numbers
 - check that the `color` is a string

You could even check that the number of `omero` 'channels' matches the number of channels in the data, although this is a bit more work since you have to handle different axes for different versions to know which is channel dimension. Could be left for now.

will-moore:
Hi @giovp - if you could merge in origin/master branch (or rebase onto it) then that should fix some of the broken jobs, thanks

will-moore:
I think the logic there is if you've got no channels then the layer name in `napari` will be e.g. "myImage" but if you've got channels with names then they will be used for layers instead, "DAPI", "GFP" etc.

giovp:
thanks a lot for the prompt reply, from the ngff specs though seems there can be both? a name and channel names? so why would that be overwritten by Omero reader?

https://ngff.openmicroscopy.org/latest/#omero-md

will-moore:
I think this is because the logic in ome-zarr-py was originally built as a napari plugin, so it was designed just to present the best name for each channel. 
If you've only got a `multiscales` name, `napari_ome_zarr` uses that for each layer:

Screenshot 2023-04-03 at 08 36 19

But if you've got an `omero` metadata, we can provide better layer names:

Screenshot 2023-04-03 at 08 38 53

But we only have a single place to store that data: `node.metadata["name"]`, so we have to use it for both cases.

It would certainly be an improvement for `ome-zarr-py` to store *both* the multiscales name and the `omero` channel names, so that both are available to all tools - and then `napari-ome-zarr` could just pick from either to create layer names.

giovp:
> But we only have a single place to store that data: `node.metadata["name"]`, so we have to use it for both cases.
> 
> It would certainly be an improvement for `ome-zarr-py` to store _both_ the multiscales name and the `omero` channel names, so that both are available to all tools - and then `napari-ome-zarr` could just pick from either to create layer names.

ok great, so would it be ok to keep

```
node.metadata["name"] = multiscales[0].get("name")
```

as `name` of image and use a different one here: https://github.com/giovp/ome-zarr-py/blob/58bdf584f10d14c74ed80b9a6ab4a53a58943216/ome_zarr/reader.py#LL390C13-L390C42
like
```
node.metadata["channel_names"] = names
```

?

will-moore:
Yes, `node.metadata["channel_names"] = names` would make sense.
That would be considered a breaking change to `ome-zarr-py` so it would justify bumping the version, to 0.7.0 I guess; I don't think we're ready to go to 1.0.0 yet.
Then we'd need a corresponding release of `napari-ome-zarr` that had dependency of `ome-zarr-py 0.7.0` and updated logic for reading the `channel_names`.

ome_zarr/reader.py
```diff
@@ -383,7 +383,8 @@ def __init__(self, node: Node) -> None:
                     elif contrast_limits is not None:
                         contrast_limits[idx] = [start, end]
 
-            node.metadata["name"] = names
+            # node.metadata["channel_names"] = names
+            node.metadata["names"] = names
```
will-moore, line 387: You were going to go with `node.metadata["channel_names"] = names` here right?

ome_zarr/writer.py
```diff
@@ -306,6 +306,18 @@ def write_multiscales_metadata(
             if axes is not None:
                 ndim = len(axes)
 
+    if isinstance(metadata, dict) and "omero" in metadata["metadata"]:
+        assert isinstance(metadata["metadata"], dict), "metadata must be a `dict`"
```
will-moore, line 310: This `assert` will never fail since you check that it's true on the previous line

will-moore:
Tests are failing with:

```
>       if isinstance(metadata, dict) and "omero" in metadata["metadata"]:
E       KeyError: 'metadata'
```

will-moore:
Failing (only on python 3.8!) with:

```
    def test_omero_metadata(self, metadata: Optional[dict[str, Any]]):
E   TypeError: 'type' object is not subscriptable
```
I think this just needs to be `dict` -> `Dict`

giovp:
thanks a lot for prompt feedback and help @will-moore ! 

Just wanted to clarify one thing regarding metadata: atm, neither color nor windows are mandatory , but if they are present  then they should adhere to type check you outlined here: https://github.com/ome/ome-zarr-py/pull/261#issuecomment-1470124244

therefore, just to recap the specs, they would be as follow:
- metadata MAY contains "omero"
- if it does, then it MUST be a "dict" with either "color" or "window", and not be None
- if it contains "color" then value MUST be string
- if it contains "window" then value MUST be another dict with min, max, start, end with respective values being int.

is this correct? do you think it'd be worth to update the ngff specs to clarify this?
If this is correct I also plan to create PR in: https://github.com/JaneliaSciComp/pydantic-ome-ngff

will-moore:
The following rules are just what the JSON schema states (although there wasn't a lot of justification for this exact set of rules):

 - `omero` is optional, but if present, it MUST contain `channels` https://github.com/ome/ngff/blob/15565cb8301e14fc9ba2fc63b6332a97534cee67/0.4/schemas/image.schema#L105
 - `channels` MUST contain `window` and `color`: https://github.com/ome/ngff/blob/15565cb8301e14fc9ba2fc63b6332a97534cee67/0.4/schemas/image.schema#L98
 - `window` MUST contain `start, end, min, max` (all numbers): https://github.com/ome/ngff/blob/15565cb8301e14fc9ba2fc63b6332a97534cee67/0.4/schemas/image.schema#L78
 - `color` MUST be str https://github.com/ome/ngff/blob/15565cb8301e14fc9ba2fc63b6332a97534cee67/0.4/schemas/image.schema#L91

Certainly the NGFF specs could do with better documentation.
Great to add to pydantic-ome-ngff. I've started looking at how we might use that in ome-zarr-py...

giovp:
I've updated the tests and IO to comply with changes proposed in https://github.com/ome/ngff/pull/202 .
That is,  "windows" and "color" metadata are not mandatory but if present they should be correctly passed.

ome_zarr/writer.py
```diff
@@ -321,6 +321,29 @@ def write_multiscales_metadata(
...
+            if "color" in c:
+                if not isinstance(c["color"], str) and len(c["color"]) != 6:
+                    raise TypeError("`'color'` must be a hex code string.")
+            if "window" in c:
+                if not isinstance(c["window"], dict):
+                    raise TypeError("`'window'` must be a dict.")
+                for p in ["min", "max", "start", "end"]:
+                    if p not in c["window"]:
+                        raise KeyError(f"`'{p}'` not found in `'window'`.")
+                    if not isinstance(c["window"][p], int):
```
will-moore, line 344: I think this could be any type of number (not just an `int`), e.g. float for float data-type images where e.g. max might be less than 1.

ome_zarr/writer.py
```diff
@@ -321,6 +321,29 @@ def write_multiscales_metadata(
             axes = _get_valid_axes(axes=axes, fmt=fmt)
             if axes is not None:
                 ndim = len(axes)
+    print(metadata)
```
will-moore, line 324: best to use `LOGGER.debug` here if needed instead of `print`

giovp: removed, forgot to delete before.

giovp:
fixed conflicts and tests should pass, any chance to get this merged?

ome_zarr/writer.py
```diff
@@ -321,6 +321,28 @@ def write_multiscales_metadata(
...
+                    raise TypeError("`'color'` must be a hex code string.")
+            if "window" in c:
+                if not isinstance(c["window"], dict):
+                    raise TypeError("`'window'` must be a dict.")
+                for p in ["min", "max", "start", "end"]:
+                    if p not in c["window"]:
+                        raise KeyError(f"`'{p}'` not found in `'window'`.")
+                    if not isinstance(c["window"][p], (int, float)):
+                        raise TypeError(f"`'{p}'` must be an int.")
+            group.attrs["omero"] = omero_metadata
```
will-moore, line 345: I just noticed that this line is within the `for` loop - was that intentional? I can't see that it makes much difference to the behaviour just now, but it might cause confusion in future (e.g. if we allow `omero` metadata with no channels, or we only want to assign this once after validating all channels)?
Minor point: "must be an int" message is not exactly True now.

will-moore:
Apologies for the nit-pick comment. Otherwise I'm happy with this.
@joshmoore @sbesson - any 2nd opinion, or OK for me to merge once I'm happy with it?

giovp:
no problem at all, modified accordingly

will-moore:
thanks - I'm away just now but happy for @joshmoore or @sbesson to merge

tests/test_writer.py
```diff
@@ -557,6 +558,59 @@ def test_invalid_transformations(self, coordinateTransformations):
...
+                    self.root, datasets, axes="tczyx", metadata={"omero": metadata}
+                )
+        elif "window" in metadata["channels"][0]:
+            window_metadata = metadata["channels"][0].get("window")
+            if len(window_metadata) < 4:
+                with pytest.raises(KeyError, match=".*`'window'`.*"):
+                    write_multiscales_metadata(
+                        self.root, datasets, axes="tczyx", metadata={"omero": metadata}
+                    )
+        elif "color" in metadata["channels"][0]:
```
will-moore, line 602: Just noticed that with `elif` here, you'll only test the `color` validation if there is no `window` metadata.
Just use `if "color"..` instead.
But also I don't see that `if len(color_metadata) < 6:` below is ever true, since all the `color` in `parametrize` above are length `6`.

will-moore:
Hi @giovp - thanks for the update...
Tests are failing with:

```
    def test_omero(self):
        reader = Reader(parse_url(str(self.path)))()
        image_node = list(reader)[0]
>       omero = image_node.load(Multiscales).zarr.root_attrs.get("omero")
E       NameError: name 'Multiscales' is not defined
```

Also the `Codespell / Check for spelling errors` should be fixed by https://github.com/ome/ome-zarr-py/pull/317 so you can fix that by merging in master branch again.

 Thanks!

giovp:
hi @will-moore , test should be fixed, passes locally, sorry again for getting back to you so late

will-moore (approved):
Great, thanks - apologies for the delay. All tests passing now 👍

giovp:
looks like it can be merged?

will-moore:
@giovp yes, apologies we missed this at the last release.
@joshmoore You OK to merge this?

joshmoore:
Anything else otherwise I'll push out a quick 0.8.4

will-moore:
Thanks Josh - No I don't see any others ready to go. There have been a couple of docs and tox/deps PRs merged recently.
I noticed some build failures https://github.com/ome/ome_zarr_test_suite/actions/workflows/main.yml but I'm not sure why yet.

joshmoore:
_ewwww_

```
RecursionError: maximum recursion depth exceeded in __instancecheck__
```
on `ome_zarr info`

will-moore:
I can't work out exactly where that's coming from or how to reproduce it locally

will-moore:
I wonder if the recursion error is related to similar recursion errors noted at https://github.com/ome/ome-zarr-py/issues/339

Issue #277, opened 2023-05-04
Test suite failure for commit 


snoopycrimecop:
Test suite status: failure see https://github.com/ome/ome_zarr_test_suite/actions/runs/4881080260

jburel:
See #276

jburel:
```
source = <test_suite.GeneratedSource object at 0x7f0743000ca0>
suite = <test_suite.Suite object at 0x7f0743000850>

    def test(source: Source, suite: Suite) -> None:
        """
        Primary test matching all sources with all suites.
        """
        source.setup()
        try:
>           suite(source)

/home/runner/work/ome_zarr_test_suite/ome_zarr_test_suite/test_suite.py:192: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/home/runner/work/ome_zarr_test_suite/ome_zarr_test_suite/test_suite.py:116: in __call__
    path = source()
/home/runner/work/ome_zarr_test_suite/ome_zarr_test_suite/test_suite.py:56: in __call__
    return subprocess.check_output(self.cmd, cwd=str(self.cwd)).decode().strip()
/usr/share/miniconda/envs/test/lib/python3.9/subprocess.py:420: in check_output
    return run(*popenargs, stdout=PIPE, timeout=timeout, check=True,
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

input = None, capture_output = False, timeout = None, check = True
popenargs = (['/home/runner/work/ome_zarr_test_suite/ome_zarr_test_suite/scripts/ome_zarr_create'],)
kwargs = {'cwd': '/tmp/pytest-of-runner/pytest-0/test_create_info_0', 'stdout': -1}
process = <Popen: returncode: 1 args: ['/home/runner/work/ome_zarr_test_suite/ome_zarr...>
stdout = b'', stderr = None, retcode = 1

    def run(*popenargs,
            input=None, capture_output=False, timeout=None, check=False, **kwargs):
        """Run command with arguments and return a CompletedProcess instance.
    
        The returned instance will have attributes args, returncode, stdout and
        stderr. By default, stdout and stderr are not captured, and those attributes
[...]

PR #279, opened 2023-05-08
allow main with empty args

k-dominik:
requiring arguments breaks being able to use `main` as a python entry point. (this also breaks conda recipe currently as conda build tests the entry point)

It seems arguments to `main` are ... a testing feature here? For testing one could also manipulate `sys.argv`...

### DIFF
--- a/ome_zarr/cli.py
+++ b/ome_zarr/cli.py
@@ -2,7 +2,7 @@
 import argparse
 import logging
 import sys
-from typing import List
+from typing import List, Union
 
 from .csv import csv_to_zarr
 from .data import astronaut, coins, create_zarr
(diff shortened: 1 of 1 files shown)


joshmoore:
Thanks, @k-dominik. (I'll note my private glee as I saw you forking the repo :wink:)

FWIW, this showed up after the last release: https://github.com/ome/ome-zarr-py/issues/277#issuecomment-1534463452

ome_zarr/cli.py
```diff
@@ -73,7 +73,7 @@ def csv_to_labels(args: argparse.Namespace) -> None:
     csv_to_zarr(args.csv_path, args.csv_id, args.csv_keys, args.zarr_path, args.zarr_id)
 
 
-def main(args: List[str]) -> None:
+def main(args: Union[List[str], None] = None) -> None:
```
joshmoore, line 76: TIL: https://stackoverflow.com/questions/51710037/how-should-i-use-the-optional-type-hint#:~:text=There%20is%20technically%20no%20difference,are%20exactly%20the%20same%20thing.

Looking forward to the new recommended "List[str] | None" >= Py 3.10

k-dominik: oh right, Optional would've been nicer - it's Monday - sorry :). And I too look forward to the new syntax :)

joshmoore: It's Monday indeed. You can always update in your next PR ;)

0.7.1 pushed and building.

k-dominik: thank you :)

joshmoore: No, no, no. Thank you! :)

Issue #258, opened 2023-03-07
use declarative representation of the spec


d-v-b:
I ran into issues viewing my ome-ngff-formatted zarr data using via the napari plugin. Napari was exiting with an error emitted from [this line](https://github.com/ome/ome-zarr-py/blob/2c4d48972bc3456f72c9a7ba0993c887e80a888d/ome_zarr/reader.py#L300) of this package, which is a `try...except` block that takes ANY exception, and doesn't handle the content of the exception at all.

Unfortunately, swallowing exceptions without a useful error message made it very hard to debug any problems with my data. I suspect this style of coding is a side-effect of taking a procedural approach to the `OME-NGFF` metadata (i.e., "valid metadata" means "a bunch of code ran without errors")
 I think a much better approach would be to define a python class that represents the structure of the metadata and handles parsing / validation of the input. Such a declarative approach (i.e., "valid metadata" means "the program created an object structurally equivalent to the metadata") can produce useful error messages when a single field fails to validate (in my case, the problem was with the version string, but I had to figure that out manually)

For examples of declarative implementations of the OME-NGFF spec, see [iohub](https://github.com/czbiohub/iohub) and [pydantic-ome-ngff](https://github.com/JaneliaSciComp/pydantic-ome-ngff). As the author of the second project, I would love to see this functionality under the umbrella of the official OME organization.

joshmoore:
Hi @d-v-b. Thanks for opening this! I'll reiterate the :+1: from my gitter conversation with you as well as the :+1::heart: from the Zarr community meeting last night, and additionally a :100: to everyone in the community that we find ways to work together on fewer libraries rather than everyone needing to roll their own. (Maybe we need a community meeting ... _per language?_ ... just on that topic.)

I haven't gone through the code in detail but I know we need it. The Java crew that's met at a several hackathons similarly started pushing to https://github.com/ome/ome-ngff-java, so without putting too much thought into the completeness of it, here's a list of TODOs that I'd suggest:

 - [ ] Fix or capture the exception handling mentioned above :smile:
 - [ ] Transfer the repo to @joshmoore (Davis)
   - [ ] who transfers to @ome  (Josh)
   - [ ] or say if you'd prefer a new repo be created (Davis/Josh)
   - cf. https://github.com/JaneliaSciComp/pydantic-ome-ngff/issues/3
 - [ ] Reach out to iohub about possible overlap
 - [ ] Consider renaming to ome-ngff-py (i.e. the base model in each language)
[...]

PR #281, opened 2023-05-09
POC validation of NGFF with pydantic_ome_ngff

will-moore:
See #258.

This PR tests one possible approach for using https://github.com/JaneliaSciComp/pydantic-ome-ngff/ in `ome-zarr-py`:
 - Uses appropriate class to parse JSON data when reading NGFF data

However, this doesn't use `pydantic-ome-zarr` when *writing* data yet, as the integration there will likely have to be much closer

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,5 +1,5 @@
 [settings]
-known_third_party = dask,numcodecs,numpy,pytest,scipy,setuptools,skimage,zarr
+known_third_party = dask,numcodecs,numpy,pydantic_ome_ngff,pytest,scipy,setuptools,skimage,zarr
 multi_line_output = 3
 include_trailing_comma = True
 force_grid_wrap = 0
--- a/ome_zarr/reader.py
+++ b/ome_zarr/reader.py
@@ -298,8 +298,9 @@ class Multiscales(Spec):
         LOGGER.info("datasets %s", datasets)
 
         for multi in multiscales:
-            # raises Exception if can't be parsed
-            Multiscale.parse_obj(multi)
+            # raises Exception if can't be parsed. Only v0.4 supported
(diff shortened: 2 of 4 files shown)


will-moore:
With https://github.com/JaneliaSciComp/pydantic-ome-ngff/pull/7 merged and released as https://github.com/JaneliaSciComp/pydantic-ome-ngff/releases/tag/v0.2.3 tests are passing except for python 3.8 which isn't supported by `pydantic-ome-zarr`.

joshmoore:
@will-moore: what were your plans here? Is this viable? Waiting on something/someone else?

will-moore:
I guess there's a few points to discuss... - You made a good checklist at https://github.com/ome/ome-zarr-py/issues/258#issuecomment-1461487408 which is mostly still outstanding.

 - Do we want to introduce this dependency on pydantic-ome-ngff yet? Cost/benefit?
 - pydantic-ome-ngff requires `python = "^3.9"`. Would we want to handle python 3.8 by optionally not installing pydantic-ome-ngff. Or we drop 3.8 support. Or we try add 3.8 support to pydantic-ome-ngff?
 - The usage of pydantic-ome-ngff is very minimal currently (just in the reading of NGFF). This is probably where it's most useful, Do we stop there and only use it when reading or should we look at using it when writing NGFF too?
 - pydantic-ome-ngff only supports v0.4 OME-Zarr. It's possible we might want to release v0.5 support in `ome-zarr-py` before it's supported in pydantic-ome-ngff. Are we OK with that possibility?

clbarnes:
> Or we drop 3.8 support

Numpy has [already dropped 3.8](https://numpy.org/neps/nep-0029-deprecation_policy.html) in new releases; most of the data ecosystem follows suit.

Issue #284, opened 2023-06-20
Writing ome-zarr in a serial fashion, one 2D slice at a time


ckolluru:
Hi

Is there a good way to setup writing to zarr format using this package, by writing one 2D XY slice at a time?

My current workflow is below:
```
import zarr

# Create a new zarr folder
ZARR_STORE = zarr.open(directory, mode="w")
ch0 = ZARR_STORE.zeros("ch0", shape=(num_time_points, 1, Y_SHAPE_ZARR, X_SHAPE_ZARR), chunks=(1, 1, Y_SHAPE_ZARR, X_SHAPE_ZARR), dtype="i2")

print('Zarr directory tree')
print(ZARR_STORE.tree())

dataset = ZARR_STORE["ch0"]

for i in range(num_time_points):
     dataset[i, 0, :, :] = image   # image at timepoint i
```

I can create a zarr directory structure this way, but I'm unable to open the stack in Fiji after installing the MoBIE plugin and trying the Open OME-Zarr from file system command. Am I missing something here? I'm assuming this package should let me write images serially from python and open in Fiji.  

I checked the documentation but write_image seems to need the full numpy array beforehand.

Thanks,
Chaitanya

ckolluru:
@will-moore : thanks for your reply.

I'm able to create the test image using the script [here](https://ome-zarr.readthedocs.io/en/stable/python.html#writing-ome-ngff-images) and open that in BigDataViewer in Fiji via the MoBIE plugin. It looks fine.

I don't think dask would work in my case since I'm collecting images and different timepoints across several hours. I would like to have the zarr dataset built up during acquisition and ready to go.

I tried running the ome-ngff validator on my dataset and it initially gave an error that it cannot find the .zattrs file. In fact, I can only see a .zgroup file in my folder. This is the folder structure:

filename.zarr
-- ch0
------.zarray
------0.0.0.0
------1.0.0.0
------2.0.0.0
....

--.zgroup

I tried creating a .zattrs file with the text below, and the validator now thinks the schema is correct, but raises an **Error: Dimension separator must be / for version 0.4**. I'm not creating a multiresolution pyramid.

```
{
    "multiscales": [
        {
            "axes": [
		{
		    "name": "t",
		    "type": "time"
		},
                {
                    "name": "z",
                    "type": "space",
		    "unit": "micrometer"
                },
                {
                    "name": "y",
                    "type": "space",
		    "unit": "micrometer"
                },
                {
                    "name": "x",
[...]

PR #285, opened 2023-06-22
View in validator

will-moore:
I had another case where I pasted some python script to tell a user how to serve the local files so that they can be viewed in ome-ngff-validator etc. at https://github.com/ome/ome-zarr-py/issues/284

It occurred to me that I've done this a few times, so instead we should just ship the code.

This PR allows you to do:

```
$ ome_zarr view path/to/image.zarr
```

and it will serve that `path/to` directory with the correct CORS headers and open your browser at https://ome.github.io/ome-ngff-validator/?source=http://localhost:8000/image.zarr to view your data (and allow you to open it in vizarr etc).

### DIFF
--- a/docs/source/cli.rst
+++ b/docs/source/cli.rst
@@ -21,6 +21,14 @@ Use the `ome_zarr` command to interrogate Zarr datasets::
     # Local data (after downloading as below)
     $ ome_zarr info 6001240.zarr/
 
+view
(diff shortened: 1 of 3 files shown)


joshmoore (approved):
Nice! I recently needed exactly this (I think I tried using `file:///`).

PR #290, opened 2023-07-03
Changes to scale.py and writer.py to support scaling methods.

sandip-shah:
Based on this - https://github.com/scikit-image/scikit-image/issues/4294
multichannel=False nees to be changed to channel_axis=None on lines 190 & 201.

In writer.py line 885
mip = scaler.nearest(image)
should be
mip = getattr(scaler, scaler.method)(image)
(else it always calls the 'nearest' scale method).

### DIFF
--- a/ome_zarr/scale.py
+++ b/ome_zarr/scale.py
@@ -187,7 +187,7 @@ class Scaler:
                 base,
                 downscale=self.downscale,
                 max_layer=self.max_layer,
-                multichannel=False,
+                channel_axis=None,
             )
         )
 
@@ -198,7 +198,7 @@ class Scaler:
                 base,
                 downscale=self.downscale,
                 max_layer=self.max_layer,
-                multichannel=False,
+                channel_axis=None,
             )
         )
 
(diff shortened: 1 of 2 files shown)


will-moore:
Thanks for the contribution.
It looks like usage of `channel_axis` requires https://github.com/scikit-image/scikit-image/releases/tag/v0.20.0 so this would require updating the dependency?

Other change looks good...
Running tests now...

sandip-shah:
Yes, you are right.  It would require updating the dependency.

Pardon my ignorance here - but how does this work?  Do you want me to specify the dependencies (in environment.yml and setup.py, and maybe requirements) in the PR?

Since both the changes are tied with each other, we may have to do that?

Thanks,

Sandip

will-moore:
Hi - apologies for my delayed response (been away)...
Yes, please update `setup.py`, and it may be worth doing `environment.yml` too, although that's less critical as it's usually installing the latest versions of everything each time (rather than updating), and the tests that use it are passing.

Issue #294, opened 2023-08-02
writer.write_image requires fsspec>=0.8.0


elyall:
I installed a package that has a `fsspec<0.8` dependency which broke my tiff to ome-zarr conversion pipeline. Having `fsspec=0.7.4` installed results in the error:
```
AttributeError: 'FSMap' object has no attribute 'setitems'
```

`FSMap` gains the `setitems` method in `fsspec` version `0.8.0` [[source](https://github.com/fsspec/filesystem_spec/blob/9180c4544ece9cd8c1ca4ff80261d2aee755791d/fsspec/mapping.py)], so the solution is to set a minimum version of `fsspec>=0.8` in the setup.py file [here](https://github.com/ome/ome-zarr-py/blob/49b8c9e6276fd6c9efb616a8b1520dcf806036cf/setup.py#L22). 

<details>

<summary>Code to reproduce</summary>
The code from the documentation reproduces the error.

```python
path = "test_ngff_image.zarr"
os.mkdir(path)

mean_val=10
size_xy = 128
size_z = 10
rng = np.random.default_rng(0)
data = rng.poisson(mean_val, size=(size_z, size_xy, size_xy)).astype(np.uint8)

# write the image data
store = parse_url(path, mode="w").store
root = zarr.group(store=store)
write_image(image=data, group=root, axes="zyx", storage_options=dict(chunks=(1, size_xy, size_xy)))
```
</details>

<details>

<summary>Full error</summary>

```shell
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[1], line 24
     22 store = parse_url(path, mode="w").store
     23 root = zarr.group(store=store)
---> 24 write_image(image=data, group=root, axes="zyx", storage_options=dict(chunks=(1, size_xy, size_xy)))
     25 # # optional rendering settings
     26 # root.attrs["omero"] = {
     27 #     "channels": [{
   (...)
     32 #     }]
     33 # }

File [~/micromamba/envs/test/lib/python3.10/site-packages/ome_zarr/writer.py:492], in write_image(image, group, scaler, chunks, fmt, axes, coordinate_transformations, storage_options, compute, **metadata)
    490 else:
    491     mip, axes = _create_mip(image, fmt, scaler, axes)
--> 492     dask_delayed_jobs = write_multiscale(
    493         mip,
    494         group,
    495         chunks=chunks,
    496         fmt=fmt,
    497         axes=axes,
[...]

PR #295, opened 2023-08-02
Set minimum fsspec version

elyall:
Peg dependency `fsspec`'s minimum version to `0.8.0` to address https://github.com/ome/ome-zarr-py/issues/294.


will-moore (approved):
Running the demo test code with `fsspec==0.7.4` reproduces the reported exception and updating to `fsspec==0.8.0` fixes the issue.
Also, tests above are passing.

will-moore:
Thanks for your contribution.

PR #299, opened 2023-08-21
Fix compute() on plate grid

### LINKED ISSUE #297: Dask arrays for HCS nodes do not `compute` to NumPy arrays
The Dask arrays returned by the reader does not `compute` the underlying chunk type, which should be NumPy arrays. The FOV/image level behaves as normal.

This is likely a bug since Dask [documentation](https://docs.dask.org/en/latest/generated/dask.array.Array.compute.html) claims:

> This turns a lazy Dask collection into its in-memory equivalent. For example a Dask array turns into a NumPy array and a Dask dataframe turns into a Pandas dataframe. The entire dataset must fit into memory before calling this operation.

To reproduce:

```py
import dask.array as da
from ome_zarr.reader import Reader

location = parse_url("hcs.ome.zarr")
reader = Reader(location)

data = next(reader()).data[0]
# this passes
assert isinstance(data, da.Array), type(data)

result = data.compute()
# this fails
assert not isinstance(result, da.Array), type(result)
```

will-moore:
Fixes #297

This fixes the issue above: The `get_tile()` already returns a dask array so it doesn't need to be wrapped in `dask.delayed()`.

Test using the code sample on the ticket above.

### DIFF
--- a/ome_zarr/reader.py
+++ b/ome_zarr/reader.py
@@ -7,7 +7,6 @@ from typing import Any, Dict, Iterator, List, Optional, Type, Union, cast, overl
 
 import dask.array as da
 import numpy as np
-from dask import delayed
 
 from .axes import Axes
 from .format import format_from_version
@@ -420,7 +419,7 @@ class Well(Spec):
         self.img_metadata = image_node.metadata
         self.img_pyramid_shapes = [d.shape for d in image_node.data]
 
-        def get_field(tile_name: str, level: int) -> np.ndarray:
+        def get_field(tile_name: str, level: int) -> da.core.Array:
             """tile_name is 'row,col'"""
             row, col = (int(n) for n in tile_name.split(","))
             field_index = (column_count * row) + col
@@ -430,10 +429,8 @@ class Well(Spec):
                 data = self.zarr.load(path)
             except ValueError:
                 LOGGER.error("Failed to load %s", path)
-                data = np.zeros(self.img_pyramid_shapes[level], dtype=self.numpy_type)
+                data = da.zeros(self.img_pyramid_shapes[level], dtype=self.numpy_type)
             return data
-
-        lazy_reader = delayed(get_field)
 
         def get_lazy_well(level: int, tile_shape: tuple) -> da.Array:
             lazy_rows = []
@@ -447,11 +444,7 @@ class Well(Spec):
                         col,
                         level,
                     )
-                    lazy_tile = da.from_delayed(
-                        lazy_reader(tile_name, level),
-                        shape=tile_shape,
-                        dtype=self.numpy_type,
-                    )
+                    lazy_tile = get_field(tile_name, level)
                     lazy_row.append(lazy_tile)
                 lazy_rows.append(da.concatenate(lazy_row, axis=x_index))
             return da.concatenate(lazy_rows, axis=y_index)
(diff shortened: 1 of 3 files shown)


ziw-liu:
Hi @will-moore I installed from this branch with:

```sh
pip install git+https://github.com/will-moore/ome-zarr-py.git@plate_dask_compute
```

And I can see the updated lines in the site package of my environment.

I can confirm that 286a4f2b9364c8f07ea9e843abb8b683ece5f2c7 fixes the plate level. However the well level still have the same issue:

```py
import dask.array as da
from ome_zarr.reader import Reader

# well level in an HCS store
location = parse_url("hcs.ome.zarr/A/1")
reader = Reader(location)

data = next(reader()).data[0]
# this passes
assert isinstance(data, da.Array), type(data)

result = data.compute()
# this fails
assert not isinstance(result, da.Array), type(result)
```

ome_zarr/reader.py
```diff
@@ -541,7 +541,8 @@ def get_tile(tile_name: str) -> np.ndarray:
             LOGGER.debug("LOADING tile... %s with shape: %s", path, tile_shape)
 
             try:
-                data = self.zarr.load(path)
+                # compute() to get the data from dask - we want to return array
+                data = self.zarr.load(path).compute()
```
joshmoore, line 545: Should the computation not happen as late as possible, e.g., during the concatenation below, to make use of dask's scheduler?

will-moore: Aren't we already making use of dask's scheduler? Or do you mean make *better* use of it?

Would that actually give better performance? Not being very familiar with it, I'm reading https://docs.dask.org/en/stable/scheduler-overview.html which says there are gains to be made if you have a large graph and "share many intermediates", but I don't see that we can do that here?

I've simplified the code a fair bit in ee9ef7d to remove the extra layer of `delayed()` wrapping. This looks cleaner and still works as expected. It just means that we call `self.zarr.load(path)` for every tile right at the start, even if we never need to load/compute that region. I don't know if this makes any difference to performance but probably not?

joshmoore: Better use, yes, so that the right things happen parallel.

will-moore:
@ziw-liu Thanks for that - fixed for Wells too now. And I added some tests.
@joshmoore any ideas for why readthedocs has started to fail? (also seen on https://github.com/ome/ome-zarr-py/pull/300)

joshmoore:
No, I don't. The error is:

```
NameError: name 'html_theme' is not define
```

Possibly due to the deprecation back in 2021? https://www.sphinx-doc.org/en/master/changes.html#release-4-1-0-released-jul-12-2021

jburel:
@will-moore @joshmoore This is a problem with readthedocs I faced the same problem on https://github.com/ome/omero-scripts/pull/187
I ended up adding the dependencies see https://github.com/ome/omero-scripts/pull/187/commits/07e130dda4e9b58de423a0d0c48be8dd40190e6d

joshmoore (approved):

will-moore:
That last commit fixes failing tests with a bug that comes from #300.
In that PR, `image_path = image_paths[field_index]` was failing when viewing a Well of e.g. 3 fields, since it tries to build a 2 x 2 grid - the `field_index` of the bottom right spot is 3, but there are only 3 image_paths.

That bug wasn't caught in the tests for that PR since that code is only run when the you call `dask.compute()`.
It was failing in tests above because in this PR, that code gets run immediately, since it's not wrapped in dask.delayed() and we also add tests for loading a single Well.

Issue #248, opened 2023-02-10
FOV names and array names are hard-coded for the HCS dataset reader


ziw-liu:
The [v0.4 spec states](https://ngff.openmicroscopy.org/0.4/index.html#image-layout):

```
# The name of the array is arbitrary with the ordering defined by
# by the "multiscales" metadata, but is often a sequence starting at 0.
```

And ome-zarr-py works with arbitrary array names in the single-FOV images layout, but not in an HCS plate layout. Only integer names starting with "0" is recognized.

It should instead be discovered from metadata (`multiscales.datasets[0].path`).

Also, FOV names not starting with "0" is not supported. The cause is likely this line hard-coding 'first field':

https://github.com/ome/ome-zarr-py/blob/c1302e05998dfe2faf94b0f958c92888681f5ffa/ome_zarr/reader.py#L493

It should instead be discovered from metadata (`well.images[0].path`). [Spec](https://ngff.openmicroscopy.org/0.4/index.html#well-md):

```
The well dictionary MUST contain an images key whose value MUST be a list of JSON objects specifying all fields of views for a given well. Each image object MUST contain a path key whose value MUST be a string specifying the path to the field of view. The path MUST contain only alphanumeric characters, MUST be case-sensitive, and MUST NOT be a duplicate of any other path in the images list. 
```

Edit: separate description for array and position names.

ziw-liu:
According to https://github.com/ome/ome-zarr-py/issues/200#issuecomment-1131586455, one can view all the FOVs by loading individual wells. However this only works for incremental integer FOV path names, not any 'case sensitive, alphanumeric' path names defined in the specification. Possible cause:

https://github.com/ome/ome-zarr-py/blob/c1302e05998dfe2faf94b0f958c92888681f5ffa/ome_zarr/reader.py#L427-L431

will-moore:
Thanks for the feedback.
You're right that there is an assumption made about the path to first image in a Well.
It would be a pretty easy fix to load the Well images for a Well and read that data. It's a bit of a bigger change to read that data for ALL Wells in a plate if you can't assume that all Wells have the same path to first image.

This is a similar type of change to https://github.com/ome/ome-zarr-py/pull/241 where it becomes slower to load plate data if you don't make assumptions about all the Wells being the same.

Reading Well images may not have such a big impact as the changes on that PR, but it raises similar questions:
For plates where all Wells have the same path to first image, we would prefer NOT to have to load the Images for every Well to check.

[...]

PR #300, opened 2023-08-23
Avoid hard-coded field-of-view path "0"

aeisenbarth:
Fixes https://github.com/ome/ome-zarr-py/issues/248

The current code assumes image paths under a well always match field of view indices, especially that "0" exists. However, the NGFF spec allows alphanumeric paths.

This change looks up the actual image path from the first well. This should not have a performance impact since the first well's `.zattrs` is read anyways. It is still assumed all wells have the same first image path (which the previous code did as well). In that case, an array of zeros is read (as previously).

### DIFF
--- a/ome_zarr/reader.py
+++ b/ome_zarr/reader.py
@@ -424,7 +424,8 @@ class Well(Spec):
             """tile_name is 'row,col'"""
             row, col = (int(n) for n in tile_name.split(","))
             field_index = (column_count * row) + col
-            path = f"{field_index}/{level}"
+            image_path = image_paths[field_index]
+            path = f"{image_path}/{level}"
             LOGGER.debug("LOADING tile... %s", path)
             try:
                 data = self.zarr.load(path)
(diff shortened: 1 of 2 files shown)


will-moore:
Looks good, and works for me. Is this still a `[WIP]`?
Don't know why the docs build failed, but it's not because of this PR

aeisenbarth:
I still wanted to look into how to cover this by tests. Maybe by parametrizing [`test_multiwells_plate`](https://github.com/ome/ome-zarr-py/blob/30f55fccb2528b69482139a1a7592c9d0ed5e2da/tests/test_reader.py#L88C17-L88C34) with `fields = map(str, range(3))` and `fields = ["img1", "img2", "img3"]`.

aeisenbarth:
I added a test. Since failed image lookups fail gracefully and return zeros, I had to change the image values of the test arrays to something non-zero.

The alphanumeric test case failed on master and succeeds on this branch.

joshmoore (approved):

PR #303, opened 2023-09-05
Dump sample scripts - maybe add to docs

will-moore:
Several forum discussions etc have asked for more detailed examples on how to write OME-Zarr.

Added 2 new examples using different strategies to write images that are bigger than memory:

 - Write tiles to disk until we have whole image. Then downsample to build OME-Zarr.
 - Use dask to stitch delayed tiles together, then `write_image(dask_data)`.


joshmoore:
Anything you need here, @will-moore ?

will-moore:
@joshmoore Thanks for the ping. Do you think these look useful and they are in line with "best practice" for how to achieve what was asked for in https://forum.image.sc/t/writing-tile-wise-ome-zarr-with-pyramid-size/85063?

If so, I'll tidy them up, add a few words of intro/explanation etc.

joshmoore:
If we're sending people to this from image.sc, shall we get them into the repo?

will-moore:
@josh - just tested and tweaked those 2 scripts... Ready to review/merge now, thx.

Issue #304, opened 2023-09-05
`auto_mkdir`  not propagated resulting in FileNotFoundError during `write_image` 


npeschke:
I am trying to save a multidimensional image using ome-zarr, however saving fails with the following error:
```
FileNotFoundError: [Errno 2] No such file or directory: '/path/to/input_image.ome.zarr/0/0/0/0'
```
I am using the following code to save the image:
```python
loc = parse_url(path, mode="w").store
group = zarr.group(store = loc)

n_channels = input_image.shape[0]
channels = [f"Channel{i}" for i in range(n_channels)]

group.attrs["omero"] = {
    "name": self.DEFAULT_INPUT_IMAGE_NAME,
    "channels": [{"label":channel, "color":self.channel_colors[i], "active":True} for i, channel in enumerate(channels)]
}

write_image(input_image, group = group, axes = "cyx", storage_options=dict(chunks=(1, 1024, 1024)))
```

The following directory structure is generated:
```
input_image.ome.zarr/
├── 0
│   └── .zarray
├── .zattrs
└── .zgroup
```

I've narrowed down the issue to `auto_mkdir` not being propagated to the `LocalFileSystem` object in the store

Is this intended behaviour?

will-moore:
Hi, thanks for the feedback.
I was just trying to reproduce this issue locally, and failing until I was notified of https://github.com/ome/ome-zarr-py/issues/302, which I believe is the same issue.

On updating my local version of `fsspec-2022.2.0` to `fsspec-2023.9.0` also (`s3fs-2022.2.0` to `s3fs-2023.9.0`) I now see the same issue.
So it looks like the short-term solution is to install `fsspec-2022.2.0` (assuming you have a more recent version) while we try to fix the issue.

will-moore:
Seems the commit at https://github.com/fsspec/filesystem_spec/commit/285094f0771dcc98467bf0a0e5e5b14a6cf096b3 is involved in this issue: https://github.com/fsspec/filesystem_spec/commit/285094f0771dcc98467bf0a0e5e5b14a6cf096b3

will-moore:
Same issue reported at #305 (with stack trace) and also causes test failures at #302.

See discussion on this issue at https://github.com/fsspec/filesystem_spec/commit/285094f0771dcc98467bf0a0e5e5b14a6cf096b3

davidpross:
[...]

PR #301, opened 2023-08-24
Add sphinx_rtd_theme to docs dependencies

will-moore:
Try to fix sphinx readthedocs build - similar to:
https://github.com/ome/omero-scripts/commit/07e130dda4e9b58de423a0d0c48be8dd40190e6d

PR #307, opened 2023-09-13
Exclude fsspec 2023.9.0

will-moore:
In https://github.com/ome/ome-zarr-py/pull/301/files we needed to avoid the latest fsspec version `fsspec 2023.9.0` due to avoid https://github.com/fsspec/filesystem_spec/issues/1357 (also reported at https://github.com/ome/ome-zarr-py/issues/304).

The fix at https://github.com/fsspec/filesystem_spec/pull/1358 is merged and should be included in the next fsspec release so we only need to exclude the current release version.

This makes the corresponding change to setup.py

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ install_requires += (["numpy"],)
 install_requires += (["dask"],)
 install_requires += (["distributed"],)
 install_requires += (["zarr>=2.8.1"],)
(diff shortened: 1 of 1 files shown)


joshmoore (approved):
Thanks, @will-moore. Interested to see how the filter list looks over time...

I'll create an immediate 0.8.1

PR #313, opened 2023-09-29
Add conda to installation options

dstansby:
Since `ome-zarr` is available on conda-forge, I thought it would be good to list that option in the installation instructions.


will-moore:
Thanks!

PR #315, opened 2023-09-29
Add dask to intersphinx mapping

dstansby:
Since there's some dask references (e.g. on https://ome-zarr.readthedocs.io/en/stable/api/writer.html#ome_zarr.writer.write_image), I've added dask to the intersphinx mapping.

I also updated the Python link, which was showing up as a redirect before.


will-moore:
Thanks!

PR #317, opened 2023-10-02
update docs to include write example for HCS dataset

elyall:

PR #320, opened 2023-10-23
fix layout

jburel:
Follow up from https://github.com/ome/ome-zarr-py/pull/317


will-moore (approved):
looks good at https://ome-zarr--320.org.readthedocs.build/en/320/python.html#writing-hcs-datasets-to-ome-ngff

PR #326, opened 2023-11-20
Simplify first example in docs

dstansby:
I personally found the first example encountered in the docs quite complicated. This PR simplifies it so it's a minimal working example of using the basic functionality of writing an ome-zarr file. I've then split up the extensions to the example (rendering settings and adding labels) to extra sections below, so someone new to ome-zarr-py can build up their code bit by bit instead of being overwhelmed by it all in one go.


will-moore:
Sample works fine and is certainly a bit simpler, thanks

PR #327, opened 2023-11-20
Add tests on python 3.11

dstansby:
since Python 3.11 has been out for a while, I thought it was worth adding tests


will-moore:
👍 Thanks

PR #328, opened 2023-11-20
Only install required deps when testing

dstansby:
Since tox creates a new environment and installs dependencies there, no need to install all the dependencies beforehand.


will-moore:
Looks good and all checks are passing. Thanks

PR #334, opened 2023-11-28
Clarify that dask input images are never computed

dstansby:
`compute=False` is hardcoded in _write_dask_image, so clarify the docstring and remove the unused compute argument from _write_dask_image to avoid potential confusion.


dstansby:
Woops, this isn't true!

Issue #319, opened 2023-10-11
Readme should include how to install sphinx-rtd-theme


iandobbie:
If you try to build the docs locally with sphinx as suggested in the README you need to install the sphinx-rtd-theme as well as sphinx. This should probably be added to the readme. 

eg
```
 It can be built locally with::
 
    $ pip install sphinx
    $ pip install sphinx-rtd-theme
    $ sphinx-build -b html docs/source/ docs/build/html
```

PR #336, opened 2023-12-04
Fix doc install instructions in README

dstansby:
Fixes https://github.com/ome/ome-zarr-py/issues/319


will-moore (approved):
LGTM, works as expected

Issue #325, opened 2023-11-20
Releases on GitHub out of date


dstansby:
At https://github.com/ome/ome-zarr-py/releases the latest release is version 0.2.0, but that was a long time ago!

<img width="251" alt="Screenshot 2023-11-20 at 12 00 40" src="https://github.com/ome/ome-zarr-py/assets/6197628/722beb29-b87d-4386-8955-71ed279ac312">

I'd suggest either deleting the existing old releases, or to start creating releases for new versions so it doesn't look like the last release was over two years ago.

joshmoore:
Thanks for the heads up, @dstansby. Now updated:

![image](https://github.com/ome/ome-zarr-py/assets/88113/eae7a220-c9d6-456d-ab5d-e415ad0ae681)

PR #54, opened 2020-10-19
Pin tifffile<2020.09.22 on Python 3.6

joshmoore:
`pip install --upgrade napari` on Python 3.6 attempts
to update tifffile to versions which require Python 3.7+

PR #346, opened 2024-01-17
Pyproject.toml & setuptools_scm support

joshmoore:
As a follow on from https://github.com/ome/ome-zarr-py/issues/325, this moves ome-zarr-py to the use of `setuptools_scm` (like https://github.com/ome/omero-cli-zarr/pull/159) so that tags **with their releases** can be created directly from the GitHub UI. cc: @dstansby 

To simplify this, I've gone ahead and migrated setup.py to pyproject.toml. I imagine this is a template that we can follow elsewhere. The major discrepancy in the mapping between the two was that `test_requires` is now an optional dependencies key which may require some adjustments.

Major question from my side is which python version to pin to. Note that zarr-python has [dropped 3.8 for the next release](https://github.com/zarr-developers/zarr-python/pull/1557) and there is already discussion about dropping 3.9 too based on [SPEC-0000](https://scientific-python.org/specs/spec-0000/).

A minor question is whether or not we still want to write the version number to a file or to just leave folks to use `importlib.metadata.version("ome_zarr")`.


joshmoore:
Still needs a replacement for bsdist & wheel creation, e.g., https://stackoverflow.com/questions/58753970/how-to-build-a-source-distribution-without-using-setup-py-file

joshmoore:
Trying to get this green to see if it addresses @akhanf's issues in #353.

joshmoore:
Reviewers requested, largely with an `OR`.

pyproject.toml
```diff
@@ -0,0 +1,62 @@
...
+name = "ome-zarr"
+readme = "README.rst"
+dynamic = ["version"]
+description="Implementation of images in Zarr files."
+
+requires-python = ">3.8"
+
+dependencies = [
+    "dataclasses;python_version<'3.7'",
+    "tifffile<2020.09.22;python_version<'3.7'",
```
will-moore, line 20: Do we need the `<'3.7'` support here, since we have `requires-python = ">3.8"`

will-moore:
Re: minor question - we haven't previously written the version number to a file, right?

will-moore:
> Reviewers requested, largely with an `OR`.

What's an `OR`?

joshmoore:
Sorry, I mentioned recently elsewhere to Seb, that I'm not expecting *all* of these reviewers (i.e. with an `AND`), but any of you would be great (`OR`).

pyproject.toml
joshmoore: No! Thanks. Let's get rid of these.

joshmoore:
> Re: minor question - we haven't previously written the version number to a file, right?

I'm not sure what you're asking. Previously, we've hard-coded the version in setup.py: https://github.com/zarr-developers/zarr-developers.github.io/pull/105#issuecomment-1957313681

will-moore:
With this branch...

```
pip install -e .
...
Successfully built ome-zarr
Installing collected packages: ome-zarr
ERROR: pip's dependency resolver does not currently take into account all the packages that are installed. This behaviour is the source of the following dependency conflicts.
napari-ome-zarr 0.5.2 requires ome-zarr>=0.3.0, but you have ome-zarr 0.0.0 which is incompatible.
omero-cli-zarr 0.5.5.dev10+g893d4ee requires ome-zarr>=0.5.0, but you have ome-zarr 0.0.0 which is incompatible.
Successfully installed ome-zarr-0.0.0
```

Tried getting version number as described:

```
>>> import importlib
>>> importlib.metadata.version("ome_zarr")
  File "<stdin>", line 1, in <module>
AttributeError: module 'importlib' has no attribute 'metadata'
```
After googling alternative approach:
```
>>> import pkg_resources
>>> pkg_resources.get_distribution("ome_zarr").version
'0.0.0'
```

sbesson (commented):
At the code level, the changes look reasonable.

On the build system, I am not strongly opinionated about either form and would need to read more to have an informed view. My primary feedback is practical as updating the style/layout can introduce some churn with effectively zero value for end-users and additional burden for maintainers e.g. when looking at the code history. This is mostly related to the suggestion that the template could be followed elsewhere. Only suggestion is that I would weigh the impact before rolling out the new style to all OME Python repositories.

On the `bumpversion` vs `setuptools-scm` (or similar) workflow, it all boils down to the usual trade-off between a quick release workflow via a tag only vs managing the version lifecycle explicitly. Here again, I don't have a strong opinion as we already have a mixture of workflows. The biggest caveats of the `setuptools-scm` approach that I remember is the requirement for a Git checkout and also tags (esp. for forks).
The readme section about the release process should likely be updated accordingly.

pyproject.toml
```diff
@@ -0,0 +1,61 @@
...
+description="Implementation of images in Zarr files."
+
+requires-python = ">3.8"
+
+dependencies = [
+    "tifffile",
+    "numpy",
+    "dask",
+    "distributed",
+    "zarr>=2.8.1",
```
dstansby, line 23:
```suggestion
    "zarr>=2.8.1,<3",
```

pyproject.toml
```diff
@@ -0,0 +1,61 @@
...
+[project]
+name = "ome-zarr"
+readme = "README.rst"
+dynamic = ["version"]
+description="Implementation of images in Zarr files."
+
+requires-python = ">3.8"
+
+dependencies = [
+    "tifffile",
```
dstansby, line 19:
```suggestion
```
This doesn't seem to be an actual dependency?

dstansby (approved):
I see a ping to me from many months ago! This looks great to me, and as well as hopefully helping with https://github.com/ome/ome-zarr-py/issues/353 would also help for folks using `uv` (including myself), that needs a `pyproject.toml` to read from.

On Python support, I *think* SPEC-0 forces that hand of any package that depends on numpy (or other scientific python packages) to follow it, so I'd advocate for >=3.11 here.

pyproject.toml
joshmoore: :+1: This is a remainder from when napari was a dependency in this repo (see https://github.com/ome/ome-zarr-py/pull/54)

will-moore (approved):
LGTM. python >= 3.11 sounds good, same as zarr-python v3

joshmoore:
Ok. Merging as we move towards 0.11.0.

PR #349, opened 2024-01-26
Support init of ome_zarr_py.io.ZarrLocation with zarr.storage.FSStore

berombau:
By supporting FSStore as input, more complex ZarrLocations are possible like private remote stores.

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -1,5 +1,5 @@
 [settings]
-known_third_party = dask,numcodecs,numpy,pytest,scipy,setuptools,skimage,zarr
+known_third_party = dask,fsspec,numcodecs,numpy,pytest,scipy,setuptools,skimage,zarr
 multi_line_output = 3
 include_trailing_comma = True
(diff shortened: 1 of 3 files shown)


berombau:
@joshmoore Sure, I added some basic tests. More bespoke testing would require a dependency on s3fs or universal_pathlib, but this covers my use case.

berombau:
It's ready to go in the next release, I think. I worked with `universal-pathlib` and `moto` in a spatialdata fork and it works well, but it's better to work on that if needed in a separate PR.

joshmoore:
Thanks, all! Due to the change of the API, going to release this now as 0.9.0. @pgarrison, interested to hear if it solves all your troubles. 🙂

pgarrison:
@joshmoore yes I was [able to use this pattern](https://github.com/bioio-devs/bioio-ome-zarr/pull/25) to read a public `s3://` path without authentication, thanks all!

PR #358, opened 2024-03-04
Allow to change downsample function

nhatnm52:
- Define property ``func`` to get downsample function.
- Use ``scaler.func`` instead of ``scaler.nearest`` in function ``_create_mip``.

### DIFF
--- a/ome_zarr/scale.py
+++ b/ome_zarr/scale.py
@@ -77,9 +77,7 @@ class Scaler:
 
     def scale(self, input_array: str, output_directory: str) -> None:
         """Perform downsampling to disk."""
-        func = getattr(self, self.method, None)
-        if not func:
-            raise Exception
+        func = self.func
 
         store = self.__check_store(output_directory)
         base = zarr.open_array(input_array)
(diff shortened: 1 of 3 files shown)


nhatnm52:
Hi Will, I have a private project and I need to write an image into a pyramid image with ``local_mean``  method, your ``_create_mip`` function only allows ``nearest`` so I use ``func`` property to allow this. I set up a test for 2 cases of Scaler class. I hope this change will be helpful.
Nhat Nguyen.

will-moore:
Hi, thanks for the clarification.
Looking at the tests, it occurred to me that for several tests it's not really checking that the correct downsampling method was chosen because we're only checking the dimensions of the downsampled data, not the pixels.

E.g. if I edit your `scale.py` to always use `nearest' like this:

```
@property
    def func(self) -> Callable[[np.ndarray], List[np.ndarray]]:
        """Get downsample function."""
        func = getattr(self, "nearest", None)
```

the only tests that fail are the `def test_scale_dask()` since that's the only one that checks the pixels.

Instead of adding the extra `directly` parameter to every test, you could instead do something like this: (you don't need to do this for all tests, especially the ones that are marked as failing. Maybe just do it for `test_local_mean` and `test_scale_dask`

e.g.  `def test_local_mean()`
```
        downscaled = scaler.local_mean(data)
        scaler.method = "local_mean"
        assert np.array_equal(downscaled, scaler.func(data))
        self.check_downscaled(downscaled, shape)
```

I also noticed that our tests don't actually test `scaler.scale()` anywhere, but that's not a blocker for this PR.

Cheers!

nhatnm52:
Hi, I rolled back pytest.fixture as original and added testing by pixel-wise for ``local_mean``, ``nearest`` and ``scale_dask``, I skipped test cases using ``skip`` or ``xfail`` flags. This test will ensure that the ``func`` property performs its function.

Have a nice working day.

Issue #364, opened 2024-03-25
write_image with scalar chunks only sets chunks of last axis


dstansby:
I'm trying to set the same chunk size in each dimension when writing a file. In `zarr` this is easily done by giving an integer value to the `chunks` parameter. From the docs for `zarr.creation.create`:

> If an int, the chunk size in each dimension will be given by the value of chunks. 

If I do this in `ome-zarr` using the following code, only the chunk size of the last dimension is set:
```python
write_image(
    image=arr,
    group=root,
    axes="xyz",
    storage_options={'chunks': 64},
    scaler=None
)
```
The chunks on the resulting dataset are (512, 512, 64). I would expect them to be (64, 64, 64).

PR #365, opened 2024-03-25
Fix scalar chunking

dstansby:
Fixes https://github.com/ome/ome-zarr-py/issues/364. I noticed there wasn't a changelog, so I added one, I hope that's okay? I think it's important to communicate bug fixes like this.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,11 @@
 - Correctly specify maximum compatible fsspec version. ([#338](https://github.com/ome/ome-zarr-py/pull/338))
 - Add tests on Python 3.12. ([#338](https://github.com/ome/ome-zarr-py/pull/338))
 - Write OMERO metadata. ([#261](https://github.com/ome/ome-zarr-py/pull/261))
+- Fixed chunking when a scalar value for chunks is given. Previously
+  passing ``storage_options={"chunks": <int>}`` only set the chunk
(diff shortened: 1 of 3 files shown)


will-moore:
Thanks for the contribution and the changelog.
Could you add a link from the changelog to this issue? In other repos do this (e.g. https://github.com/ome/omero-web/blob/master/CHANGELOG.md) which also allows you to be more concise in the changelog if you wish?

The change looks good and makes sense. The only question I have is whether anyone would consider this a breaking API change (and therefore require a major version bump) or if this is really just a bug fix? cc @joshmoore

will-moore (approved):
Thanks for the update. LGTM.

Previously, doing `write_image(....storage_options={"chunks": 32})` would have resulted in chunks of e.g `(sizeZ, sizeY, 32)` but this feature wasn't documented or have tests and the behaviour was different from zarr-python.
So, I think we could consider a change in this behaviour to be a bug-fix rather than a breaking API change, but happy to hear other opinions @joshmoore @sbesson ?

joshmoore (changes requested):
Please see https://github.com/ome/ome-zarr-py/blob/master/CHANGELOG.md

dstansby:
👍 changelog in the right place now - I think tests are failing because codecov now requires a token to upload coverage reports.

PR #381, opened 2024-05-26
Drop support for Python 3.8

dstansby:
As per [SPEC 0](https://scientific-python.org/specs/spec-0000/), don't need to support Python 3.8 or 3.9 any more. As per a comment below, I've only dropped support for 3.8 here.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ['3.8', '3.9', '3.10', '3.11', '3.12']
+        python-version: ['3.9', '3.10', '3.11', '3.12']
         os: ['windows-latest', 'macos-latest', 'ubuntu-latest']
     steps:
       - uses: actions/checkout@v4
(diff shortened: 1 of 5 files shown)


joshmoore:
Thanks, @dstansby. I've recently heard of folks having issues on various platforms so let's see if any of that surfaces here. If not, I can definitely see getting this into the next largish bump.

will-moore:
I was having issues getting Ice bindings installed on python 3.10 (our docs state that only python 3.9 was supported). In the end we found a work-around but for users who have working python 3.9 environments it could cause inconvenience to drop python 3.9 support here before we need to.
Happy to drop 3.8 support, but dropping 3.9 support feels a little early (unless supporting it is a burden to us that we are keen to drop?)

dstansby:
The motivation here was `zarr` follows SPEC 0, so won't (in theory) be supporting Python 3.8/3.9 now anyway, so it makes sense to mirror that here? Happy to change this to just drop 3.8 though.

dstansby:
👍 updated to just drop support for 3.8

joshmoore:
Thanks again, @dstansby. Let's get this in and then we can always iterate from there.

PR #385, opened 2024-06-17
Pin numpy<2 in setup.py

will-moore:
numpy 2.0.0 has been released with breaking changes. Avoid installing till this can be reviewed etc.
Same as https://github.com/ome/omero-py/pull/414

cc @sbesson

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -33,4 +33,3 @@ jobs:
         with:
           file: ./coverage.xml
           fail_ci_if_error: true
-          token: ${{ secrets.CODECOV_TOKEN }}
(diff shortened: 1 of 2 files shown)


will-moore:
Tests failing with `Error: Codecov token not found. Please provide Codecov token with -t flag.`

joshmoore:
I *do* see the token, but the environment variable wasn't set. I've pushed a commit that matches what I did in https://github.com/zarr-developers/zarr-python/pull/1647. Waiting to see if it goes green now.

joshmoore:
Somehow codecov is only not finding the token on 3.12 (😕) I'm going to merge as is since this is clearly passing most workflows and we may have to come back to it soon.

PR #388, opened 2024-06-25
Write metadata using delayed that depends on writing array(s)

joshua-gould:
If writing a dask array and `compute = False`, the image metadata will be written even if some of the array chunks are not written successfully. This ensures that metadata is written only after all chunks are stored. Thanks.

### DIFF
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -7,6 +7,7 @@ import warnings
 from pathlib import Path
 from typing import Any, Dict, List, Optional, Tuple, Union
 
+import dask
 import dask.array as da
 import numpy as np
 import zarr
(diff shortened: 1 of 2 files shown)


joshmoore:
Thanks for the PR, @joshua-gould! Definitely agree with the goal of properly catching errors on write. :+1: There are two failures in the tests now, across a number of the jobs, related to no longer finding an image after the write, which sounds much like the metadata not being written. Did you see anything similar while preparing the change?

joshua-gould:
Sorry, I fixed the logic to invoke persist when using dask in test_write_image_dask

joshmoore:
(Launched the workflows)

joshmoore:
:+1:

joshmoore:
Released in [0.10.2](https://pypi.org/project/ome-zarr/0.10.2/). Thanks again, @joshua-gould!

PR #404, opened 2024-11-07
Zarr v3

will-moore:
Work in progress to update to Zarr v3.
This is currently focussed on getting existing Zarr v2 tests to pass...

### DIFF
--- a/ome_zarr/io.py
+++ b/ome_zarr/io.py
@@ -11,7 +11,7 @@ from urllib.parse import urljoin
 
 import dask.array as da
 import zarr
-from zarr.storage import RemoteStore
+from zarr.storage import RemoteStore, LocalStore, StoreLike
 
 from .format import CurrentFormat, Format, detect_format
 from .types import JSONDict
@@ -21,7 +21,7 @@ LOGGER = logging.getLogger("ome_zarr.io")
 
 class ZarrLocation:
     """
-    IO primitive for reading and writing Zarr data. Uses RemoteStore for all
+    IO primitive for reading and writing Zarr data. Uses a store for all
     data access.
 
     No assumptions about the existence of the given path string are made.
@@ -30,7 +30,7 @@ class ZarrLocation:
 
     def __init__(
         self,
-        path: Union[Path, str, RemoteStore],
+        path: StoreLike,
         mode: str = "r",
         fmt: Format = CurrentFormat(),
     ) -> None:
@@ -41,7 +41,7 @@ class ZarrLocation:
             self.__path = str(path.resolve())
         elif isinstance(path, str):
             self.__path = path
-        elif isinstance(path, RemoteStore):
+        elif isinstance(path, RemoteStore, LocalStore):
             self.__path = path.path
         else:
             raise TypeError(f"not expecting: {type(path)}")
@@ -52,7 +52,6 @@ class ZarrLocation:
         self.__store: RemoteStore = (
             path if isinstance(path, RemoteStore) else loader.init_store(self.__path, mode)
         )
-
         self.__init_metadata()
         detected = detect_format(self.__metadata, loader)
         LOGGER.debug("ZarrLocation.__init__ %s detected: %s", path, detected)
@@ -68,16 +67,18 @@ class ZarrLocation:
         """
         Load the Zarr metadata files for the given location.
         """
-        self.zarray: JSONDict = self.get_json(".zarray")
         self.zgroup: JSONDict = self.get_json(".zgroup")
+        self.zarray: JSONDict = {}
         self.__metadata: JSONDict = {}
         self.__exists: bool = True
         if self.zgroup:
-            self.__metadata = self.get_json(".zattrs")
-        elif self.zarray:
-            self.__metadata = self.get_json(".zattrs")
+            self.__metadata = self.zgroup
         else:
-            self.__exists = False
+            self.zarray: JSONDict = self.get_json(".zarray")
+            if self.zarray:
+                self.__metadata = self.zarray
+            else:
+                self.__exists = False
 
     def __repr__(self) -> str:
         """Print the path as well as whether this is a group or an array."""
@@ -155,14 +156,7 @@ class ZarrLocation:
         All other exceptions log at the ERROR level.
         """
         try:
-            store = zarr.storage.RemoteStore.from_url("https://uk1s3.embassy.ebi.ac.uk")
-            group = zarr.open_group(store=store, path="idr/zarr/v0.4/idr0062A/6001240.zarr")
-            print("Zarr group", group.attrs.asdict())
-
-            print("self.__path", self.__path)
-            print("subpath", subpath)
-            # data = self.__store.get(subpath)
-            group = zarr.open_group(store=self.__store, path="/")
+            group = zarr.open_group(store=self.__store, path="/", zarr_version=2)
             return group.attrs.asdict()
         except KeyError:
             LOGGER.debug("JSON not found: %s", subpath)
@@ -199,10 +193,11 @@ class ZarrLocation:
         Return whether the current underlying implementation
         points to a local file or not.
         """
-        return self.__store.fs.protocol == "file" or self.__store.fs.protocol == (
-            "file",
-            "local",
-        )
+        # return self.__store.fs.protocol == "file" or self.__store.fs.protocol == (
+        #     "file",
+        #     "local",
+        # )
+        return isinstance(self.__store, LocalStore)
 
     def _ishttp(self) -> bool:
         """
--- a/ome_zarr/io.py
+++ b/ome_zarr/io.py
@@ -10,7 +10,8 @@ from typing import List, Optional, Union
 from urllib.parse import urljoin
 
 import dask.array as da
-from zarr.storage import FSStore
+import zarr
+from zarr.storage import RemoteStore
 
 from .format import CurrentFormat, Format, detect_format
 from .types import JSONDict
@@ -20,7 +21,7 @@ LOGGER = logging.getLogger("ome_zarr.io")
 
 class ZarrLocation:
     """
-    IO primitive for reading and writing Zarr data. Uses FSStore for all
+    IO primitive for reading and writing Zarr data. Uses RemoteStore for all
     data access.
 
     No assumptions about the existence of the given path string are made.
@@ -29,7 +30,7 @@ class ZarrLocation:
 
     def __init__(
         self,
-        path: Union[Path, str, FSStore],
+        path: Union[Path, str, RemoteStore],
         mode: str = "r",
         fmt: Format = CurrentFormat(),
     ) -> None:
@@ -40,7 +41,7 @@ class ZarrLocation:
             self.__path = str(path.resolve())
         elif isinstance(path, str):
             self.__path = path
-        elif isinstance(path, FSStore):
+        elif isinstance(path, RemoteStore):
             self.__path = path.path
         else:
             raise TypeError(f"not expecting: {type(path)}")
@@ -48,8 +49,8 @@ class ZarrLocation:
         loader = fmt
         if loader is None:
             loader = CurrentFormat()
-        self.__store: FSStore = (
-            path if isinstance(path, FSStore) else loader.init_store(self.__path, mode)
+        self.__store: RemoteStore = (
+            path if isinstance(path, RemoteStore) else loader.init_store(self.__path, mode)
         )
 
         self.__init_metadata()
@@ -104,7 +105,7 @@ class ZarrLocation:
         return self.__path
 
     @property
-    def store(self) -> FSStore:
+    def store(self) -> RemoteStore:
         """Return the initialized store for this location"""
         assert self.__store is not None
         return self.__store
@@ -154,10 +155,15 @@ class ZarrLocation:
         All other exceptions log at the ERROR level.
         """
         try:
-            data = self.__store.get(subpath)
-            if not data:
-                return {}
-            return json.loads(data)
+            store = zarr.storage.RemoteStore.from_url("https://uk1s3.embassy.ebi.ac.uk")
+            group = zarr.open_group(store=store, path="idr/zarr/v0.4/idr0062A/6001240.zarr")
+            print("Zarr group", group.attrs.asdict())
+
+            print("self.__path", self.__path)
+            print("subpath", subpath)
+            # data = self.__store.get(subpath)
+            group = zarr.open_group(store=self.__store, path="/")
+            return group.attrs.asdict()
         except KeyError:
             LOGGER.debug("JSON not found: %s", subpath)
             return {}
--- a/ome_zarr/io.py
+++ b/ome_zarr/io.py
@@ -41,8 +41,10 @@ class ZarrLocation:
             self.__path = str(path.resolve())
         elif isinstance(path, str):
             self.__path = path
-        elif isinstance(path, RemoteStore, LocalStore):
+        elif isinstance(path, RemoteStore):
             self.__path = path.path
+        elif isinstance(path, LocalStore):
+            self.__path = str(path.root)
         else:
             raise TypeError(f"not expecting: {type(path)}")
 
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -190,7 +190,7 @@ def write_multiscale(
     :param pyramid:
         The image data to save. Largest level first. All image arrays MUST be up to
         5-dimensional with dimensions ordered (t, c, z, y, x)
-    :type group: :class:`zarr.hierarchy.Group`
+    :type group: :class:`zarr.Group`
     :param group: The group within the zarr store to store the data in
     :type chunks: int or tuple of ints, optional
     :param chunks:
@@ -265,7 +265,12 @@ Please use the 'storage_options' argument instead."""
                 dask_delayed.append(da_delayed)
 
         else:
-            group.create_dataset(str(path), data=data, chunks=chunks_opt, **options)
+            # v2 arguments
+            options["shape"] = data.shape
+            options["chunks"] = chunks_opt
+            options["dimension_separator"] = "/"
+
+            group.create_array(str(path), data=data, **options)
 
         datasets.append({"path": str(path)})
 
@@ -305,7 +310,7 @@ def write_multiscales_metadata(
     """
     Write the multiscales metadata in the group.
 
-    :type group: :class:`zarr.hierarchy.Group`
+    :type group: :class:`zarr.Group`
     :param group: The group within the zarr store to write the metadata in.
     :type datasets: list of dicts
     :param datasets:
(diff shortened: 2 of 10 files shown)


will-moore:
Zarr v3 is not supported on python 3.9 or 3.10. I'll remove them from the build...

Build on python 3.12 is failing tests with:
```
  = 106 failed, 282 passed, 3 skipped, 6 xfailed, 2270 warnings in 75.86s (0:01:15) =
```

joshmoore:
@will-moore: what's the next step here?

will-moore:
With that previous commit (even without a `FormatV05`) reading works, including via `napari-ome-zarr`:
NB: Labels not working for this example yet:
```
$ napari --plugin napari-ome-zarr https://uk1s3.embassy.ebi.ac.uk/idr/share/ome2024-ngff-challenge/0.0.5/6001240.zarr
```
Screenshot 2024-12-09 at 12 12 52

will-moore:
@joshmoore can you have a look at this now? See updated description. Thx.

ome_zarr/io.py
```diff
@@ -104,7 +132,7 @@ def path(self) -> str:
         return self.__path
 
     @property
-    def store(self) -> FSStore:
+    def store(self) -> RemoteStore:
```
joshmoore, line 135: Unsure why this isn't also `| LocalStore`

ome_zarr/io.py
```diff
@@ -67,16 +70,41 @@ def __init_metadata(self) -> None:
         """
         Load the Zarr metadata files for the given location.
         """
-        self.zarray: JSONDict = self.get_json(".zarray")
```
joshmoore, line 70: can `get_json()` now be deprecated?

joshmoore (commented):
Few general comments and one or two more pointed ones (especially around dimension separator). In general, though, I think this is looking good. Thanks, @will-moore.

joshmoore:
Updated with #346.

tests/data/v2/0/.zarray
```diff
@@ -13,6 +13,7 @@
         "id": "blosc",
         "shuffle": 1
     },
+    "dimension_separator": "/",
```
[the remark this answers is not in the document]
will-moore: How does zarr know to use the `/`  dimension_separator if it's missing? (how was this working before)?

will-moore:
I have tests passing locally with changes similar to https://github.com/dask/dask/pull/11660 and https://github.com/dask/dask/pull/11668

will-moore:
The first test failure `tests/test_cli.py` `def test_astronaut_download()`
is currently failing with:

```
    store = zarr.storage.FsspecStore.from_url(
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/wmoore/opt/anaconda3/envs/zarrv3_py312/lib/python3.12/site-packages/zarr/storage/_fsspec.py", line 182, in from_url
    return cls(fs=fs, path=path, read_only=read_only, allowed_exceptions=allowed_exceptions)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/wmoore/opt/anaconda3/envs/zarrv3_py312/lib/python3.12/site-packages/zarr/storage/_fsspec.py", line 96, in __init__
    raise TypeError("Filesystem needs to support async operations.")
```

This can be replicated with

```
from skimage import data
import dask.array as da
da.to_zarr(da.from_array(data.coins(), chunks=(64, 64)), "test_dask_to_zarr.zarr", compute=True, storage_options={"chunks": (64, 64)})
```
Or, digging a bit deeper, this gives the same error:

```
import zarr
store = zarr.storage.FsspecStore.from_url("test_dask_to_zarr.zarr", read_only=True, storage_options={"chunks": (64, 64)})
```

And we can test fsspec directly with:
```
from fsspec.core import url_to_fs
fs, path = url_to_fs("test_dask_to_zarr.zarr")
# False! Why?
print("fs.async_impl", fs.async_impl)
```

will-moore:
Another issue with dask and latest zarr causing test failures: Using `da.to_zarr()` with a store:

```
from skimage import data
import dask.array as da
import zarr

dask_data = da.from_array(data.coins(), chunks=(64, 64))

# v2 store - works dask==2023.5.0 zarr==2.18.2
# store = zarr.storage.FSStore("test_dask_to_zarr_v2.zarr")

# v3 store - fails dask==2025.1.0 zarr==3.0.1
store = zarr.storage.FsspecStore("test_dask_to_zarr_v3.zarr")
da.to_zarr(dask_data, store, compute=True, storage_options={"chunks": (64, 64)})
```
Fails with:
```
  File "/Users/wmoore/Desktop/python-scripts/zarr_scripts/test_dask_to_zarr.py", line 27, in <module>
    store = zarr.storage.FsspecStore("test_dask_to_zarr_v3.zarr")
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/wmoore/opt/anaconda3/envs/zarrv3_py312/lib/python3.12/site-packages/zarr/storage/_fsspec.py", line 95, in __init__
    if not self.fs.async_impl:
           ^^^^^^^^^^^^^^^^^^
AttributeError: 'str' object has no attribute 'async_impl'
```

will-moore:
Currently failing to make progress here due to the issues raised above, reported in more detail at https://github.com/dask/dask/issues/11691

will-moore:
Just tied running tests locally after updating to latest `zarr==3.0.5`, `dask==2025.2.0` and `fsspec==2025.3.0` and I get the same 27 test failures as in the last build above.

Most of them are from `da.to_zarr()` like this:

```
../ome_zarr/writer.py:794: in write_multiscale_labels
    dask_delayed_jobs = write_multiscale(
../ome_zarr/writer.py:262: in write_multiscale
    da_delayed = da.to_zarr(
../../../../opt/anaconda3/envs/zarrv3_py312/lib/python3.12/site-packages/dask/array/core.py:3939: in to_zarr
    store = zarr.storage.FsspecStore.from_url(
../../../../opt/anaconda3/envs/zarrv3_py312/lib/python3.12/site-packages/zarr/storage/_fsspec.py:176: in from_url
    fs, path = url_to_fs(url, **opts)
../../../../opt/anaconda3/envs/zarrv3_py312/lib/python3.12/site-packages/fsspec/core.py:403: in url_to_fs
    chain = _un_chain(url, kwargs)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _

path = LocalStore('file:///private/var/folders/l7/fy41yg5n67q_4ck550wbkrfh0000gn/T/pytest-of-wmoore/pytest-12/test_two_label_images_from_arr0/data.ome.zarr')
kwargs = {'asynchronous': True, 'chunks': ((1,), (2,), (1,), (256,), (256,))}

    def _un_chain(path, kwargs):
        # Avoid a circular import
        from fsspec.implementations.cached import CachingFileSystem
    
>       if "::" in path:
E       TypeError: argument of type 'LocalStore' is not iterable
```

joshmoore:
```
  >       if "::" in path:
  E       TypeError: argument of type 'LocalStore' is not iterable
```
sounds like it's going to be a matter of our assumptions about the API.

will-moore:
The major issue that was addressed in [ebea6a4](https://github.com/ome/ome-zarr-py/pull/404/commits/ebea6a4f7c61c4761e82cfcfd8dcc9380ce17eac) above was that if we call 

```
da.to_zarr(arr, url, storage_options={...})
```
this fails if url is `store` (rather than a string) AND `storage_options` is also specified.
With zarr v3, and setting `storage_options` the 'url` parameter is treated as a string, even though it's a store, hence the error with `if "::" in path:`.

See:
https://github.com/dask/dask/blob/9789fc69797309298a7d8d3705741709cb07a5f3/dask/array/core.py#L3936

Our tests can be made to pass by simply NOT passing in the `storage_options`. I guess we just don't have any tests that test that they are actually used!

Also, it's kinda strange that the only test that failed was `test_write_multiscale_labels()` as it was the only one that was calling `write_multiscale()` with dask data. The only other (1) test in `test_writer.py` that calls `write_multiscale()` is `test_default_compression()`.

will-moore:
It seems like `storage_options` gets passed to `url_to_fs(url, **kwargs)`
 https://github.com/fsspec/filesystem_spec/blob/961412dc83cb242728860f8637ab234485962f40/fsspec/core.py#L369 

which says kwargs are "Extra options that make sense to a particular storage connection, e.g. host, port, username, password, etc."
and there the `known_kwars = {"compression", "encoding", "errors", "expand", "mode", "name_function", "newline", "num"}` are ignored. 

So I'm not sure what `storage_options` we need to support in these tests or in ome-zarr-py?

JoOkuma:
Hi @will-moore and @joshmoore,

What's the state of this PR? Is there an expected timeline?

We are [planning to move to zarr 3](https://github.com/czbiohub-sf/iohub/pull/275) and we were hoping to be compatible with most packages in the ecosystem, as ome-zarr-py is a big part of it.

Is there any way we could help?

ome_zarr/writer.py
```diff
@@ -251,21 +257,40 @@ def write_multiscale(
...
+                options["chunks"] = data.chunks
             da_delayed = da.to_zarr(
                 arr=data,
                 url=group.store,
                 component=str(Path(group.path, str(path))),
-                storage_options=options,
-                compressor=options.get("compressor", zarr.storage.default_compressor),
-                dimension_separator=group._store._dimension_separator,
+                # IF we pass storage_options then dask NEEDS url to be a string
+                storage_options=None,
```
will-moore, line 267: This means that any `storage_options` passed in by the user are ignored (except for 'chunks')
and is the most likely blocker on this PR.

will-moore: Is there a way to get the URL from the `group.store`? Then instead of `url=group.store` above we could do `url=path/to/store` and this would allow us to pass in `storage_options`.
Or maybe there just aren't any other `storage_options` we need to support here and it's OK to ignore?

will-moore:
@JoOkuma Thanks for asking... I think that this is at least ready for review. Just need to look into the most recent test failure above...

NB: this PR simply replaces usage of zarr-python v2 with v3 but doesn't support *writing* of Zarr v3 data (as used in OME-Zarr v0.5) although it may support reading it.

If you have some tests or scripts that use ome-zarr-py, it would be great if you could try this branch and see if it's working for you.

psobolewskiPhD:
Hi! Also gently bumping this!
We've migrated our public image repository to use ome-zarr 0.5 but now those files can't be read!
I've pulled this down and with no other changes than installing this branch, I can now read those files via `napari-ome-zarr` plugin. If I checkout the main branch, then it fails.
I tested some IDR ome-zarr 0.4 images (e.g https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0062A/6001240.zarr )and they are unaffected.

Only thing I notice is some errors if you quit napari without removing the added layers:
```
15:38:30 : ERROR : MainThread : Unclosed client session
client_session: <aiohttp.client.ClientSession object at 0x155fdd7c0>
15:38:30 : ERROR : MainThread : Unclosed connector
connections: ['deque([(<aiohttp.client_proto.ResponseHandler object at 0x32a9592b0>, 257318.807015), (<aiohttp.client_proto.ResponseHandler object at 0x3116abad0>, 257319.097734083), (<aiohttp.client_proto.ResponseHandler object at 0x32a1d47d0>, 257319.101275708), (<aiohttp.client_proto.ResponseHandler object at 0x32a95a9f0>, 257319.183711166), (<aiohttp.client_proto.ResponseHandler object at 0x16c16d8b0>, 257319.272412166), (<aiohttp.client_proto.ResponseHandler object at 0x32a1d7650>, 257319.332259666), (<aiohttp.client_proto.ResponseHandler object at 0x32a95ae70>, 257319.525722666), (<aiohttp.client_proto.ResponseHandler object at 0x32a958890>, 257319.585263833), (<aiohttp.client_proto.ResponseHandler object at 0x32a958bf0>, 257319.697598916)])']
connector: <aiohttp.connector.TCPConnector object at 0x16bd3d940>
```
If you delete the layers then close napari, then I don't get those.

ziw-liu:
> @JoOkuma Thanks for asking... I think that this is at least ready for review. Just need to look into the most recent test failure above...
> 
> NB: this PR simply replaces usage of zarr-python v2 with v3 but doesn't support _writing_ of Zarr v3 data (as used in OME-Zarr v0.5) although it may support reading it.
> 
> If you have some tests or scripts that use ome-zarr-py, it would be great if you could try this branch and see if it's working for you.

Our [round-trip tests](https://github.com/czbiohub-sf/iohub/pull/297) for OME-Zarr v0.4 seems to be working again with this branch!

jni:
@will-moore @joshmoore reading #452, I understand the motivation, and I support it, but is there a compelling reason not to merge and release this (seems like a bunch of folks can already use it), and then work on the Great Reunification after?

ome_zarr/io.py
will-moore: Marked as deprecated now

ome_zarr/io.py
```diff
@@ -40,18 +40,21 @@ def __init__(
...
 
         loader = fmt
         if loader is None:
             loader = CurrentFormat()
-        self.__store: FSStore = (
-            path if isinstance(path, FSStore) else loader.init_store(self.__path, mode)
+        self.__store: RemoteStore = (
+            path
+            if isinstance(path, RemoteStore)
+            else loader.init_store(self.__path, mode)
```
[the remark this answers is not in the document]
will-moore: Done

ome_zarr/writer.py
will-moore: If the storage_options only specify arguments for *creating* a store, then they would already be ignored if you're also passing in an existing store.
If you're passing in an existing store, created with various arguments, then you wouldn't want to replace that with a url as I suggested on `Apr 7` above.

joshmoore:
:+1: @will-moore but I assume the writing of 0.5 data is still an issue, right?

joshmoore:
Discussing with @will-moore, the plan will be to get this PR and #413 merged into 0.12rc1. More testing from the community (*hint, hint* @psobolewskiPhD @jni) would be very welcome.

**WARNING:** You won't be able to write without specifying a `format <= 0.5` since `V05` is not yet implemented. (If you want to write V04 by default, use 0.11.x)

joshmoore:
More apologies from my side. What you said matches what I understood, just that #413 went *beyond* what we had discussed :+1:

will-moore:
@joshmoore: With those changes, you now are forced to specify `fmt=FormatV04()` with `parse_url()` if `mode="w"`.
E.g. like this...

```
import numpy as np
import zarr
from ome_zarr.format import FormatV04

data = np.random.default_rng(0).poisson(lam=10, size=(10, 256, 256)).astype(np.uint8)
store = parse_url("image.zarr", mode="w", fmt=FormatV04()).store
root = zarr.group(store=store)
write_image(image=data, group=root, axes="zyx", storage_options=dict(chunks=(1, 64, 64)))
```

That code works, but unfortunately we end up with data that looks like `v0.4` apart from it contains `"version": "0.5",` because `write_image()` defaults to using `CurrentFormat()`. The `FormatV05` generates exactly the same output as `FormatV04` except for the version.

After testing a matrix of using different versions for `parse_url()` and for `write_image()` with *this* branch, it seems that the format passed to `parse_url()` has no effect.

Testing the same matrix of versions with *released* ome-zarr-py produces all valid output **except** when you use `FormatV01` in `parse_url()`: in this case all the data is written with `dimension_separator: "."` so the `v0.2, v0.3, v0.4` data is invalid (when you use a **later** version to write the data).

To summarise, we probably ALSO want to prevent usage of `write_image()`, `write_multiscales()` without specifying `fmt` AND ensure that uses use `FormatV04` or earlier. Do we change the method signature so that `fmt` is no-longer optional?And `raise NotImplementedError` if the fmt is not `FormatV04` or earlier?

So users must then do:

```
store = parse_url("image.zarr", mode="w", fmt=FormatV04()).store
root = zarr.group(store=store)
write_image(image=data, group=root, axes="zyx", fmt=FormatV04(), storage_options=dict(chunks=(1, 64, 64)))
```

Does this look acceptable?

imagejan:
Sorry for chiming in after only superficially following; my opinion might be naive, but:

> ```
> store = parse_url("image.zarr", mode="w", fmt=FormatV04()).store
> root = zarr.group(store=store)
> write_image(image=data, group=root, axes="zyx", fmt=FormatV04(), storage_options=dict(chunks=(1, 64, 64)))
> ```

For me, this screams for an object-oriented API where I can get some `omezarr= OMEZarrFormat(version="0.4")` that I then can ask to `omezarr.parse_url(...)` and `omezarr.write_image(...)`, doesn't it?

will-moore:
With zarr v3.0.8, and this current PR, I was testing the docs examples, e.g. for adding `omero` metadata at https://ome-zarr.readthedocs.io/en/stable/python.html#rendering-settings

Actually, that fails in zarr v2 - you can't create a group where one exists already. Should have tested that better!

However, in Zarr v3 with this branch, when you do this:
```
store = parse_url(path, mode="w")
```
it completely replaces EVERYTHING at `path` with an empty zarr group!
This is pretty scary!

I'll fix the docs for the released version, but I wonder if we should add checks in to `parse_url()` to prevent that?

will-moore:
This is what we're doing under `parse_url(path, mode="w")`...

```
store = LocalStore(path, read_only=False)
# this wipes the store
group = zarr.open_group(store=store, path="/", mode="w", zarr_format=2)
```

will-moore:
I was looking at the next step of support for writing OME-Zarr v0.5 (zarr v3) over at https://github.com/ome/ome-zarr-py/pull/413#issuecomment-2949764985 and ran into issues with mixing of zarr v2 and zarr v3.

To summarise, when writing OME-Zarr, I think it will work better to specify the version when you create the store with `parse_url()` and then use the zarr_format of the store (zarr v2 or zarr v3) to inform the version of OME-Zarr that is written with `write_image()` etc.
This should allow you to specify the OME-Zarr version in just *one* place and also avoid mixing zarr v2 and zarr v3.
This will look like the first code sample in https://github.com/ome/ome-zarr-py/pull/404#issuecomment-2909287415 above.

So I'm going to go ahead and give that a try. Apologies for the back and forth on this....

joshmoore:
Thanks all to the testing here! Since this has lingered a bit, #413 (as a superset of this PR) is now almost ready for testing with write support. I'm going to close this PR so we can all focus on #413. Plan will be to get an RC out ASAP. :tada:

PR #410, opened 2024-11-26
pin zarr at < 3

will-moore:
Trying to fix build at https://github.com/ome/ome-zarr-py/actions/runs/12038548002 which failed because Zarr v3 somehow got installed.


joshmoore:
Thanks, @will-moore. It might even make sense to get this out ASAP.

joshmoore:
Released as 0.10.0

Issue #231, opened 2022-10-19
ome_zarr download uses wrong dimension_separator


will-moore:
It seems that `ome_zarr download` is writing chunks with `"dimension_separator": "."`, even when downloading v0.4 data:
```
$ ome_zarr download https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0138A/TimEmbryos-102219/HybCycle_0/MMStack_Pos0.ome.zarr/ --output MMStack_Pos0.ome.zarr


$ ls -al MMStack_Pos0.ome.zarr/0/
total 250256
drwxr-xr-x  387 wmoore  staff   12384 Oct 19 14:09 .
drwxr-xr-x   10 wmoore  staff     320 Oct 19 14:09 ..
-rw-r--r--    1 wmoore  staff     385 Oct 19 14:09 .zarray
-rw-r--r--    1 wmoore  staff  301361 Oct 19 14:09 0.0.0.0
-rw-r--r--    1 wmoore  staff  313241 Oct 19 14:09 0.0.0.1
-rw-r--r--    1 wmoore  staff  338394 Oct 19 14:09 0.0.0.2
...
```

joshmoore:
Uff! That seems like quite the bug.

joshmoore:
Still happening with origin/master and zarr_v3 (in #404)

PR #419, opened 2024-12-28
Fix dimension separator when downloading files

dstansby:
Fixes https://github.com/ome/ome-zarr-py/issues/231

I saw there's a fix in https://github.com/ome/ome-zarr-py/pull/404 too, but I thought a self-contained fix would be easier/quicker to ship. This PR also includes a test, which I don't think https://github.com/ome/ome-zarr-py/pull/404 has.


will-moore (approved):
Thanks!

Testing with:

```
ome_zarr download https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0101A/13457539.zarr
napari --plugin napari-ome-zarr ./13457539.zarr
```
this fixes the failure reported at https://github.com/thewtex/ngff-zarr/issues/125#issuecomment-2560277033

And the data is valid as reported by the validator with
```
ome_zarr view ./13457539.zarr
```

joshmoore:
Thanks all. Releasing as 0.10.3.

PR #422, opened 2025-01-25
Enable Python 3.10+ in pre-commit hooks

DimitriPapadopoulos:
Consistent with:
https://github.com/ome/ome-zarr-py/blob/14f407f15205025e45d8288eecf1ca07a0c9e030/pyproject.toml#L17


joshmoore:
Thanks, @DimitriPapadopoulos. Going through and merging now with @will-moore

PR #425, opened 2025-01-25
flake8, isort → ruff

DimitriPapadopoulos:
I don't think ruff has an equivalent to [flake8-rst-docstrings](https://github.com/peterjc/flake8-rst-docstrings).

Either keep it, or apply ruff rules releated to docs:
* [pydocstyle (D)](https://docs.astral.sh/ruff/rules/#pydocstyle-d)
* [pydoclint (DOC)](https://docs.astral.sh/ruff/rules/#pydoclint-doc)

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -43,8 +43,7 @@ repos:
     rev: v1.35.1
     hooks:
       - id: yamllint
-        # args: [--config-data=relaxed]
-        #
+        args: [--config-data=relaxed]
 
   - repo: https://github.com/codespell-project/codespell
     rev: v2.3.0
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,23 +1,6 @@
 ---
 repos:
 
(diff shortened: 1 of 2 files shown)


.pre-commit-config.yaml
```diff
@@ -44,27 +21,17 @@ repos:
...
-          "--rst-roles=class,func,ref,mod,meth,const",
-        ]
+    - id: ruff
+      args: ["--fix", "--show-fixes"]
+
+  - repo: https://github.com/psf/black
+    rev: 24.10.0
+    hooks:
+      - id: black
+        args: [--target-version=py36]
```
joshmoore, line 34: should this not be 3.10 now?

DimitriPapadopoulos: From the [`--target-version`](https://black.readthedocs.io/en/stable/usage_and_configuration/the_basics.html) documentation:
> By default, Black will infer target versions from the project metadata in `pyproject.toml`, specifically the `[project.requires-python]` field. If this does not yield conclusive results, Black will use per-file auto-detection.

I think this should be removed altogether...

PR #438, opened 2025-02-23
Simplify _create_mip()

dstansby:
This improves the implementation of `_create_mip()` by:

- Removing the `axes` return. This isn't needed, because `axes` is never modified within the function. This simplifies the return type.
- Add an `AxesType` alias that can be used across `writer.py` to prevent having to redefine the type in lots of function signatures
- Add a docstring to `_create_mip()`

### DIFF
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -1,6 +1,4 @@
-"""Image writer utility
-
-"""
+"""Image writer utility"""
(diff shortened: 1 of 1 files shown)


joshmoore:
Green after resolving conflicts and updating some minor ruff issues.

Thanks, @dstansby

PR #441, opened 2025-03-10
Remove distributed from dependencies (unused)

m-albert:
Hey, thanks for your work here!

I noticed that although `distributed` is listed in the dependencies, as far as I can tell the package is not being used in `ome-zarr-py`. So this PR removes it.

Next to cleaning up the dependencies, this can be useful for using `ome-zarr-py` within WASM-based environments (e.g. jupyter lite), which `distributed` is not compatible with.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,7 +19,6 @@ requires-python = ">3.10"
 dependencies = [
     "numpy",
     "dask",
-    "distributed",
     "zarr>=2.8.1,<3",
     "fsspec[s3]>=0.8,!=2021.07.0,!=2023.9.0",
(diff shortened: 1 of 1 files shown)


joshmoore:
Launched and now green. @will-moore: any thoughts?

will-moore:
The dependency was added in https://github.com/ome/ome-zarr-py/commit/fbfd23b47a7beed308402dd104677f7b7d3d3073#diff-60f61ab7a8d1910d86d9fda2261620314edcae5894d5aaa236b821c7256badd7 but I don't see it being used any more. 👍

PR #450, opened 2025-04-22
Changelog 0.11.1

will-moore:
No change in released code, but I think the bump (and new tag) is still needed to trigger publishing to Pypi?


will-moore:
@joshmoore Any objection to merging this and releasing 0.11.1 ?

joshmoore:
Not at all.

PR #454, opened 2025-05-23
Change deprecated Gaussian/Laplacian pyramid args

AlbertDominguez:
Hi,

Thanks for the great effort you're doing here :) This PR updates the arguments passed to `skimage.transform.pyramid_gaussian` and `skimage.transform.pyramid_laplacian` (without any behaviour change) used by the `Scaler` class, as the arg `multichannel` was deprecated in `scikit-image 0.19.x`. Using the latest version of `scikit-image` (`0.25.x`) causes a crash if either of these methods are used for downsampling.

Note that this causes the code not to work if `scikit-image<0.19`. Unsure if it's worth pinning in the deps 🙂 

Here's a simple code to replicate:
```python
from ome_zarr.scale import Scaler
import numpy as np
import zarr

if __name__ == "__main__":
    # Example usage
    store = parse_url("dbg.ome.zarr", "w").store
    group = zarr.group(store=store, overwrite=True)
    array = np.random.randn(1,256,256)

    scaler = Scaler(
        downscale=2,
        method="gaussian", # or "laplacian"
        max_layer=2,
    )
   
    # Next line currently breaks due to deprecated args being used downstream by the Scaler instance
    write_image(
        image=array,
        group=group,
        scaler=scaler,
        axes="cyx",
    )
```

### DIFF
--- a/tests/test_scaler.py
+++ b/tests/test_scaler.py
@@ -1,11 +1,12 @@
+import pathlib
+
 import dask.array as da
 import numpy as np
-import pathlib
 import pytest
 import zarr
 
-from ome_zarr.scale import Scaler
 from ome_zarr.io import parse_url
+from ome_zarr.scale import Scaler
 from ome_zarr.writer import write_image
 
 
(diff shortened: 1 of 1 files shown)


AlbertDominguez:
Hi @joshmoore, thanks! I've just merged the latest changes into the PR + pinned `scikit-image`. Also further tested locally and everything seems to work as expected.

joshmoore:
Leaving this for @will-moore to look at, but I imagine we can roll it into the imminent 0.12.0.

PR #457, opened 2025-06-16
Visualization for the ome-zarr-py

ivanmilevtues:
This PR contains high-level diagrams for the ome-zarr-py codebase. You can see how they render in Github here:
https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/ome-zarr-py/on_boarding.md

These diagrams are designed to help people quickly understand a codebase. There are many scientists who engage with this repository, our goal is to reduce the time it takes them to get up to speed—from hours to just minutes—by using visuals (diagrams) instead of reading through the entire code. I would love to hear your opinion on diagram first documentation.

Any feedback is more than welcome! Further to reduce the burdon of another documentation that has to be updated, we have a free github action which can automatically update the docs.

Full disclosure: we're trying to turn this into a startup, but we're still in a very early stage and figuring out what will actually be useful for people.

### DIFF
--- a/.codeboarding/CLI & General Utilities.md
+++ b/.codeboarding/CLI & General Utilities.md
@@ -1,0 +1,209 @@
+```mermaid
+
+graph LR
+
+    CLI_Handler["CLI Handler"]
+
+    Zarr_Utility_Operations["Zarr Utility Operations"]
+
+    Zarr_Data_Creation["Zarr Data Creation"]
+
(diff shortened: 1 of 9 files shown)


ivanmilevtues:
Hey a quick update, CodeBoarding is fully open-source now. You can check exactly how the generation works here: https://github.com/CodeBoarding/CodeBoarding!

PR #472, opened 2025-08-18
Parse url raise exception

### LINKED ISSUE #471: Could be useful if `io.parse_url` raises an Exception instead of returning `None`
Hi, trying out this library for the first time and following the [tutorial](https://ome-zarr.readthedocs.io/en/stable/python.html#reading-ome-ngff-images) to read an existing ome zarr file. Took me a while to understand why `parse_url()` was just returning `None`. The [docstring](https://ome-zarr.readthedocs.io/en/stable/api/io.html#ome_zarr.io.parse_url) states that it returns `None` if there is an error, but would it instead be preferable to raise an exception to help the user? Eventually by looking through the various [format classes](https://ome-zarr.readthedocs.io/en/stable/api/format.html) and the zarr json I figured out that I was using an OME file that was v0.5 and the latest non-pre-release only supports upto v0.4, but then I tried the pre-release which supports v0.5 😄 . But that was just my first hunch as to why it wasn't working.

Anyways, sorry if I'm being annoying here but I just thought it'd be more useful to raise when something is wrong rather than just return `None`.

will-moore:
Fixes #471. 

TODO: fix lots of tests!

We should also consider handling the scenario described there, where user expects one OME-Zarr format but gets a different one.

See #468 where we remove handling of mismatching ome-zarr formats, because we don't want to re-create a store.

probably `parse_url()` with No format specified should be happy to read any format, but IF you specify a format, it MUST match.
This would be similar to the strategy used for `write_image()` etc where the default format is `None` (format will be chosen based on zarr group) but if you specify a format that is incompatible, you get an exception.
 
e.g.

```
    def test_wrong_format(self):
        """ Write a v0.4 zarr, read it as CurrentFormat"""
        grp = create_zarr(str(self.path), fmt=FormatV04())
        # should use correct format
        v2reader = Reader(parse_url(str(self.path)))
        # should raise exception
        reader = Reader(parse_url(str(self.path), fmt=CurrentFormat()))
        assert len(list(reader())) == 0
```

### DIFF
--- a/tests/test_reader.py
+++ b/tests/test_reader.py
@@ -100,6 +100,12 @@ class TestInvalid:
             assert len(list(reader())) == 2
         assert str(exe.value) == "Version invalid not recognized"
 
+    def test_not_found(self):
+        with pytest.raises(FileNotFoundError) as exe:
+            reader = Reader(parse_url(str(self.path / "not_found")))
+            assert len(list(reader())) == 0
+        assert "Unable to find group" in str(exe.value)
+
 
 class TestHCSReader:
     @pytest.fixture(autouse=True)


will-moore:
Closing in favour of #476

PR #474, opened 2025-08-18
read_only if mode is 'r'

### LINKED ISSUE #473: NGFF 0.5: `LocalStore` initialized with `read_only=True` for `mode in ['r+', 'a']`
With the PR for Zarr v3 (NGFF 0.5) support https://github.com/ome/ome-zarr-py/pull/413, the following change has been introduced in `format.py`:

Image

This leads to stores opened with `r+` and `a`, which should be writeable, to be initialized as `read_only`. The expected behavior is to be able to write on those stores, as [shown in the Zarr docs](https://zarr.readthedocs.io/en/stable/api/zarr/storage/index.html#zarr.storage.StorePath.open).

will-moore:
Fixes #473.

### DIFF
--- a/ome_zarr/format.py
+++ b/ome_zarr/format.py
@@ -159,15 +159,16 @@ class FormatV01(Format):
         "dimension_separator" is specified at array creation time
         """
 
+        read_only = mode == "r"
         if path.startswith(("http", "s3")):
             store = FsspecStore.from_url(
                 path,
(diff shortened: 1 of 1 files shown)


melonora (approved):
Thanks for the quick fix! This will allow us to move forward with zarr v3 support in SpatialData

Issue #445, opened 2025-04-04
Store can lose storage_options in version detection logic


berombau:
While exploring stores with various storage_options in a downstream project https://github.com/scverse/spatialdata/pull/907, @ap-- found that these storage_options can be dropped when using `.init_store(self.__path, mode)`. 

A fix would be to make sure that use of .init_store still uses `FSStore` (which was added in #349) instead of just a `str` or `Path`.

https://github.com/ome/ome-zarr-py/blob/7d1ae35c9795e666f06508971f64b0f209d15bff/ome_zarr/io.py#L62

will-moore:
Thanks for the feedback.
Looking at the codes, I see this is happening when the format (OME-Zarr version) is different from the supplied (or current) format. This means that we want to create a new store that corresponds to the detected format.
However, I'm not sure if this is really necessary since if we are reading existing data then we don't care so much about setting the dimension_separator etc that happens in store init.

We'll give it some thought...

Issue #376, opened 2024-05-07
Cannot overwrite Zarr opened with `mode= "a"`


chourroutm:
To my understanding, `zarr` supports rewriting with the  `mode= "a"`. However, it is not supported by `ome-zarr-py`.

# Example code
```python
vol = np.zeros((512,512,512))
store = ome_zarr.io.parse_url("all_zeros.ome.zarr",mode="a").store
root = zarr.group(store=store)
ome_zarr.writer.write_image(image=vol, group=root, axes="zyx",chunks=128)

store2 = ome_zarr.io.parse_url("all_zeros.ome.zarr",mode="a").store
root2 = zarr.group(store=store2)
ome_zarr.writer.write_image(image=vol, group=root2, axes="zyx",chunks=128)
```

# What I expect
To be able to overwrite the file `all_zeros.ome.zarr` when I specify `mode= "a"`.

# What I get

> ContainsArrayError                        Traceback (most recent call last)
> Cell In[10], line 8
>       6 store2 = ome_zarr.io.parse_url(\"all_zeros.ome.zarr\",mode=\"a\").store
>       7 root2 = zarr.group(store=store2)
> ----> 8 ome_zarr.writer.write_image(image=vol, group=root2, axes=\"zyx\",chunks=128)
> 
> [...]
> 
> File .venv\\Lib\\site-packages\\zarr\\storage.py:523, in _init_array_metadata(store, shape, chunks, dtype, compressor, fill_value, order, overwrite, path, chunk_store, filters, object_codec, dimension_separator, storage_transformers)
>     521 if not overwrite:
>     522     if contains_array(store, path):
> --> 523         raise ContainsArrayError(path)
>     524     elif contains_group(store, path, explicit_only=False):
>     525         raise ContainsGroupError(path)
> 
> ContainsArrayError: path '0' contains an array"

(Error log: [full_error.txt](https://github.com/ome/ome-zarr-py/files/15237148/full_error.txt))

dstansby:
I can reproduce; here's the full traceback:

```python
Traceback (most recent call last):
  File "/Users/dstansby/software/zarr/ome-zarr-py/test.py", line 12, in <module>
    ome_zarr.writer.write_image(image=vol, group=root2, axes="zyx",chunks=128)
  File "/Users/dstansby/software/zarr/ome-zarr-py/ome_zarr/writer.py", line 516, in write_image
    dask_delayed_jobs = write_multiscale(
[...]

PR #476, opened 2025-08-19
parse_url_deprecation

will-moore:
As discussed at https://github.com/ome/ome-zarr-py/pull/474#issuecomment-3199592248, we have various issues with `parse_url()` as it incompletely wraps `zarr.open_group()`.
See #376, #445, #471, #498. 

This PR updates docs and tests to avoid using `parse_url()`.
Instead of multiple lines for writing, we can now just do:

```
write_image(data, path_or_group, axes="zyx")
```

NB: this is not a breaking change as the `group` argument can be a path `(str)` or a group.

For reading, this is currently less concise than with `parse_url()` since we don't create an itterable `Reader()` - see discussion below...

### DIFF
--- a/docs/source/python.rst
+++ b/docs/source/python.rst
@@ -13,16 +13,14 @@ of 2 in the X and Y dimensions.
 Alternatively, the :py:func:`ome_zarr.writer.write_multiscale` can be used, which takes a
 "pyramid" of pre-computed `numpy` arrays.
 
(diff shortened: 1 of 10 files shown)


jo-mueller (approved):

joshmoore (approved):
Sorry, for the wait. :+1: for moving this forward.

PR #480, opened 2025-08-21
don't pass storage options when writing images

melonora:
This PR removes passing on `storage_options` to `da.to_zarr` when writing `dask_image` or `multiscale`.

In `SpatialData` we currently cannot write either `images` or `labels` to remote. The reason is that `storage_options` get passed despite an actual store being passed in `da.to_zarr`. Storage options only make sense to be passed on if we were to pass a `str` as url. In case we already have a `store`, `storage_options` does nothing, except cause an error when writing to remote as for example `chunks` is an invalid keyword argument when creating a store. The latter is done by dask when you actually pass on `storage_options`.

### DIFF
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -302,8 +302,6 @@ Please use the 'storage_options' argument instead."""
                 arr=data,
                 url=group.store,
                 component=str(Path(group.path, str(path))),
-                # IF we pass storage_options then dask NEEDS url to be a string
-                storage_options=None,
                 compute=compute,
                 zarr_format=zarr_format,
                 **options,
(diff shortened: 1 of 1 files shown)


ome_zarr/writer.py
```diff
@@ -302,8 +302,6 @@ def write_multiscale(
                 arr=data,
                 url=group.store,
                 component=str(Path(group.path, str(path))),
-                # IF we pass storage_options then dask NEEDS url to be a string
-                storage_options=None,
```
will-moore, line 306: Does this change behaviour, if we're already passing in `None`, or is it just to clean-up the code?

melonora: This is to clean up code. In both cases `storage_options` should not be passed because we already pass on a `store`. In `_write_dask_image` we still passed on `storage_options`.

PR #491, opened 2025-09-06
Make fmt.version extensions backward compatible in ome-zarr-py dependents.

melonora:
In SpatialData we use `RasterFormat` versions for writing images and labels. In discussion with @joshmoore and @bogovicj the version for `RasterFormatV02` is `0.4-dev-spatialdata` ([PR here](https://github.com/scverse/spatialdata/pull/849)). This leads to problems with newer ome-zarr-py versions that support zarr v3. Reason for this is checks in `writer.py` that check things like the following:
`fmt.version not in ("0.1", "0.2", "0.3", "0.4"):`, which leads `ome` metadata to be written which is specifically for Formats supporting zarr v3. This PR uses `.startswith` to check for the version in the string. 

In SpatialData I already checked whether this re-enables compatibility with zarr, which it does.

Note: In SpatialData so far all formats were based on FormatV04. Therefore, there are no issues with the other formats. Changing from FormatV04 to FormatV05 gives the problem at the moment and then only when it was using RasterFormatV02, given that the version here is not `0.4`, but `0.4-dev-spatialdata`. RasterFormatV01 does not have this issue as the version here was specificied as `0.4.`

Alternatives

- Follow Wills suggestion `if "ome" in attrs: attrs = attrs.get("ome", {})`. This is not explicit regarding versions.
- Make an exception only for `0.4` and "0.5" (later when there is a "0.6) which to the best of my knowledge is the only case in which an alternative version name was accepted that previously worked. Comment the reason for this exception and that this will not be supported for other versions. Reason for `0.5` as well is that we do not expect to be entirely compliant yet.
- Another option is to use a `Global` if present. Then we can handle the matching on our side.


melonora:
cc @LucaMarconato

melonora:
Thanks for approving! Would there be any timeline for a release? Right now zarr v3 in SpatialData is blocked by this as we can't have backward compatibility without this.

will-moore:
Just to understand your usage a bit better... Your `RasterFormatV02` is a subclass of `FormatV04`? 

I'm wondering if doing `if fmt.version.zarr_format == 2` might be a nicer way of checking for versions `0.1-0.4` in the future?

melonora:
Not all Format classes are based on current format. `RasterFormatV02`  is inheriting from `RasterFormatV01` which is based on ome-zarr-py `FormatV04`. This does not contain the `ome` key in attrs. The main reason for supporting this raster format is for allowing users to read the data that they have in zarr v2 and convert it to zarr v3 by then writing it back to disk with the current format. All our current formats would be based on `FormatV05`, but can contain additional metadata besides ome.

TLDR We need this to allow our users to update their zarr v2 data to zarr v3.

melonora:
I will close this PR, after talking yesterday together with @joshmoore , @jo-mueller and @LucaMarconato to discuss the potential implications and concerns of @sbesson, we have chosen for the option of overwriting the `version` metadata with the version as indicated by `SpatialData`. We know which `Format` is used with which element format so we can do the mapping on our side.

Prior to reading we then also check the metadata for the version and then request the corresponding ome-zarr version when calling `ZarrLocation`. 

This leaves still the issue that there should be a way to deviate in case of ome development versions and that there should be a way to indicate something like:
1) which format is used for reading the data
2) if the version string is an exact match, we expect that the data gets validated by NGFFvalidator, if only partial match then not.

However, this is a RFC process.

PR #504, opened 2025-10-30
Expose __version__ and ignore generated _version.py

Nirkan:
- This allows users to query ome_zarr version using ome_zarr.\_\_version\_\_ similar to zarr & other packages.
- The generated file during the editable install named _version.py added to .gitignore to avoid accidental commit.


Nirkan:
Thank you for the comments.
Yes, probably because it is not able to find _version.py file.
Will try to add a fallback part to resolve the docs build error.

will-moore (approved):
Looks good:

```
pip install -e .
```

```
>>> import ome_zarr
>>> ome_zarr.__version__
'0.12.3.dev67+g11fbcc26a'
```

jo-mueller (approved):

PR #510, opened 2025-12-15
Support for sharding

melonora:
With dask 2025.12.0, I added support for sharding in the Dask API. This PR aims to add full support for sharding in ome-zarr-py.

Note, the aim will be to support the zarr API more directly than currently is the case, e.g. storage options in the zarr API is not used for chunks. Also, parameters like `separator` are dependent not on the version the format is stored in, but the version of zarr installed. This means that since we only support `zarr` >v3.0.0, we must use the parameter `chunk_key_encoding` to define `separator`. Currently, I am first ensuring that all tests pass, before refactoring to the base implementation of direct support for the sharding API as it is in `zarr`.

@joshmoore @will-moore

### DIFF
--- a/ome_zarr/data.py
+++ b/ome_zarr/data.py
@@ -169,6 +169,7 @@ def create_zarr(
             "rdefs": {"model": "greyscale"},
         }
     else:
+        # TODO: is it valid to just assume RGB / 8bit?
         image_data = {
             "channels": [
                 {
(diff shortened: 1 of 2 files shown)


melonora:
Thank you! There will be a slight adjustment again with the next dask version as ultimately the `zarr_read_kwargs` only had to include `mode`. This was not clear due to documentation. I have opened a PR for this and hope this will be merged for the next dask release:

https://github.com/dask/dask/pull/12205

Issue #505, opened 2025-10-30
accessing nodes from the reader causes a lot of recursion but only on github CI


kushalkolar:
Hi! We're using [one of the multi-channel zarr examples](https://ome-zarr.readthedocs.io/en/stable/python.html#reading-ome-ngff-images) in the CI for [fastplotlib](https://github.com/fastplotlib/fastplotlib) as both verification for multi-channel images and as an [example in the docs](https://www.fastplotlib.org/ver/dev/_gallery/image_volume/image_volume_multi_channel.html#sphx-glr-gallery-image-volume-image-volume-multi-channel-py). However today this stopped working but only on github CI, there's a recursion depth error if you do `list(reader())`. I also tried `next(reader())` but get the same thing:

```python
examples/image_volume/image_volume_multi_channel.py:22: in <module>
    image_node = next(reader())
                 ^^^^^^^^^^^^^^
/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/site-packages/ome_zarr/reader.py:574: in __call__
    node = Node(self.zarr, self)
           ^^^^^^^^^^^^^^^^^^^^^
/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/site-packages/ome_zarr/reader.py:53: in __init__
    self.specs.append(Multiscales(self))
                      ^^^^^^^^^^^^^^^^^
/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/site-packages/ome_zarr/reader.py:316: in __init__
    node.add(child_zarr, visibility=False)
/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/site-packages/ome_zarr/reader.py:139: in add
    node = Node(zarr, self, visibility=visibility, plate_labels=plate_labels)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/site-packages/ome_zarr/reader.py:49: in __init__
    self.specs.append(Labels(self))
                      ^^^^^^^^^^^^
...

/opt/hostedtoolcache/Python/3.12.11/x64/lib/python3.12/site-packages/ome_zarr/reader.py:51: in __init__
    self.specs.append(Label(self))
                      ^^^^^^^^^^^
[...]

PR #511, opened 2025-12-27
fix recursion error if __store.fs.protocol is a tuple

will-moore:
Investigating #505, the recursion was caused by the `self.seen` list of nodes not recognising previously visited nodes because `ishttp()` was returning False, so a that `subpath()` wasn't using `urljoin(url, subpath)` but was instead using `f"{self.__path}/{subpath}"`

This means that if subpath was `../` to traverse up the hierarchy, this was simply concatenated, leading to new paths which weren't recognised as having previously been visited, hence the recursion error and paths like:

```
https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.3/idr0079A/9836998.zarr/labels/0/../../labels/0/../../labels/0/../../labels/0/../../labels/0/.
```

Also, with the release of `dask 2025.12.0` this seems to be a breaking change, so we either need to pin before that or require it and update.
Since https://github.com/dask/dask/pull/12153 adds support for zarr sharding, we want to updates so that is what this PR does.

### DIFF
--- a/tests/test_cli.py
+++ b/tests/test_cli.py
@@ -85,7 +85,7 @@ class TestCli:
         main(["info", f"{out}/{basename}"])
 
         if fmt is not None and fmt.zarr_format == 2:
-            assert directory_items(Path(out) / "data-3") == [
+            assert directory_items(Path(out) / basename) == [
                 Path(".zattrs"),
                 Path(".zgroup"),
                 Path("0"),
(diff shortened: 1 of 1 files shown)


dstansby:
Locally, if you use [`tox -r`](https://tox.wiki/en/latest/cli_interface.html#tox--r) it will create a clean env. And to get the `pip freeze` output, you could add `pip freeze` as a command in `tox.ini`. For the validation issue, could you take a look at the actual data to see if `dimension_names` isn't being written, or if it is and there's an issue with `ome-zarr-models`?

joshmoore (approved):
Thanks, @will-moore! 🙌🏽

PR #515, opened 2026-01-21
Add image class

jo-mueller:
Hi @will-moore ,

this is an implementation of my idea for a class-based, user-facing API to writing and reading.

### Key features

- **Implemented classes**: Implements the `OMEZarrImage`, `OMEZarrMultiscale` and `OMEZarrLabels` (similar to the implementation over at ngff-zarr), that serve as primary entrypoints to the writing. `OMEZarrImage` accepts the data to be written as an array and coerces it to dask internally. It requires the `axes` (i.e., `"tczyx"`) to be passed at instantiation. It also accepts kwargs for pixel sizes (`scale`), axes units (`axes_units`), display settings, channel names and the name of the image (`name`) which are later serialized in the ome-zarr metadata.
  The `OMEZarrMultiscale` and `OMEZarrLabels` then construct a pyramid using the already existing methods (`_build_pyramid`) that were implemented in #516, using different defaults for downsampling (`resize` and `nearest`).
- **Ome-zarr-models-py**: Usage of [ome-zarr-models-py](https://github.com/ome-zarr-models/ome-zarr-models-py) for the internal construction of the corresponding `Metadata` class from there for simpke serialization and de-serialization in the write/read process. Primarily, the coordinate transformation classes and the `Multiscales` metadata classes are used.
  Importantly, all metadata is internally coerced to `ozmp.v05.Multiscales`. Only on writing the metadata class is converted to whatever ome-zarr version is desired.
- **Writing**: Writing happens through the `OMEZarrMultiscale.to_ome_zarr()` method. This method makes use of the already existing writing API from #531 (`_write_pyramid_to_zarr`). It then converts the metadata to the chosen version and uses pydantic's `object.model_dump()` to create the metadata dictionary. Importantly, the version conversion is only implemented in https://github.com/ome-zarr-models/ome-zarr-models-py/pull/398, so this is currently blocked by that.
- **Reading**: The implemented `OMEZarrMultiscale`/`OMEZarrLabels` class also has an attached `from_ome_zarr(...)` classmethod. The argument is simply the path/group of the ome-zarr image. The function then reads the metadata and the multiscales as dask arrays and returns an instance of `OMEZarrMultiscale`. The version is automatically detected and again coerced to `ozmp.v05.Multiscales` internally.
- **Labels**: Writing labels can be done by converting them to instances of `OMEZarrMultiscale` and passing them as a single image or as a `dict(str, OMEZarrLabels)` to the `to_ome_zarr` writer function. The `labels` attribute is also automatically populated when an image is read using the `.from_ome_zarr` function: The reader then searches for all labels that exist under the `labels` zarr group and adds them as a `dict[str, OMEZarrLabels]` to the attribute. One can the add more label images to the attribute, call `to_ome_zarr(..., overwrite=False)`, and only the added label images will be written and the corresponding metadata will be updated
- Backwards compatibility: The existing entrypoints to writing (`write_image` and `write_labels` now use the `OMEZarrImage`,`OMEZarrMultiscale` and `OMEZarrLabels` classes under the hood. Since the scale transformations are now entirely calculated under the hood, this means that the `coordinateTransformations` argument is now essentially deprecated, which I think is good since it was unvalidated anyway. The raised warning message reflects this.
- Inheritance: The introduce `OMEZarrMultiscale` and `NgffLabels` share quite a bit of functionality regarding reading, writing and pyramid generation, so they both derive from a common ancestral class `OMEZarrMultiscaleBase`, which handles that and offers hook functions for the child classes' respective functionality:
  - `_parse_additional_metadata`: Called on end of `__init__`, can be used to parse more stuff (i.e., `omero` or `iamge-labels` metadata.
  - `_write_additional_meta_data`: Called on end of `to_ome_zarr()` - derived classes can use it to validate and dump more metadata to a store that's only relevant to the respective implementation of `OMEZarrMultiscaleBase`
  - `_read_additional_metadata`: Called on end of `read_ome_zarr()`: Can be used to populate additional (meta)data fields from store, i.e., the `.labels` field, etc.

All in all, I think especially the `to_ome_zarr` and `from_ome_zarr` methods are super convenient. I have written a follow-up implementation of the `scene` metadata from 0.6 and making use of the same API there makes a lot of sense. We could think of similar entrypoints to writing HCS layouts.

### TODOs:

- [x] Document
- [x] Support `omero` metadata and `image-label` metadata
- [x] Increase coverage
- [x] Merge #548 
- [x] Merge #557 
- [x] Merge #553


will-moore:
@jo-mueller Could you update the description to reflect where this is heading now?

Are we planning to keep all the existing write methods (any API changes)?

ome_zarr/image.py
```diff
@@ -0,0 +1,489 @@
...
+
+            if list_of_labels:
+                group_labels = group["labels"]
+                group_labels.attrs["ome"] = {
+                    "version": version,
+                    "labels": list_of_labels,
+                }
+
+        else:
+            raise ValueError(f"Unsupported OME-Zarr version: {version}")
```
will-moore: This is a duplicate of the same check and exception raise as above (line 289)

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,489 @@
...
+
+            ome_attrs = group.attrs.get("ome", {})
+            metadata_json = ome_attrs.get("multiscales", [None])[0]
+            metadata = Multiscalev05.model_validate(metadata_json)
+
+            if "labels" in group:
+                labels_ome_attrs = group["labels"].attrs.get("ome", {})
+                list_of_labels = labels_ome_attrs.get("labels", [])
+        else:
+            raise ValueError(f"Unsupported OME-Zarr version: {version}")
```
will-moore, line 540: I understand we are not going to support *writing* of v0.1-0.3 but are we going to try and support *reading* of older data?

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,489 @@
...
+            [str(label.name) for label in labels_dict.values()] if self.labels else []
+        )
+
+        # write the metadata to disk
+        if isinstance(group, str):
+            group = zarr.open(group, mode="r+")
+
+        if version == "0.4":
+            # in v0.4, metadata is stored under "multiscales" attribute
+            metadata_dict = self.metadata.to_version("0.4").model_dump()
```
will-moore: With this code...

```
img_path = "https://livingobjects.ebi.ac.uk/idr/zarr/v0.5/idr0062A/6001240_labels.zarr"
ms = NgffMultiscales.from_ome_zarr(img_path)
ms.to_ome_zarr("test_image_class_to_ome_04.zarr", version="0.4")
```

I get this error:
```
  File "/Users/wmoore/Desktop/python-scripts/zarr_scripts/ome-zarr-py/test_image_class.py", line 31, in <module>
    ms.to_ome_zarr(out_path, version="0.4")
  File "/Users/wmoore/Desktop/ZARR/ome-zarr-py/ome_zarr/image.py", line 317, in to_ome_zarr
    delayed += ms_labels.to_ome_zarr(
               ^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/wmoore/Desktop/ZARR/ome-zarr-py/ome_zarr/image.py", line 334, in to_ome_zarr
    metadata_dict = self.metadata.to_version("0.4").model_dump()
                    ^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/wmoore/opt/anaconda3/envs/omeroweb_zarrv3/lib/python3.12/site-packages/pydantic/main.py", line 991, in __getattr__
    raise AttributeError(f'{type(self).__name__!r} object has no attribute {item!r}')
AttributeError: 'Multiscale' object has no attribute 'to_version'
```

jo-mueller: This works on my end...maybe you need to update ome-zarr-models-py? The necessary function there has only been implemented in `ome-zarr-models-py >= 1.6`

will-moore: However, it does seem to write the data before failing.

will-moore: Yes, `ome-zarr-models-py >= 1.6` fixes it, thanks.

I noticed that the `omero` block in that data is not read or preserved when writing the data (both for v0.4 and v0.5).

will-moore: And for Labels images, the `image-label` metadata (e.g. at https://ome.github.io/ome-ngff-validator/?source=https://livingobjects.ebi.ac.uk/idr/zarr/v0.5/idr0062A/6001240_labels.zarr/labels/0/ ) is not preserved.

jo-mueller: Yeah...both fields aren't really handled (at all) yet 😬

ome_zarr/classes/image.py
jo-mueller: Added in the latest version! I didn't have a full set with all possible metadata (`metadata`, `type`, `omero`) available, but for relatively "bare" image data it works :)

ome_zarr/image.py
jo-mueller: Thanks, fixed in e9d6bbfa0bbd7356de2a8c7809608f4de255a872

ome_zarr/classes/image.py
jo-mueller: Ok, added in d7b4865eb6f7bb3e86aa8d0cc6ee6779af30b44a and 4c2786f93f01c0eb27fff8bc5294cba60b559d91 👍

will-moore:
```
from ome_zarr import NgffMultiscales
img_path = "https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.1/1884807.zarr"
ms = NgffMultiscales.from_ome_zarr(img_path)
out_path = "test_image_class_1884807_05.zarr"
ms.to_ome_zarr(out_path, version="0.5")
```
This writes an invalid image because the `zarr.json` has `datasets[0].path: "0"` but the array is written to path `s0`.

EDIT: Also the `omero` metadata is not preserved into the output image

will-moore:
```
img_path = "https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0076A/10501752.zarr"
NgffMultiscales.from_ome_zarr(img_path)
```

```
  File "/Users/wmoore/Desktop/ZARR/ome-zarr-py/ome_zarr/classes/image.py", line 547, in from_ome_zarr
    raise ValueError(f"Unsupported OME-Zarr version: {version}")
ValueError: Unsupported OME-Zarr version: 0.4.0
```

This image has `.zattrs` with an "unexpected" version:

```
    "_creator": {
        "name": "omero-zarr",
        "version": "0.4.0"
    },
    "multiscales": [
        { "version": "0.4"...
```

so the version lookup needs to be a bit more specific

jo-mueller:
> ```
> from ome_zarr import NgffMultiscales
> img_path = "https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.1/1884807.zarr"
> ms = NgffMultiscales.from_ome_zarr(img_path)
> out_path = "test_image_class_1884807_05.zarr"
> ms.to_ome_zarr(out_path, version="0.5")
> ```
> 
> This writes an invalid image because the `zarr.json` has `datasets[0].path: "0"` but the array is written to path `s0`.

@will-moore this turned out a tougher nut than I thought. I essentially do the following to fix:

When the class ingests a remote ome-zarr, it becomes a handle on this object. The internal metadata *should* reflect the metadata of the remote ome-zarr (i.e. dataset paths `0`, `1`, etc). **But** for writing, the zarr group names `s0`, `s1`, etc. are hardcoded, and I think it is reasonable to do it that way. Still, this introduces what you were observing - the written metadata deviates from the internal metadata representation.

Now, what I did in 66dd14eb65a38ce7cea64cf2269e17e1a4b16952 was to create a copy of the internal metadata *on write* that will also use the same hardcoded path formalism (`s0`, `s1`...) and dump this to json. This way, the written metadata and zarr groups line up.

As for the `NgffMultiscales` object, the actual kept internal metadata representation is left unchanged, so that the object still retains a correct handle on the original ome-zarr object.

Phew!

will-moore:
Thanks @jo-mueller - testing with...

```
img_path = "https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.4/idr0076A/10501752.zarr"
ms = NgffMultiscales.from_ome_zarr(img_path)
out_path = "test_image_class_10501752_05.zarr"
ms.to_ome_zarr(out_path)
```

This reads the version correctly 👍 .
This is now valid with dataset path `s0` etc in the metadata for the Image and the labels too. 👍 

It still seems to be missing the `omero` metadata in the written image.

jo-mueller:
@will-moore 

> It still seems to be missing the omero metadata in the written image.

Right...in 0.4 neither omero nor image-labels are stunder in the multiscales group. Fixed!

jo-mueller:
@will-moore I added a bunch more documentation on how handling labels works in the context of class-based image handling, I think this is coming together nicely :) Some precommit issues to clean up, but otherwise I think this is good to go. I can add some more documentation on omero metadata handling but tht doesn't have to be part of this PR (it's big enough as it is)

docs/source/advanced/labels/adding_labels.ipynb
```diff
@@ -0,0 +1,269 @@
...
+ "cells": [
+  {
+   "cell_type": "markdown",
+   "id": "f2b8b166",
+   "metadata": {},
+   "source": [
+    "# Adding labels to ome-zarrs\n",
+    "(advanced:labels:adding_labels)=\n",
+    "\n",
+    "Unlinke demonstrated in the [basic labels example](basic:labels), in many scenarios, it is nt desirable to write a complete structure consisting of an ome-zarr image and some labels at once.\n",
```
will-moore: Typos "Unlinke" and "nt"

will-moore:
Docs error on `basic/write_image.html`

Screenshot 2026-05-05 at 15 51 52

docs/source/basic/read_image.ipynb
```diff
@@ -2,14 +2,267 @@
...
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "id": "9eb6730c",
+      "metadata": {},
+      "source": [
+        "## Read OME-ZARR images: Legacy API\n",
+        "\n",
+        "The code below here demonstrates the legacy API for reading OME-ZARR images.\n",
```
will-moore: Is this really a "legacy" API?
I think that `write_image()` is very handy and shouldn't be deprecated / considered legacy?

All the "Customizing the pyramid" examples all use `write_image()` and it's not described as legacy there.

docs/source/basic/read_image.ipynb
```diff
@@ -2,14 +2,267 @@
...
+    {
+      "cell_type": "code",
+      "execution_count": null,
+      "id": "ad27af87",
+      "metadata": {},
+      "outputs": [],
+      "source": [
+        "from ome_zarr import NgffMultiscales\n",
+        "\n",
+        "url = \"https://uk1s3.embassy.ebi.ac.uk/idr/zarr/v0.5/idr0062A/6001240_labels.zarr\"\n",
```
will-moore: All `uk1s3.embassy` URLs will need to be updated to `livingobjects` e.g. 
https://livingobjects.ebi.ac.uk/idr/zarr/v0.5/idr0062A/6001240_labels.zarr since the storage is being migrated. `uk1s3.embassy` will be decommissioned soon.

docs/source/advanced/labels/adding_labels.ipynb
jo-mueller: 50a826c0e6023a1d333c46d4980772d3c3517bf2

docs/source/basic/read_image.ipynb
jo-mueller: Should also happen on main, but also changed in 50a826c0e6023a1d333c46d4980772d3c3517bf2

docs/source/basic/read_image.ipynb
jo-mueller: > I think that write_image() is very handy and shouldn't be deprecated / considered legacy?

@will-moore we discussed it a bit [above](https://github.com/ome/ome-zarr-py/pull/515#issuecomment-4012795166). Since `write_image` is now just using the classes under the hood, I don't think it's too much hassle to keep `write_image` and possibly `write_labels` around?

I'm not so sure about the reader, but maybe that's a different discussion.

For now, I have changed the wording in the docs to document the class-based API and  `write_image(...)` as equally functional approaches. I don't think I would document future expansions to the class-based API (i.e., #570) equally balanced, though.

jo-mueller:
pre-commit.ci autofix

jo-mueller:
@will-moore do you think it's worthwhile to close and reopen this PR for a clean commit history?

docs/source/basic/write_image.ipynb
```diff
@@ -8,39 +8,107 @@
...
+              "[]"
+            ]
+          },
+          "execution_count": 3,
+          "metadata": {},
+          "output_type": "execute_result"
+        }
+      ],
+      "source": [
+        "ngff_image = NgffImage(\n",
```
will-moore: The previous code block was preceded by text "Let’s first create some random data to write:" so it looks like the equivalent is missing here e.g. "We create an NgffImage with the data and wrap it with an NgffMultiscales object which builds a pyramid of dask-based NgffImages by downsampling in x and y. The data is finally written to disk at the specified path"

docs/source/basic/write_image.ipynb
```diff
@@ -54,27 +122,31 @@
         "\n",
         "The default version of OME-NGFF is v0.5, which is based on Zarr v3.\n",
         "A zarr v3 group and store is created by `zarr.open_group()` below.\n",
-        "To write OME-NGFF v0.4 (Zarr v2), add the `fmt=FormatV04()` argument."
+        "To write OME-NGFF v0.4 (Zarr v2), add the `zarr_format=2` argument."
```
will-moore: Maybe not broken in this PR, but `zarr.open_group()` is not shown below.
Also the `zarr_format=2` doesn't match the `fmt=FormatV04()` in the code below.

docs/source/basic/read_image.ipynb
```diff
@@ -2,34 +2,287 @@
...
+      "source": [
+        "ngff_image.images[0].data"
+      ]
+    },
+    {
+      "cell_type": "markdown",
+      "id": "9eb6730c",
+      "metadata": {},
+      "source": [
+        "## Direct read\n",
```
will-moore, line 290: I would be happy to remove this "Direct read" alternative from the docs and to deprecate it, but I guess it doesn't have to happen in this PR.

tests/test_writer.py
```diff
@@ -110,14 +109,211 @@ def create_data(self, shape, dtype=np.uint8, mean_val=10):
     @pytest.fixture(
         params=(
             (1, 2, 1, 256, 256),
-            (3, 512, 512),
-            (300, 500),  # test edge chunks of different shapes
+            (32, 256, 256),
+            (256, 256),  # test edge chunks of different shapes
```
will-moore, line 118: I don't remember if it was important to "test edge chunks of different shapes" but the test will still pass if we use e.g. `(640, 320)` here, since both those dimensions can be cleanly divided by 2 for each of 4 levels, and also "test edge chunks of different shapes"

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,671 @@
...
+            metadata_json = cast(
+                dict[str, Any], group.attrs.get("multiscales", [None])[0]
+            )
+
+            axes = (
+                AxisV05(name="t", type="time"),
+                AxisV05(name="c", type="channel"),
+                AxisV05(name="z", type="space"),
+                AxisV05(name="y", type="space"),
+                AxisV05(name="x", type="space"),
```
will-moore: axes shouldn't be hard-coded for v0.3 data as these images can be less than 5D.
E.g. https://livingobjects.ebi.ac.uk/idr/zarr/v0.3/9836842.zarr has 3 axes and `NgffMultiscales.from_ome_zarr("https://livingobjects.ebi.ac.uk/idr/zarr/v0.3/9836842.zarr") fails` because of this mismatch

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,671 @@
...
+            if "omero" in group.attrs:
+                omero_dict = group.attrs.get("omero", None)
+            if "image-label" in group.attrs:
+                image_label_dict = group.attrs.get("image-label", None)
+        elif version == "0.5":
+            from ome_zarr_models.v05.multiscales import Multiscale as Multiscalev05
+
+            ome_attrs = cast(dict, group.attrs.get("ome", {}))
+            metadata_json = ome_attrs.get("multiscales", [None])[0]
+            metadata = Multiscalev05.model_validate(metadata_json)
```
will-moore, line 410: If no `multiscales` is found (e.g. you passed in a `plate` or a `bioformats2raw.layout` group, then you don't get a useful error message (below) so maybe a `None` check here would be nice:

```
  File "/Users/wmoore/Desktop/ZARR/ome-zarr-py/ome_zarr/classes/image.py", line 603, in from_ome_zarr
    metadata = Multiscalev05.model_validate(metadata_json)
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/wmoore/opt/anaconda3/envs/omeroweb_zarrv3/lib/python3.12/site-packages/pydantic/main.py", line 705, in model_validate
    return cls.__pydantic_validator__.validate_python(
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
pydantic_core._pydantic_core.ValidationError: 1 validation error for Multiscale
  Input should be a valid dictionary or instance of Multiscale [type=model_type, input_value=None, input_type=NoneType]
    For further information visit https://errors.pydantic.dev/2.11/v/model_type
```

pyproject.toml
```diff
@@ -26,6 +26,7 @@ dependencies = [
     "scikit-image>=0.19.0",
     "toolz",
     "rangehttpserver",
+    "ome-zarr-models >= 1.6",
```
will-moore, line 29: With this, we can remove `ome-zarr-models` from test dependencies above

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,715 @@
...
+        # validate dimensions match data shape
+        if len(self.axes) != len(self.data.shape):
+            raise ValueError(
+                f"Number of dimensions in data ({len(self.data.shape)}) "
+                f"does not match number of dims ({len(self.axes)})"
+            )
+
+
+@dataclass
+class NgffMultiscales:
```
lubianat: Why `NgffMultiscales` and not `NgffMultiscale`?

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,715 @@
...
+    method : str or Methods, optional
+        Downsampling method to use. Default: Methods.RESIZE.
+    labels : :class:`NgffMultiscales` or dict of str to :class:`NgffMultiscales`, optional
+        Optional labels to associate with the image pyramid.
+        Can be a single :class:`NgffMultiscales` instance (for a single label pyramid)
+        or a dict mapping label names to :class:`NgffMultiscales` instances
+        (for multiple label pyramids), e.g.
+        `{'nuclei': nuclei_multiscale, 'cells': cells_multiscale}`.
+        Default is None (no labels).
+    omero : dict or Omero, optional
```
lubianat: why does omero metadata go in `NgffMultiscales` and not in `NgffImage`?

I guess it makes sense in practice, but for me it feels like semantically separate things.

No strong opinions

lubianat: also, it _could_ make sense to not use "omero" in the API and abstract it away. E.g. "display_config" or something

As the idea is to keep the API more "future proof", it could make sense to label it something else. It may well be that keeping "omero" for now is clearer, though.  

Again, no strong opinions.

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,715 @@
...
+        Optional Omero metadata to include in the OME-Zarr attributes.
+        Can be passed as a dict or an instance of the [Omero model](https://ome-zarr-models-py.readthedocs.io/en/stable/api/v04/image/#omero-metadata)
+        Default is None (no Omero metadata).
+        For example metadata, see [ngff specification](https://ngff.openmicroscopy.org/specifications/0.5/index.html#omero-metadata-transitional)
+    image_label : dict or Label, optional
```
lubianat: For me it is confusing to have a "labels" and a "image_label" parameter; the explanation is clear, but the meaning of the names is not immediately obvious. I think this inherits the names in the 0.5 spec, right? It may be too coupled. 

A few options (none great): "labels_rendering_metadata", "label_metadata"

lubianat:
hi @jo-mueller I've tested it here, but not sure I am doing things right

first I converted OME-TIFF to OME-Zarr with the new API (https://github.com/lubianat/bbbc032_to_zarr/blob/test_new_ome_zarr_api/test_new_api.py). I think it went ok, similar to what I had done previously (https://github.com/lubianat/bbbc032_to_zarr/blob/test_new_ome_zarr_api/ome_tiff_to_ome_zarr.py). 

grafik


then I tried to add `omero` and label information, but that did not work. that could be on me, but maybe that is informative either way; 

https://github.com/lubianat/bbbc032_to_zarr/blob/test_new_ome_zarr_api/test_new_api_add_labels.py

got me 
```
  File "/home/lubianat/Documents/gerbi/bbbc032_to_zarr/test_new_api_add_labels.py", line 33, in <module>
    ms.to_ome_zarr(out_path, version="0.5")
    ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/lubianat/Documents/gerbi/bbbc032_to_zarr/.venv-pr515/lib/python3.13/site-packages/ome_zarr/classes/image.py", line 400, in to_ome_zarr
    for label_name, ms_labels in labels_dict.items():
                                 ^^^^^^^^^^^^^^^^^
AttributeError: 'NgffMultiscales' object has no attribute 'items'
```

If instead of `ms.labels = labels_multiscale` I use `ms.labels = {"nuclei": labels_multiscale}`, it runs and write the labels... but deletes the array data! 

before: 
grafik

after: 
grafik

A similar failure happens to me when I try to add omero (https://github.com/lubianat/bbbc032_to_zarr/blob/test_new_ome_zarr_api/test_new_api_add_omero.py)

I see the omero in the `ms` object when I print it, but when I write it back (in place or elsewhere), I don't see the metadata block. 


Again, I may be getting things very wrong, but it is a kind of user testing any ways :)

jo-mueller:
@lubianat thanks for the test cases, this is suoer useful! I think I know what's happening. Will reproduce and fix first thing on Monday 👍

ome_zarr/classes/image.py
jo-mueller: That's a good idea. The point of the API is to abstract reading and writing a bit away from the actual terms of the spec (i.e., `omero` or `image_label`, especially since they may not always be obvious or subject to change.

how about `display_config` and `label_object_metadata`? The latter maybe transports more clearly that the metadata describes individual labels in the image.

ome_zarr/classes/image.py
jo-mueller: Errr....no good reason. I guess when I talk/think about images, and query myself "what kind of object is it?" The answer is either *a Multiscales* or *a multiscale image* 🤔

jo-mueller:
@lubianat I think 8ab91e1d48f232eb9043c01d8005c60023c83a4c fixes most of what you describe.

> If instead of ms.labels = labels_multiscale I use ms.labels = {"nuclei": labels_multiscale}, it runs and write the labels... but deletes the array data!

I think what happened here was primarily that setting `ms.labels` directly (i.e., doing `ms.labels = ...` rather than `NgffMultiscales(...,  labels=some_labels_image`) doesn't pass through the internal model validation step. That's not a problem on init of the class because the error is escaped, but it is a problem on write, where I don't think invalid metadata is escaped properly. So it stops the write halfway through, i.e., after writing the first folders or so.

The same happens with the ´omero´ attribute. It is passed through validation on init of the class, but not on write. So if it is injected post-init (i.e., by `my_class.omero = some_dict`), it's actually not validated and weird things happen on write).

Long story short, I added proper setter methods that validate the `labels`, `omero` and `image_label` field irrespective of how they are passed, which fixes the example on my end. 

Except for this one:

> but when I write it back (in place)

I expect this to fail. Primarily, if you open an existing ome zarr like this:

```python
ms = NgffMultiscales(path)
ms.omero = some_metadata
ms.to_ome_zarr(path)
```

this fails because the image data is never actually in memory, but just loaded lazily on write. Since the store is deleted on write though, you'd delete the store you're trying to lazily read from, which has to fail.

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,671 @@
...
+        # set default scale if unset
+        if not self.scale:
+            self.scale = tuple(1.0 for _ in range(len(self.axes)))
+
+        # coerce axes to list
+        if isinstance(self.axes, str):
+            self.axes = list(self.axes)
+
+        # coerce scale to dict if it's a sequence
+        if isinstance(self.scale, Sequence):
```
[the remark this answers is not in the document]
clbarnes: You could also use `zip(a, b, strict=True)`.

ome_zarr/classes/image.py
clbarnes: I would go with Multiscale or maybe even Image for this class, because multiscale is the only type of image supported by OME-Zarr and image in the spec always refers to multiscales. I would then refer to the current NgffImage class as something like NgffSingleScale or NgffScaleLevel to make it clear that it should only ever be used as part of a multiscale, not independently.

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,785 @@
...
+
+        # validate dimensions match data shape
+        if len(self.axes) != len(self.data.shape):
+            raise ValueError(
+                f"Number of dimensions in data ({len(self.data.shape)}) "
+                f"does not match number of dims ({len(self.axes)})"
+            )
+
+
+@dataclass
```
clbarnes: I would recommend not using dataclasses here. Even though we can use post_init to do validation, it still exposes all of the input arguments as-is and means we have to type-annotate the flexible input type even if we normalise it to a single type (e.g. with labels).

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,785 @@
...
+        The physical scale for each axis. If a sequence is provided, it should
+        match the order of `axes`. If a dict is provided, keys should be axis names,
+        e.g. {'x': 0.1, 'y': 0.1, 'z': 0.5}. Default is None, which sets all scales to 1.0.
+    axes_units : dict of str to str, optional
+        Units for each axis, e.g. {'x': 'micrometer', 'y': 'micrometer'}.
+        Default is None (no units).
+    name : str, optional
+        Name of the image. Default is "image".
+
+    Attributes
```
clbarnes: Personal preference but most IDEs and doc generators support type annotations and [docstrings-after-variables](https://pdoc.dev/docs/pdoc.html#document-variables) so I tend to rely on that rather than duplicating the information in the docstring explicitly.

Same thing for the arguments section - I don't explicitly write the types any more if I'm using type annotations, just an explanation. You can also explicitly add a docstring using the `Annotated[float, "this is a number"]` form but I don't know what that looks like in practice.

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,785 @@
...
+    axes_units: dict[str, str] | None = None
+    name: str | None = "image"
+
+    def __post_init__(self):
+        # set default scale if unset
+        if self.scale is None:
+            self.scale = tuple(1.0 for _ in range(len(self.axes)))
+
+        # coerce axes to list
+        if isinstance(self.axes, str):
```
clbarnes, line 102: Don't need the `if` here, as `list(["a", "b", "c"])` is the same as `list("abc")`.

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,785 @@
...
+        The physical scale for each axis.
+    axes_units : dict of str to str
+        Units for each axis, e.g. {'x': 'micrometer', 'y': 'micrometer'}.
+        Default is None (no units).
+    name : str
+        Name of the image.
+    """
+
+    data: da.Array | np.ndarray
+    axes: Sequence[str] | str
```
clbarnes, line 95: Up to you but I personally would prefer to pass in a list of pre-validated `Axis` objects or even a pre-validated `Axes` object rather than separate axes and axes_units (plus any further info which might go into axes later, like types).

ome_zarr/classes/image.py
clbarnes: Moving away from dataclasses would allow decoupling the input arguments and internal representation, as well as making omero keyword-only so it's easier to deprecate later.

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,785 @@
...
+        from ome_zarr.scale import _build_pyramid
+
+        if scale_factors is None:
+            scale_factors = (2, 4, 8, 16)
+
+        self.name = image.name
+        method = self.method
+
+        if isinstance(method, Methods):
+            method = str(method.value)
```
clbarnes, line 156: This cast wouldn't be necessary if `Methods` were a `StrEnum`.

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,785 @@
...
+            method = str(Methods.RESIZE.value)
+
+        if self.image_label is not None and method != str(Methods.NEAREST.value):
+            warnings.warn(
+                f"Image label metadata provided but method is {method}, not 'nearest'. "
+                f"Consider using method='nearest' label image multiresolution pyramids."
+            )
+
+        # Build the pyramid data
+        pyramid = _build_pyramid(
```
clbarnes, line 161: Does this generate scale levels on the fly even if we already have them stored on disk?

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,785 @@
...
+                # We don't want to fail the entire initialization
+                # if the omero metadata is invalid, so we
+                # escape possible validation errors and just warn
+                # the user that the omero metadata is invalid
+                return Omero.model_validate(omero)
+            except ValidationError as e:
+                warnings.warn(f"Invalid Omero metadata: {e}")
+                return None
+        else:
+            raise ValueError(
```
clbarnes: TypeError could be more clear.

ome_zarr/writer.py
```diff
@@ -283,6 +283,11 @@ def write_multiscale(
         5-dimensional with dimensions ordered (t, c, z, y, x)
     :type group: :class:`zarr.Group`
     :param group: The group within the zarr store to store the data in
+    :type scale: dict of str to float, optional
```
clbarnes, line 286: Docstring style is inconsistent with the new classes - either is fine but a mixture isn't ideal.

ome_zarr/writer.py
```diff
@@ -556,17 +564,23 @@ def write_well_metadata(
...
     fmt: Format | None = None,
     axes: AxesType = None,
     coordinate_transformations: list[list[dict[str, Any]]] | None = None,
     storage_options: JSONDict | list[JSONDict] | None = None,
-    compute: bool | None = True,
+    compute: bool = True,
     scale: dict[str, float] | None = None,
     axes_units: dict[str, str] | None = None,
-    **metadata: str | JSONDict | list[JSONDict],
+    **metadata: JSONDict,
```
clbarnes, line 583: I know this fits with the previous API but using `**kwargs` here makes it risky to change the signature of this function in future; I'd usually use `dict[str, JSONDict]` to keep it contained.

ome_zarr/writer.py
```diff
@@ -711,32 +729,25 @@ def write_image(
         else:
             method = Methods.RESIZE
 
-    if method is None:
-        method = Methods.RESIZE
+    omero = None
+    if "omero" in metadata:
+        omero = metadata["omero"]
```
clbarnes: `metadata.get("omero")`

ome_zarr/writer.py
```diff
@@ -1184,11 +1204,15 @@ def write_labels(
...
+    scale: dict of str to float, optional
+        The physical pixel size for each dimension, e.g. {"z": 0.1, "y": 0.1, "x": 0.5}.
+        The pixel sizes for every resolution level are calculated directly from the defined `scale` and
+        `scale_factors` for each level.
     scaler : ome_zarr.scale.Scaler, optional
         [DEPRECATED] Scaler implementation for downsampling the label data. Passing this
         argument will raise a warning and is no longer supported. Use `scale_factors` and
         `method` instead.
-    scale_factors : tuple of int, optional
+    scale_factors : Sequence[int] | tuple[int, ...] | list[dict[str, int]], optional
```
clbarnes, line 1214: `tuple[int, ...]` is contained by `Sequence[int]`, I think, but maybe type checkers don't like it.

jo-mueller:
## Plan to move forward

- **Easier omero/image-label metadata dump**: Currently, while a bunch of other parameters are abstracted away nicely from the terms of the spec (i.e., scale, axes_units, etc), other parameters aren't - like, not at all, such as the `omero` and `image_labels` fields, which you need to construct yourself an then pass to the `NgffMultiscales` class. As @70Gage70 [suggested](https://forum.image.sc/t/ngff-weekly-dev-update-thread/110810/91), this can be added as additional kwargs to the `NgffImage` class, i.e. `channel_names`, `channel_color` and `contrast_limits` keywords that are then coerced to the `Omero` metadata class under the hood.
- **Renaming image classes**: Refactor `NgffImage` and `NgffMultiscales` into `OMEZarrImage` and `OMEZarrMultiscales`
- **Explicit image classes for images and labels**: Separate of `OMEZarrMultiscales` into distinct `OMEZarrMultiscales` and `OMEZarrLabels` classes. This will require abstracting common functionality of both into a common base class (`OMEZarrMultiscalesBase`), that defines the basic writing and reading functions (i.e., `to_ome_zarr` and `from_ome_zarr` and provides hooks to the derived classes `OMEZarrMultiscales` and `OMEZarrLabels` to write their respective additional metadata and fields. I.e., `OMEZarrMultiscales` would only write the `omero` metadata and provide the `.labels` attribute to access assoicated labels. The `OMEZarrLabels` class would make sure the correct downsampling is used and serializes the `image_label` metadata along with the rest of the metadata
- **No dataclasses**: As @clbarnes pointed out, we'll probably need more control over the properties and their associated getters and setters. In that case, there's not much advantage to the use of Python `dataclasses` any more, I think.

This will probably be a bit of a lift from the current code. I'll work on it in a separate branch/PR and then port the changes over to this one if it comes out nicely. Feedback on the plans very welcome!

cc @will-moore @clbarnes @lubianat @70gage70 @dstansby @melonora @kevinyamauchi

ome_zarr/classes/image.py
jo-mueller: That's kind of what we had before...but we were trying to save users from the additional labor of constructing (valid) axes dicts. It's through though, this abstractive approach will get harder as we add more and more metadata into these fields

jo-mueller:
@will-moore @clbarnes @lubianat @70Gage70 @dstansby @melonora @kevinyamauchi I think this is ready for another look. I basically implemented the above TODOs. Things to look out for:

- No matter how I turned it, passing the image-labels metadata to the `OMEZarrImage` entrypoint is clunky. For images I think it works well (i.e., `channel_names` as a keyword is reasonably clear and easy to convert to omero metadata), but for labels it didn't fi so well. Soooo I added an option in the `OMEZarrLabels` to autopopulate it with some default settings for the image-labels metadata by finding all the unique labels in the array, which could be costly for large images.
- The mentioned `channel_names` kwargs are only used during init of the `OMEZarrImage` classes and then converted to the ome-zarr-models-py `Omero` class instances. Now, if you read an ome-zarr from somewhere else, the `channel_names` of the respective `OMEZarrImage` singlescale images are not reverse-populated. Is that a problem?

I updated all the docs, there should be plenty of code to go by and test it 👍 Have a good weekend everyone!

lubianat:
Thanks for the push! It is looking better and better. How hard would it be to reverse-populate the "channel names" field? I think it should not block this PR, but it would be good to have a ticket for it at least.

jo-mueller:
> How hard would it be to reverse-populate the "channel names" field? 

Probably not at all :)

70Gage70:
This is looking great!

Not a blocker, but could we maybe introduce a method like 

```python 
OMEZarrMultiscale.add_labels(
  group,
  labels: Union[OMEZarrLabels, List[OMEZarrLabels]], 
  other_kwargs, 
  compute=True, 
  overwrite=False
)
```

The idea is that the loop looks like

```python
# create image
path2out = "image_with_labels.zarr"

img = OMEZarrImage(img_data, axes="zyx", name="image")
img_multiscales = OMEZarrMultiscale(image=img)
img_multiscales.to_ome_zarr(path2out)

# have `labels_data1` and labels_data2` from a later segmentation
labels1 = OMEZarrImage(labels_data1, axes="zyx", name="labels1")
labels1_multiscales = OMEZarrLabels(image=labels1)

labels2 = OMEZarrImage(labels_data2, axes="zyx", name="labels2")
labels2_multiscales = OMEZarrLabels(image=labels2)

img_multiscales = OMEZarrMultiscale.from_ome_zarr(path2out)
img_multiscales.add_labels(path2out, [labels1_multiscales, labels2_multiscales])
```

Under the hood, `.add_labels` would update the `.labels` field appropriately, then call `to_ome_zarr(..., overwrite=False)` as per the docs. If a label already exists with any of the names in the provided list, this should fail quickly or warn-and-skip, unless we call `add_labels(..., overwrite=True)`, in which case it should overwrite those labels, but not the image. The point is that the `add_labels` method will ensure that 

- it is impossible to accidentally overwrite the image (users who want to do so can just directly use `.to_ome_zarr(..., overwrite=True)`)
- we can overwrite specific labels as needed
- the `OMEZarrImage` is now the only entry point for the label's name, rather than it also being set in the dict that is passed to the `.labels` field

Perhaps the path could be stored when `from_ome_zarr` is used to avoid `.add_labels` needing that argument.

docs/source/advanced/build_custom_pyramid.ipynb
```diff
@@ -20,36 +20,78 @@
...
       "cell_type": "code",
-      "execution_count": 16,
+      "execution_count": 1,
       "id": "642955b1",
       "metadata": {},
+      "outputs": [],
+      "source": [
+        "import numpy as np\n",
+        "\n",
+        "from ome_zarr import NgffImage, NgffMultiscales\n",
```
will-moore: Renamed `NgffImage` to `OMEZarrImage` now

jo-mueller:
@70Gage70 would you be up to send this functionality to ome-zarr-py in a follow-up PR to this one? I think `add_labels` would indeed give you some needed additional fine-grained control over labels that is inevtiably difficult to shoe-horn into the `.labels` field and the `to_ome_zarr` method, but at least as far as the class-based API is concerned, I'm becoming a bit hesitant to add more functionality to it 🤔 

For the specific points you mention above:

- > it is impossible to accidentally overwrite the image (users who want to do so can just directly use .to_ome_zarr(..., overwrite=True))

  I did change the default for `overwrite` to `False`, so that should reduce some risk
- > we can overwrite specific labels as needed

  If you wanted to overwrite a specific labels image that already existed under `image/labels/labels1`, you can also do `some_labels_multiscales.to_ome_zarr(group=image/labels/labels1, overwrite=True)`, and that'll have all the metadata correct, because it has previously been written
- > the OMEZarrImage is now the only entry point for the label's name

  THAT is a good point. And then again, in practice, nobody forces the `name` in the labels multiscale group to be the same as the group itself, so having two different entrypoints kind of reflects that? I think right now, the name passed in the dictionary takes precedence. But re what I was writing above, `add_labels` would give more control in that regard so 👍

jo-mueller:
Some feedback from @will-moore and @kevinyamauchi:

- It's not good for the `OMEZarrImage` class to retain too many attributes (i.e., `channel_names`, etc) that can end up out of sync from the `metadata` field, which contains all the metadata that is later dumped to disk
- More verbosity if field values are inferred (i.e., scale values 1.0 for unspecified scale values)
- Move `channel_names`, etc down into `__init__` of `OMEZarrMultiscale` because that's where it will be parsed into `omero` metadata

70Gage70:
@jo-mueller 

re: `add_labels`. Agree with all your points, I'll make a PR at a later time 👍

docs/source/basic/write_image.ipynb
```diff
@@ -8,54 +8,133 @@
...
+          },
+          "execution_count": 3,
+          "metadata": {},
+          "output_type": "execute_result"
+        }
+      ],
+      "source": [
+        "ngff_image = OMEZarrImage(\n",
+        "  data=data,\n",
+        "  axes=\"zyx\",\n",
```
dstansby: I thought there was some special meaning to passing `axes='zyx'` here, and it wasn't obvious how to generalise to arbitrary axes names. Might be nice to replace with `axes=['z', 'y', 'x']` to make it more obvious you can have a list of arbitrary names.

docs/source/basic/write_image.ipynb
```diff
@@ -8,54 +8,133 @@
...
+        "  scale={\"z\": 0.5, \"y\": 0.1, \"x\": 0.1},\n",
```
dstansby: What units are these in?

docs/source/basic/write_image.ipynb
```diff
@@ -8,54 +8,133 @@
...
+        ")\n",
+        "\n",
+        "ngff_multiscales = OMEZarrMultiscale(image=ngff_image)\n",
```
dstansby: It wasn't obvious to me what the point of the `OMEZarrMultiscale` class is from this line, and I had to go digging in the API docs. Passing `scale_factors=...` here would make it more obvious that this is creating the multiscale levels.

docs/source/basic/write_image.ipynb
```diff
@@ -8,54 +8,133 @@
...
+        "ngff_multiscales.to_ome_zarr(path, overwrite=True)"
```
dstansby: Why is the output of this cell an empty list?

docs/source/basic/write_image.ipynb
```diff
@@ -8,54 +8,133 @@
...
+    },
+    {
+      "cell_type": "markdown",
+      "id": "9015365c",
+      "metadata": {},
+      "source": [
+        "## API alternative: Direct write\n",
+        "\n",
+        "Besides the above-described class-based approach, another principle entry-point for writing OME-ZARR images is using the {py:func}`ome_zarr.writer.write_image` function.\n",
+        "This takes an n-dimensional `numpy` array or `dask` array and writes it to the specified zarr group according to the OME-ZARR specification.\n",
```
dstansby, line 130: Or Zarr array it seems in the example below?

docs/source/basic/write_image.ipynb
```diff
@@ -68,27 +147,31 @@
...
         }
-      },
-      "outputs": [],
+      ],
       "source": [
         "from ome_zarr.format import FormatV04\n",
         "\n",
         "path = \"test_ngff_image_v2.ome.zarr\"\n",
-        "write_image(data, path, axes=\"zyx\", fmt=FormatV04())"
+        "write_image(data, path, axes=\"zyx\", fmt=FormatV04()) "
```
dstansby, line 193: It would be a lot nicer if you could pass `format="0.4"` instead of creating a whole new object here.

dstansby (commented):
The examples are very nice! I've added some feedback on the write image example.

docs/source/basic/write_image.ipynb
jo-mueller: Yeah....that's a bit of a remnant of older API days. Will probably change in the future!

docs/source/basic/write_image.ipynb
jo-mueller: The writer has the option for delayed writing. The `to_ome_zarr` function (and the older `write_image` function) has a kwarg `compute`, that defaults to `True`. If changed, the function would return a list of delayed objects. So in theory, you could set up a bunch of conversion jobs (i.e., for writing plates, etc) and set it up as a single, massive list of concurrent write jobs that would all be executed in parallel. 

It's not documented though. If `compute=True`, the list of delayed objects is empty, and that's what you see. I would see that as something for a follow-up PR though to
- document behavior and usage of the `compute` parameter
- return `None` if `compute=False`

docs/source/basic/write_image.ipynb
jo-mueller: 5f965a7f95595b1af03ad7ca67e07bb7e4747d18

docs/source/basic/write_image.ipynb
jo-mueller: 5f965a7f95595b1af03ad7ca67e07bb7e4747d18

docs/source/basic/write_image.ipynb
jo-mueller: 5f965a7f95595b1af03ad7ca67e07bb7e4747d18

docs/source/basic/write_image.ipynb
jo-mueller: I'll mark this as resolved for now and created an [issue ](https://github.com/ome/ome-zarr-py/issues/585) to follow up

docs/source/basic/write_image.ipynb
jo-mueller: Follow-up issue: https://github.com/ome/ome-zarr-py/issues/586

jo-mueller:
Alright, I think I would stop loading this PR with more features other than possibly more tests. I think this is ready for a final, critical look, otherwise I'd rather continue in follow-up PRs so we can go on to use this framework to make 0.6 happen here :)

docs/source/basic/write_image.ipynb
```diff
@@ -68,27 +147,31 @@
         "\n",
         "The default version of OME-NGFF is v0.5, which is based on Zarr v3.\n",
         "A zarr v3 group and store is created by `zarr.open_group()` below.\n",
```
will-moore, line 168: Could you remove this sentence as it's no-longer correct (not due to this PR).

docs/source/basic/write_image.ipynb
```diff
@@ -8,54 +8,133 @@
...
+        "\n",
+        "# create some random data to write\n",
+        "size_xy = 128\n",
+        "size_z = 10\n",
+        "rng = np.random.default_rng(0)\n",
+        "data = rng.poisson(lam=10, size=(size_z, size_xy, size_xy)).astype(np.uint8)"
+      ]
+    },
+    {
+      "cell_type": "code",
```
will-moore, line 63: Before this code block, it would be nice to have a `markdown` block with something like "We create an `OMEZarrImage` with our data, then build a multiscale pyramid of dask arrays by downsampling in x and y by defaul. This is then written to disk to create a OME.zarr image."

docs/source/advanced/build_custom_pyramid.ipynb
will-moore: Renamed NgffImage to OMEZarrImage now

will-moore:
Apologies - I didn't realise that I was in the middle of "submitting a review" but never actually submitted it. So a bunch of the comments in that review are either duplicated or out of date!

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,1043 @@
...
+        each dict should specify scale factors for each axis, e.g. {'x': 2, 'y': 2, 'z': 1}.
+        Default is (2, 4, 8, 16).
+    method : ome_zarr.scale.Methods | str | None
+        Rescaling method to use when generating pyramid levels. Default is Methods.RESIZE.
+    coordinateTransformations :
+        Additional coordinate transformations to include in the metadata for each level.
+    labels : OMEZarrLabels | list[OMEZarrLabels] | dict[str, OMEZarrLabels] | None
+        Labels associated with the image. Can be a single OMEZarrLabels instance, a list of them,
+        or a dict mapping label names to OMEZarrLabels instances. Default is None (no labels).
+    channel_names : list[str] | None
```
will-moore, line 602: It would be nice to see the `channel_names/colors/limits` in the docs somewhere. Maybe another example  on the `write_image.html` page?

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,1043 @@
...
+        contrast_limits: list[tuple[float, float]] | None,
+    ) -> None:
+        """
+        Build omero metadata from channel parameters.
+        """
+        if "c" not in self._images[0].axes:
+            return
+
+        # Validate that channel_names, channel_colors and
+        # contrast_limits are lists of the same length if provided
```
will-moore: Not a blocker, but this whole validation block could be removed as you check that all `channel_*` lists are the same length as `n_channels` below.

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,1043 @@
...
+            else:
+                name = f"Channel {i}"
+
+            if channel_colors is not None:
+                color = channel_colors[i]
+                # Coerce RGBA/RGB list values to hex strings
+                if isinstance(color, (list, tuple)):
+                    # Convert RGB/RGBA to hex, taking first
+                    # 3 values and ignoring alpha
+                    color = f"#{color[0]:02x}{color[1]:02x}{color[2]:02x}"
```
will-moore: doesn't need the `#` - Some implementations may handle it, but it's not in https://ngff.openmicroscopy.org/specifications/dev/index.html#omero-metadata-transitional

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,1043 @@
...
+    "#00FFFF",  # cyan
+    "#FF00FF",  # magenta
+    "#FFFF00",  # yellow
+    "#FF0000",  # red
+    "#00FF00",  # green
+    "#0000FF",  # blue
+    "#FFFFFF",  # white
+    "#FFA500",  # orange
+    "#800080",  # purple
+    "#008000",  # dark green
```
will-moore: As below, can remove the `#` from these.

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,1043 @@
...
+        overwrite: bool = False,
+    ) -> list:
+        from ome_zarr.utils import _recursive_pop_nones
+
+        delayed: list = []
+
+        # Write omero metadata
+        if self._omero and isinstance(self._omero, Omero):
+            omero_dict = _recursive_pop_nones(self._omero.model_dump(by_alias=True))
+            omero_dict["version"] = version
```
will-moore: This is not needed for v0.5

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,1043 @@
...
+        self,
+        channel_names: list[str] | None,
+        channel_colors: list[list[int]] | list[str] | None,
+        contrast_limits: list[tuple[float, float]] | None,
+    ) -> None:
+        """
+        Build omero metadata from channel parameters.
+        """
+        if "c" not in self._images[0].axes:
+            return
```
will-moore: We have examples of images with no `c` dimension where it's still useful to store a single channel info (contrast limits, channel name etc). e.g. https://ome.github.io/ome-ngff-validator/?source=https://livingobjects.ebi.ac.uk/idr/zarr/v0.5/idr0066/ExpD_chicken_embryo_MIP.ome.zarr

So maybe we should consider an image with no `c` dimension as having a single channel?

will-moore:
Not a blocker here, but just some testing/observations...

Tried reading sample from https://ome.github.io/ome-ngff-validator/?source=https://livingobjects.ebi.ac.uk/idr/zarr/v0.5/idr0062A/6001240_labels.zarr
All 3 datasets have `chunk: [1,1,256,256]` there, but when we read, the chunksize gets limited to the shape of the array, so the chunksize is different for 2nd and 3rd arrays...
This means that you have to specify an array for `storage_options` for different shard sizes at each when we write...

```
img_path ="https://livingobjects.ebi.ac.uk/idr/zarr/v0.5/idr0062A/6001240_labels.zarr"
ms = OMEZarrMultiscale.from_ome_zarr(img_path)
print("Read images", len(ms.images), ms.images)
# chunksize=(1, 1, 256, 256)
# chunksize=(1, 1, 137, 135)
# chunksize=(1, 1, 68, 67)

out_path = "test_s3_shards_to_ome_05.zarr"
opts = [{"shards": [1,20,512,512]},
        {"shards": [1, 20, 137, 135]},
        {"shards": [1, 20, 68, 67]}
    ]
ms.to_ome_zarr(out_path, version="0.5", storage_options=opts)
```

The code above writes the image OK, but still fails because it also tries to write the child `labels` image with the same storage_options, but that is not compatible with the chunks of the label image:

```
  File "/Users/wmoore/opt/anaconda3/envs/omeroweb_zarrv3/lib/python3.12/site-packages/zarr/codecs/sharding.py", line 325, in validate
    raise ValueError(
ValueError: The array's `chunk_shape` (got (1, 20, 512, 512)) needs to be divisible by the shard's inner `chunk_shape` (got (1, 59, 69, 136)).
```

It might be nice to have an example in the docs with `ms.to_ome_zarr()` with `storage_options`? I guess it depends how much we want users to be able to understand from the tutorials alone, without looking at the API or code?

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,1043 @@
...
+            else:
+                channel_contrast = (
+                    0,
+                    dtype_max,
+                )  # TODO: best way to get max value from dtype?
+
+            channel_metadata.append(
+                {
+                    "label": name,
+                    "color": color,
```
will-moore, line 802: It would be nice to add `"active": True` to each channel to improve the experience in viewers (when I opened my new image in vizarr, it was completely black as no channels were active).

jo-mueller: d9b720285510ccf7ada8ef88fb64aa0540544d74

pyproject.toml
jo-mueller: 18feea554119bc238f3fedb26897939443bbd8c6

docs/source/basic/write_image.ipynb
jo-mueller: 1f75eba4947195e5ec621271b71a938b205044cf

docs/source/basic/write_image.ipynb
jo-mueller: 1f75eba4947195e5ec621271b71a938b205044cf

ome_zarr/classes/image.py
jo-mueller: Check! 1f75eba4947195e5ec621271b71a938b205044cf

ome_zarr/classes/image.py
jo-mueller: ab2416cc2200aa7ad0ff4832c9b65763038190b2

docs/source/advanced/build_custom_pyramid.ipynb
jo-mueller: d9392f4bc62d02fd0a92c27c01d95a70c1326054

jo-mueller: d9392f4bc62d02fd0a92c27c01d95a70c1326054

ome_zarr/classes/image.py
jo-mueller: d9b720285510ccf7ada8ef88fb64aa0540544d74

ome_zarr/classes/image.py
jo-mueller: 9bcb0827b5ae7163f3bc4044e0adea573f4db54d

ome_zarr/classes/image.py
jo-mueller: 2d5b39e4ff3ddf5e7c5af242af3c07fc0ee339ff

ome_zarr/classes/image.py
jo-mueller: I left it there because it renders nicely in my IDE, but removed it when parsed below (2d5b39e4ff3ddf5e7c5af242af3c07fc0ee339ff)

ome_zarr/classes/image.py
jo-mueller: Fixed in 7a5f28ed98b6b687841054b4801360485c467f3e

Ultimately, we will need something like a class-idependent reader function (i.e. `from_ome_zarr()`) that inspects the ome-zarr and then invokes the respective classes' `OMEZarrclass.from_ome_zarr()` function.

jo-mueller:
@will-moore Thanks for the thorough testing around the chunking. Maybe @melonora can chime in here as well - for once it is a bit weird that the labels data is written with a chunksize different from the image it labels, but ofc it's not forbidden to do that. I see two options:

- Adding a check to the storage options that would limit the passed sharding values down to allowed values (i.e., fall back to the chunksize as value for shard) and issue a warning about it.
- Rechunk image and labels to the same chunking scheme? I feel like that would a bit of an overreaching behavior.
- Better documentation on the details of IO around labels. storage options, etc?

My favorite would be option 1, but ultimately the API I came up with limits us to having the same sharding configuration used for the image data and the associated labels. I think at this point I'd rather get this merged and pursue that in a follow-up PR 😅

ome_zarr/classes/image.py
jo-mueller: 95a7b4816defb83d1e1af729e205d537d9bf30db

docs/source/basic/write_image.ipynb
```diff
@@ -8,54 +8,152 @@
...
-        "size_xy = 128\n",
-        "size_z = 10\n",
-        "rng = np.random.default_rng(0)\n",
-        "data = rng.poisson(lam=10, size=(size_z, size_xy, size_xy)).astype(np.uint8)\n",
-        "\n",
-        "write_image(\n",
-        "    data, path, axes=\"zyx\",scale={\"z\": 1.0, \"y\": 0.5, \"x\": 0.5},\n",
-        "    axes_units={\"z\": \"micrometer\", \"y\": \"micrometer\", \"x\": \"micrometer\"},\n",
-        "    )"
+        "write_image(data, path, axes=\"zyx\")"
```
will-moore: This line is failing now, since the data above has been updated to 4 dimensions `czyx`.

docs/source/basic/write_image.ipynb
will-moore: This sentence "A zarr v3 group and store is created by `zarr.open_group()` below" still needs removing.

docs/source/basic/write_image.ipynb
jo-mueller: fixed in 5b4aa534b512b22b3e4f45e78e20b9837ac95bb9

will-moore:
The docs need to have this sentence removed from `write_image.ipynb` 
"A zarr v3 group and store is created by zarr.open_group() below".

I can open a separate PR if you want (since it's not coming from this PR)?

pennycuda:
Hi @jo-mueller and @will-moore ! We spoke the other day on my other PR to accept spatialdata zipped zarr stores. I went ahead and cloned Jo's branch and added that support https://github.com/hubmapconsortium/ome-zarr-py/tree/pennycuda/new-ome-api
Should I make a PR to @jo-mueller with these changes? Let me know how to proceed! I also made a new SpatialData branch based off of https://github.com/scverse/spatialdata/pull/1107 and these exciting API updates that I am going to share over on that PR 1107. https://github.com/hubmapconsortium/spatialdata/tree/pennycuda/zip-store-rework

Thank you both for your contributions and guidance! 

Penny

docs/source/basic/write_image.ipynb
```diff
@@ -8,54 +8,152 @@
...
+        "  scale={\"c\": 1.0, \"z\": 0.5, \"y\": 0.1, \"x\": 0.1},\n",
+        "  axes_units={\"z\": \"micrometer\", \"y\": \"micrometer\", \"x\": \"micrometer\"},\n",
+        ")\n",
+        "\n",
+        "multiscales = OMEZarrMultiscale(\n",
+        "    image=image,\n",
+        "    scale_factors=(2, 4, 8),\n",
+        "    method=\"resize\",\n",
+        "    channel_names=[\"DAPI\", \"GFP\"],  # optional\n",
+        "    channel_colors=[\"#00FFFF\", \"#FF00FF\"],  # optional\n",
```
70Gage70:
```suggestion
        "    channel_colors=[\"00FFFF\", \"FF00FF\"],  # optional\n",
```

As per [this](https://github.com/ome/ome-zarr-py/pull/515/changes/2d5b39e4ff3ddf5e7c5af242af3c07fc0ee339ff#diff-dbf24ab87100e7c77eb61cc16f71d378fa635270491922a3702af4466284d335R593)? I think this also fails silently downstream (I found this from napari-ome-zarr) if users add a leading `#` to `channel_colors`.

jo-mueller:
Hi @70Gage70 , thanks for the suggestion. I added a hard removal of any remaining`#`s in the parser function, so that one *could* pass the `#`, if one wanted. Mainly because one can then do this in an IDE (at least in VSCode):

https://github.com/user-attachments/assets/e420eb56-d05a-4b55-95c6-4af697447492

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,1044 @@
...
+                )
+
+        # warn about missing axes
+        for d in self.axes:
+            if d not in self.scale:
+                warnings.warn(
+                    f"Scale value not provided for axis '{d}'. "
+                    f"Using default scale of 1.0.",
+                    stacklevel=2,
+                )
```
melonora:
```suggestion
        if (scale_set := set(self.scale)) != (axes_set := set(self.axes)):
            if diff := scale_set.difference(axes_set):
                raise ValueError(
                    f"Scale contains invalid ax(i)(e)s: {diff}. Valid axes are: {axes_set}"
                )
            
            warnings.warn(
                f"Scale value not provided for ax(i)(e)s '{axes_set.difference(scale_set)}'. "
                f"Using default scale of 1.0.",
                stacklevel=2,
            )             
```
no need to loop, bit shorter code. Not really important as it is small anyway

ome_zarr/classes/image.py
melonora: also there is no need for the conditionals, just do `method = Methods(method or Methods.RESIZE)` in the end. I think that is equivalent.

ome_zarr/classes/image.py
melonora: given that the input is OMEZarrImage which is single scale yes.

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,1044 @@
...
+        else:
+            store_exists = True  # zarr.Group was passed in, so it exists
+
+            # Decide whether to write main image data
+        write_image_data = not store_exists or overwrite
+
+        if write_image_data:
+            # Delete existing store if overwriting
+            if overwrite and isinstance(group, str) and os.path.exists(group):
+                shutil.rmtree(group)
```
melonora, line 265: is it possible that there would be consolidated metadata? If so could this lead to issues?

jo-mueller:
@pennycuda sorry for missing your comment here! We will see to have this merged this week, so to contribute it's probably easiest to check out this branch and open it as a PR to main and rebase later (won't do any force pushes here anymore)

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,1041 @@
...
+            scale_factors=scale_factors,
+            method=method,
+        )
+
+        # build scales for each level based on the original image shape
+        # and the pyramid level shapes
+        scales = []
+        # image.scale is guaranteed to be a dict after NgffImage.__post_init__
+        image_scale = image.scale
+        assert isinstance(image_scale, dict)
```
melonora: hmm not typical to have assert in production code. If it is always certain that here we are dealing with a dict then I think we could just cast here. Otherwise I would prefer an error message giving more context to the user.

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,1041 @@
...
+            if d in SPATIAL_DIMS:
+                axes.append(Axis(name=d, type="space", unit=image.axes_units.get(d)))
+            elif d == "t":
+                axes.append(Axis(name=d, type="time", unit=image.axes_units.get(d)))
+            elif d == "c":
+                axes.append(Axis(name=d, type="channel", unit=image.axes_units.get(d)))
+            else:
+                axes.append(Axis(name=d, type="custom", unit=image.axes_units.get(d)))
+
+        self.metadata = MultiscaleV05(
```
melonora, line 228: I think if this class is made public, it should be in the docstring that this by default would create a MultiScaleV05.

ome_zarr/classes/image.py
melonora: Discussed in the ome-zarr-py meeting, consolidated metadata is outside the spec, though it might be interesting for performance on cloud. @kevinyamauchi, @joshmoore opinions? I know @joshmoore consolidated metadata brings an extra pain and for now maybe better not to support as it is experimental.

ome_zarr/classes/image.py
jo-mueller: 09388e7de08f47a6e9ee11690ef652a308fbf605

ome_zarr/classes/image.py
jo-mueller: d9bad78342dd6287ee49b7d0b07766a0cb954b33

ome_zarr/classes/image.py
jo-mueller: Follow-up issue: https://github.com/ome/ome-zarr-py/issues/591

will-moore (approved):
LGTM: let's get this in and then followup with any remaining issues we need...

PR #534, opened 2026-02-17
Support sharding

melonora:
This PR adds support back for all dask versions 2025.2.0 up to and including 2025.11.0. Furthermore, it adds support for dask versions 2026.3.0 and up and with that also adds the possibility for among others passing on sharding as arguments. The dask versions inbetween the mentioned versions are no longer supported as they required particular keyword arguments not used anymore.

Lastly, both different zarr formats being used as well as different dask versions (2026.3.0 and up versus below) have different support for arguments. Because of this, I did not think it reasonable to truly speak of deprecated arguments at this stage and did not add deprecation warnings. I rather add this information in the docstrings.


melonora:
yes, have some new incoming changes now with the new dask release. We can discuss during the meeting to review this PR and then after that I would do a rebase if thats ok. Just to prevent having to make changes later

tests/test_writer.py
```diff
@@ -50,6 +51,45 @@
     [{"scale": [1, 1, 8.0, 2.88, 2.88], "type": "scale"}],
 ]
 
+FORMAT_VERSIONS = [
+    pytest.param(FormatV03, id="V03"),
```
jo-mueller:
```suggestion
```

ome_zarr/writer.py
```diff
@@ -690,17 +693,13 @@ def _write_pyramid_to_zarr(
     # Set up common kwargs for da.to_zarr
     # zarr_array_kwargs needs dask 2025.12.0 or later
     zarr_array_kwargs: dict[str, Any] = {}
-    zarr_format = fmt.zarr_format
+    zarr_format = zarr_array_kwargs["zarr_format"] = fmt.zarr_format
```
jo-mueller, line 716: Uh, I didn't know this was allowed :D So that means fmt.zarr_format will have precedence over `zarr_array_kwargs`?

ome_zarr/writer.py
```diff
@@ -10,15 +10,19 @@
...
+from packaging.version import Version
 
 from .axes import Axes
 from .format import CurrentFormat, Format, FormatV01, FormatV02, FormatV03, FormatV04
 from .scale import Methods, Scaler
 from .types import JSONDict
 
 LOGGER = logging.getLogger("ome_zarr.writer")
+# If not 2026.3.0 it must be 2025.11.0 or lower. Name indicates kwargs only contain array kwargs in the dask version.
+DASK_ARRAY_KWARGS = Version(dask_version) >= Version("2026.3.0")
```
jo-mueller:
```suggestion
USE_DASK_ARRAY_KWARGS = Version(dask_version) >= Version("2026.3.0")
```

Suggestion for refactor so that it's clear we're looking at a boolean flag elsewhere

ome_zarr/writer.py
```diff
@@ -709,33 +708,47 @@ def _write_pyramid_to_zarr(
...
             # Use bytes-to-bytes codecs instead.
-            zarr_array_kwargs["compressor"] = options.pop("compressor")
+            zarr_array_kwargs["compressors"] = options.pop("compressor")
+    elif zarr_format == 2:
+        zarr_array_kwargs["dimension_separator"] = "/"
+
+    if axes is not None and zarr_format != 2:
+        zarr_array_kwargs["dimension_names"] = [
+            a["name"] for a in axes if isinstance(a, dict)
+        ]
```
jo-mueller, line 738: This *could* already become a conditional thing to do because 0.6 will not require `dimension_names` to be written anymore, I think.

ome_zarr/writer.py
jo-mueller: Also, maybe this could go into the `ome_zarr` init? 🤔

melonora: yeah I am doing some fixing up and refactoring as part of the rebase now:)

ome_zarr/writer.py
melonora: yes, that must be the case as the different versions are tied to specific zarr formats. E.g. FormatV04 is tied to zarr format 2 and not 3 so we cannot allow the user to overwrite this.

ome_zarr/writer.py
melonora: I will let others weigh in here a bit I think:)

ome_zarr/writer.py
melonora: refactored now

melonora:
@jo-mueller the codespell seems to be a docker authentication issue not due to any code change made.

melonora:
@jo-mueller could you try retriggering the codespell workflow? This could just be a temporary docker issue.. It seems unrelated to the code.

ome_zarr/writer.py
jo-mueller: Oh, for now this can totally stay here. As long as 0.6 isn't released there's no need to concern oneself with these possibilities. Just a note - feel to mark as resolved.

jo-mueller:
@melonora just tested on my end (Win 11), and I run into this issue that I've seen before, but I can't recall where; When I write some sample data on your branch, I get this error:

```python
ile c:\Users\johan\Documents\GitHub\ome-zarr-py\ome_zarr\writer.py:659, in write_image(image, group, scale_factors, method, scaler, fmt, axes, coordinate_transformations, storage_options, compute, **metadata)
    656 name = metadata.pop("name", None)
    657 name = str(name) if name is not None else None
--> [659](file:///C:/Users/johan/Documents/GitHub/ome-zarr-py/ome_zarr/writer.py:659) dask_delayed_jobs = _write_pyramid_to_zarr(
    660     pyramid,
    661     group,
    662     fmt=fmt,
    663     axes=axes,
    664     coordinate_transformations=coordinate_transformations,
    665     storage_options=storage_options,
    666     name=name,
    667     compute=compute,
    668     **metadata,
    669 )
    671 return dask_delayed_jobs

File c:\Users\johan\Documents\GitHub\ome-zarr-py\ome_zarr\writer.py:804, in _write_pyramid_to_zarr(pyramid, group, fmt, axes, coordinate_transformations, storage_options, name, compute, **metadata)
    802 # Computing delayed jobs if necessary
...
   1375     """
-> [1376](https://file+.vscode-resource.vscode-cdn.net/c%3A/Users/johan/Documents/GitHub/ome-zarr-py/~/AppData/Roaming/uv/python/cpython-3.12.12-windows-x86_64-none/Lib/pathlib.py:1376)     os.replace(self, target)
   1377     return self.with_segments(target)

PermissionError: [WinError 5] Zugriff verweigert: 'test.ome.zarr\\s4\\c\\0\\0\\0\\0\\0.b2f37e7e7a3146dc8f8be5eb9868ebac.partial' -> 'test.ome.zarr\\s4\\c\\0\\0\\0\\0\\0'
```

" Zugriff verweigert" means permission denied.

My sample code:

```python
import dask.array as da
import os
import shutil

array = da.random.random((1, 1, 64, 64, 64), chunks=(1, 1, 32, 32, 32))

if os.path.exists("test.ome.zarr"):
	shutil.rmtree("test.ome.zarr")
write_image(array, "test.ome.zarr")
```

Here's the strange part: If I run the code for writing (the last three lines) a few times, it sometimes works, and sometimes doesn't. My guess is that the dask scheduling sometimes (by chance) wants to write to the same shard concurrently and sometimes it it's already done with doing so when the next job wants to write 🤔 

*Edit*: Using `dask==2026.3.0` and `zarr==3.1.5`

melonora:
hehe I understand German and yeah I agree, that indeed seems to be the issue. This would honestly be challenging to fix in a performant fashion. Currently, you would have to specify the lock in dask as `True`. This can be done by specifying `mode="w"` in `to_zarr`. The problem is that this will set a lock on the whole array, rather than the shard, which defeats the whole speedup. Let me check if I can reproduce locally with your example and whether the rust implementation also gives this problem.

melonora:
@jo-mueller you even get this when not writing shards? This should not happen as shards by default are set to `None`. This issue typically only happened for me when actually writing shards. This should not happen when purely writing chunked data.

jo-mueller:
> you even get this when not writing shards?

Yes, I was only using the example above. Maybe we are running into this issue here? https://github.com/zarr-developers/zarr-python/issues/3522

The downsampling *does* use `map_blocks` which seems to be indicative of that problem in the referenced issue.

melonora:
I can't seem to reproduce it at all anymore. Can you try when you set mode to `w`? e.g. pass it on as kwarg

melonora:
Just for the record for others, @jo-mueller and I discussed and we found the issue. The problem is the "auto" default for chunks in zarr. This can create different chunks then the dask chunking scheme which then can cause multiple writes at the same time. This can be fixed by disallowing "auto" and having as default the dask chunks. It is not easy to test for this as usually the issue only occurs after multiple writes if at all (@jo-mueller could produce it consistently on his system, but I could not produce it at all while we both use windows).

melonora:
@jo-mueller ready for review. Please also play around a bit with the sharding while no chunks are specified.

jo-mueller:
@melonora Am I doing this correctly? 🙈 

```python
import dask.array as da
import os
import shutil

array = da.random.random((1, 1, 64, 64, 64), chunks=(1, 1, 8, 8, 8))

if os.path.exists("test.ome.zarr"):
	shutil.rmtree("test.ome.zarr")
write_image(array, "test.ome.zarr", storage_options={"shards": [1, 1, 2, 2, 2]})

---> [13](vscode-notebook-cell:?execution_count=13&line=13) write_image(array, "test.ome.zarr", storage_options={"shards": [1, 1, 2, 2, 2]})

File c:\Users\johan\Documents\GitHub\ome-zarr-py\ome_zarr\writer.py:671, in write_image(image, group, scale_factors, method, scaler, fmt, axes, coordinate_transformations, storage_options, compute, **metadata)
    668 name = metadata.pop("name", None)
    669 name = str(name) if name is not None else None
--> [671](file:///C:/Users/johan/Documents/GitHub/ome-zarr-py/ome_zarr/writer.py:671) dask_delayed_jobs = _write_pyramid_to_zarr(
    672     pyramid,
    673     group,
    674     fmt=fmt,
    675     axes=axes,
    676     coordinate_transformations=coordinate_transformations,
    677     storage_options=storage_options,
    678     name=name,
    679     compute=compute,
    680     **metadata,
    681 )
    683 return dask_delayed_jobs

File c:\Users\johan\Documents\GitHub\ome-zarr-py\ome_zarr\writer.py:770, in _write_pyramid_to_zarr(pyramid, group, fmt, axes, coordinate_transformations, storage_options, name, compute, **metadata)
    767 elif shards_opt is not None:
    768     # This ensures that shards are always divisible by chunks, which is a requirement.
    769     if chunks_opt != "auto":
--> [770](file:///C:/Users/johan/Documents/GitHub/ome-zarr-py/ome_zarr/writer.py:770)         chunks_opt = _retuple(chunks_opt, level.shape)  # type: ignore[arg-type]
    771     else:
    772         # Technically not needed as ultimately in this case dask chunks will correspond to shards.
    773         # Simply adding this warning here to make the user used to not using "auto".
    774         if chunks_opt == "auto":

File c:\Users\johan\Documents\GitHub\ome-zarr-py\ome_zarr\writer.py:1205, in _retuple(chunks, shape)
   1202 if isinstance(chunks, int):
   1203     return tuple([chunks] * len(shape))
-> [1205](file:///C:/Users/johan/Documents/GitHub/ome-zarr-py/ome_zarr/writer.py:1205) dims_to_add = len(shape) - len(chunks)
   1206 return (*shape[:dims_to_add], *chunks)

TypeError: object of type 'NoneType' has no len()
```

Dask version is `2026.3.0` 🤔

jo-mueller:
Cool, works like a charm!

jo-mueller (approved):

jo-mueller:
Written array is missing `dimension_names` in zarr array `zarr.json`; I use this code

```python
import dask.array as da
import os
import shutil

array = da.random.random((1, 3, 128, 128, 128), chunks=(1, 1, 32, 32, 32))

if os.path.exists("test.ome.zarr"):
	shutil.rmtree("test.ome.zarr")
write_image(
	array,
	group="test.ome.zarr",
	storage_options={"shards": [1, 1, 64, 64, 64]},
)
```

Gives me this when I open the image from the CLI like this:

```bash
ome_zarr view test.ome.zarr
```

image

will-moore (approved):
This looks very nice and is working well for me.

I get nice exceptions if I try something invalid like incompatible chunk/shard sizes, or try to write shards for OME-Zarr v0.4.

The only question I have is whether we want to mention sharding in the docs, since I think it could be of interest to some users who would not necessarily read the docstrings.

E.g. something like:

```
# storage_options depends on dask version. Please refer to write_image() docs for more details.
storage_options = {"chunks": (3,100, 100), "shards": (3, 200, 200)}
write_image(data, "with_shards.zarr", axes="cyx", storage_options=storage_options)
```

Thanks for all your effort on this @melonora

jo-mueller:
@will-moore I'll write up a documentation PR for this either as a PR to this one, or as a separate follow-up PR.

PR #544, opened 2026-03-04
refactor: write scales to zarr group s<idx>

jo-mueller:
Hi @will-moore ,

small refactoring PR to change the names of the zarr groups to which the multiscales are written from `0, 1, 2, ...` to `s0, s1, s2, ...`.


will-moore (approved):
Looking good to me. Anything else needed here?

PR #548, opened 2026-03-10
Use markdown and Jupyter notebooks in doc pages

### LINKED ISSUE #536: Give the documentation a glow-up?
With more functionality coming to this repo in the near future (i.e., class-based API #515 , ome-zarr 0.6, etc), I think we should think about giving the documentation pages a makeover to make it easier to find certain things. 

I would suggest a structure according to the [four types of documentation](https://docs.divio.com/documentation-system/):

- (API) Reference - already exists and autogenerated by sphinx
- Tutorials: Example little workflows  to accomplish some goal that a user can walk along and achieve as well I.e., how to write an image? how to write labels? How to create Wells/Plates etc.?
- How-Tos: Slightly more complex project for more advanced users aimed to turn learners into users. For instance, this could be about how to formulate coordinate transformations to write to the ome-zarr metadata.
- Explanations: Could be a reference to other resources about related things (i.e., ngff specification, zarr project, dask, S3 storage, etc).

On a functional level, I would suggest to move from Restructured text more to markdown? Partially because I'm more familiar with it (😜) but I also like the rich formatting options of frameworks like [MyST](https://mystmd.org).

jo-mueller:
Fixes #536


jo-mueller:
@will-moore this would be my take on how I'd structure the docs going forward.

I essentially broke all the examples from the `python.rst` file into separate jupyter notebooks that are actually executed on build 👍

jo-mueller:
> Ah - is this the output from the code above?

Yep and I think it kind of makes sense. If `compute=False` is passed to the function, the writer will not execute but return a list of delayed objects. If `compute=True`, everything is written, all jobs are finished so the list is empty.

will-moore (approved):
Looks great, thanks! 👍

Issue #501, opened 2025-10-15
write_image() API limitations


will-moore:
Expanding on list at https://github.com/ome/ome-zarr-py/issues/407...

 - write_image() only downsamples in 2D (x and y) 👎 #262 
 - No control of downsample factor.
 - Not easy to write pixel sizes. Scale starts at [1, 1, 1, 1, 1]. If you want to supply pixel sizes, the only way to do this is via `coordinate_transformations` but it's a real pain to know how long to make the list. See #481. Also you have to fully specify `axes`.
 - Axes created automatically: 'type' inferred by name. No units.

jo-mueller:
@will-moore I think the issue raised here are essentially all solved now, right? Since downsampling factors are now calculated and validated internally  (#516) and the scale and scale factors can be passed as arguments to the `write_image` function (#553), one now has complete control over the downsampling process.

Also, `axes` can simply be passed as a list of strings (i.e., `["z", "y", "x"]`) and it's valid.

PR #553, opened 2026-03-11
feat: add `scale` argument to entrypoints

### LINKED ISSUE #532: Pass `scale` instead of coordinate transformations
In the current writing logic, the user has to calculate the `coordinate_transformations` themselves for each pyramidal layers before passing them on to the `write_image` function. I think it's much easier to simply pass a `scale` argument as a per-axis dict (i.e., `{"z": 0.5, "y": 0.2, "x": 0.2}`) and then just auto-calculate all the coordinate transformations for every scale level. 

The default behavior for unpassed coordinatetransformations can stay as it is.

Working plan:

- add `scale` argument to kwargs of `write_image`, `write_labels` functions
- deprecate `coordinate_transformations` argument of these functions

cc @will-moore 
cc @melonora : I think this is also orthogonal to your efforts regarding improving passage of arguments to zarr

jo-mueller:
Fixes https://github.com/ome/ome-zarr-py/issues/532
Partially solves https://github.com/ome/ome-zarr-py/issues/501

## Key features

This PR adds a new kwarg `scale` to all the writer functions (i.e., `write_image`, `write_labels`, etc) that will replace passing a pre-generated list of `coordinateTransformations` to the function. I marked the later keyword as deprecated; for now, it will be used instead of `scale` under the hood if passed.

## TODO

- [x] Document
- [x] Tests

### DIFF
--- a/ome_zarr/axes.py
+++ b/ome_zarr/axes.py
@@ -1,16 +1,72 @@
 """Axes class for validating and transforming axes"""
 
+import warnings
 from typing import Any
 
 from .format import CurrentFormat, Format
 
 KNOWN_AXES = {"x": "space", "y": "space", "z": "space", "c": "channel", "t": "time"}
+KNOWN_SPATIAL_UNITS = [
+    "angstrom",
+    "attometer",
+    "femtometer",
+    "picometer",
+    "nanometer",
+    "millimeter",
+    "micrometer",
+    "centimeter",
+    "decimeter",
+    "meter",
+    "kilometer",
+    "megameter",
+    "hectometer",
+    "gigameter",
+    "terameter",
+    "petameter",
+    "exameter",
+    "inch",
+    "foot",
+    "mile",
+    "yard",
+    "parsec",
+    "yoctometer",
+    "yottameter",
+    "zeptometer",
+    "zettameter",
+]
+
+KNOWN_TEMPORAL_UNITS = [
+    "attosecond",
+    "centisecond",
+    "day",
+    "decisecond",
+    "exasecond",
+    "femtosecond",
+    "gigasecond",
+    "hectosecond",
+    "hour",
+    "kilosecond",
+    "megasecond",
+    "microsecond",
+    "millisecond",
+    "minute",
+    "nanosecond",
+    "petasecond",
+    "picosecond",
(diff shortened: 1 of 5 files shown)


ome_zarr/axes.py
```diff
@@ -23,6 +79,18 @@ def __init__(
         elif fmt.version in ("0.1", "0.2"):
             # strictly 5D
             self.axes = self._axes_to_dicts(["t", "c", "z", "y", "x"])
+
+        for ax in self.axes:
+            if axes_units and ax["name"] in axes_units:
+                ax["unit"] = axes_units[ax["name"]]
```
will-moore, line 85: the `self.axes` (from `axes` argument above) may already have `unit` set, in which case we should simply validate that. I guess it's possible to provide 2 different values for unit in this way; within the `axes` dict and with `axes_units`. Do we need both?

jo-mueller: Oh, so `axes` is intended (or capable) to accept a complete dict with units, longname, name, etc? In that case I guess I would simplify the axes keyword to only receive the axis names in the correct order (i.e., `"zyx"` or `["z", "y", "x"]`) so that we have a clear separation of concerns between all keyword arguments.

I.e., in other words, I would prefer having more keywords with distinct purpose rather than supercharging fewer kwargs with many different purposes we then need to parse 😬

ome_zarr/writer.py
```diff
@@ -767,6 +819,13 @@ def _write_pyramid_to_zarr(
         # shapes = [data.shape for data in delayed]
         coordinate_transformations = fmt.generate_coordinate_transformations(shapes)
 
+        if coordinate_transformations:
+            for idx, transform in enumerate(coordinate_transformations):
```
will-moore: `idx` isn't used so you don't need to `enumerate()`

jo-mueller: Huh. Feel like precommit should've caught that 🤔 271df3fe4193a274e6096631a1f3f91057b0bbf0

will-moore (approved):
Works as expected. Looks good, thx 👍

jo-mueller:
Awesome. Mind if I merge? Will then rebase #515

PR #557, opened 2026-03-13
BREAKING CHANGE: Deprecate writing v01, v02 and v03

### LINKED ISSUE #556: Deprecate writing 0.1/0.2/0.3
To go forward with #515 , I propose to deprecate the functionality to write ome-zarr images of version 0.1, 0.2 and 0.3. THey are not widely used anymore and we can still make sure that they can be read and accessed nonetheless 👍 I think the roadmap to do so would be *fairly* simple? Raise an exception if format 0.1/0.2/0.3 is passed to the writer and remove v01/v02/v03 from the tests?

The question is: Should this be a soft deprecation? (I.e., do we add a warning that we will remove it in the future? Or do we consider it old enough to just remove the feature alltogether?

jo-mueller:
Fixes #556

### DIFF
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -14,7 +14,7 @@ from dask.graph_manipulation import bind
 from numcodecs import Blosc
 
 from .axes import Axes
-from .format import CurrentFormat, Format, FormatV01, FormatV02, FormatV04
+from .format import CurrentFormat, Format, FormatV01, FormatV02, FormatV03, FormatV04
 from .scale import Methods, Scaler
 from .types import JSONDict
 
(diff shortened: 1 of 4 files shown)


jo-mueller:
Huh, actually the coverage comes out ok here. It's just the lines of the deprecation warning that are not covered, which I'd think would be acceptable?

jo-mueller:
@will-moore just for the formalities - could you hit the `Approve` button here?

will-moore (approved):
As discussed: agreed that we don't need to support writing of v0.1-v0.3 now.

Changes look good, thx

PR #567, opened 2026-03-31
chore: fix defaults for scale factors

jo-mueller:
@will-moore very small fix that I discovered while working on #515 . In one of the entrypoints (`write_labels`), the `Scaler` class was still used as a default, which it shouldn't be.


will-moore (approved):
LGTM 👍

PR #568, opened 2026-03-31
feat: do not delay writing metadata

jo-mueller:
@will-moore Another small PR in the wake of #515 :) I realized that some tests were failing there for an unrelated reason: In #515, I use the existing code in `_write_pyramid_to_zarr` to write the metadata, but it's effectively later overwritten by the metadata dump from the ome-zarr-models-py classes.

*However*, if writing the metadata is *delayed* inside the `_write_pyramid_to_zarr` function, it will overwrite the metadata dump from ozmp 🤦‍♂️ . Simple fix: I am chosing to always write metadata right away here. The performance-impact of writing a bit of json should be manageable.

It's not strictly related to #515 so I'm sending it as a breakout-PR.

### DIFF
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -774,16 +774,9 @@ def _write_pyramid_to_zarr(
     if coordinate_transformations is not None:
         for dataset, transform in zip(datasets, coordinate_transformations):
             dataset["coordinateTransformations"] = transform
-    if not compute:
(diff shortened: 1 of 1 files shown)


will-moore:
Looks good. Do we care whether this gets merged before / after #515?

jo-mueller:
It's already part of #515, so merging this ahead will make review there a wee bit easier.

will-moore (approved):
Looks good, tests green etc.

PR #569, opened 2026-04-09
Document sharding

jo-mueller:
Based on #534 - adds documentation :)

### DIFF
--- a/docs/source/explanation/zarr_concepts.md
+++ b/docs/source/explanation/zarr_concepts.md
@@ -30,6 +30,14 @@ import zarr
 arr = zarr.zeros((10000, 10000), chunks=(1000, 1000))
 ```
 
+### Sharding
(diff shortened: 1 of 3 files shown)


docs/source/explanation/zarr_concepts.md
```diff
@@ -30,6 +30,14 @@ import zarr
 arr = zarr.zeros((10000, 10000), chunks=(1000, 1000))
 ```
 
+### Sharding
+Zarr now provides the option to bundle multiple chunks into a so-called shard file (see [sharding proposal](https://zarr.dev/zeps/accepted/ZEP0002.html)).
+On local file systems, this reduces the number of files written to disk, which improves several aspects of data handling (i.e., data transfer speeds).
```
melonora: perhaps also worthwhile adding here that local filesystems can't handle to many binary files and that sharding aims to alleviate that.

will-moore (commented):
Just reading through https://github.com/jo-mueller/ome-zarr-py/blob/3ec9eba36d91698190465d553b039fa6504c7df3/docs/source/advanced/sharding.ipynb
Looks good 👍

PR #582, opened 2026-05-18
Introduce image class v06

jo-mueller:
Relevant for https://github.com/ome/ome-zarr-py/issues/595

Built on top of #515. This PR adds writing and reading of 0.6 metadata to the capabilities of the `NgffMultiscales` class. Compared to #515, it's a fairly minor change.

### TODOs:

- [x] https://github.com/ome-zarr-models/ome-zarr-models-py/pull/433: Support reading all ome-zarr versions by internal coercion to 0.6 Multiscale metadata
- [x] https://github.com/ome-zarr-models/ome-zarr-models-py/pull/434: write versions 0.4 and 0.5: Metadata is internally represented as 0.6, which requires down conversion to 0.4/0.5 on write
- [x] New release in ome-zarr-models-py to be pinned here
- [x] update legacy reader to ingest and convert 0.1/0.2/0.3 metadata to 0.6 rather than 0.5
- [x] Document how to populate `NgffMultiscales.metadata.coordinateTransformations` field to build a proper transform graph


jo-mueller:
@will-moore @kevinyamauchi @melonora ready for testing :) 

I added a demo notebook to the docs to explain how writing additional transforms and coordinate systems work.

docs/source/advanced/transforms/add_transforms_to_multiscales.ipynb
```diff
@@ -0,0 +1,168 @@
...
+  "cells": [
+    {
+      "cell_type": "markdown",
+      "id": "56309244",
+      "metadata": {},
+      "source": [
+        "# Adding transformations to multiscale images\n",
+        "\n",
+        "Adding transformations to individual multiscale images is a way to enrich the metadata associated with the image.\n",
+        "THis can be useful to inform implementations on how to view or interpret the image data,\n",
```
will-moore: Typo `This`

docs/source/advanced/transforms/add_transforms_to_multiscales.ipynb
```diff
@@ -0,0 +1,168 @@
...
+        "    ],\n",
+        "    \"input\": {\"name\": \"physical\"},\n",
+        "    \"output\": {\"name\": \"world\"}\n",
+        "}\n",
+        "\n",
+        "world_cs = {\n",
+        "    \"name\": \"world\",\n",
+        "    \"axes\": [\n",
+        "        {\"name\": \"x\", \"type\": \"space\"},\n",
+        "        {\"name\": \"y\", \"type\": \"space\"}\n",
```
will-moore: Axis order should be `yx` not `xy`

docs/source/advanced/transforms/add_transforms_to_multiscales.ipynb
```diff
@@ -0,0 +1,168 @@
...
+      "cell_type": "code",
+      "execution_count": 9,
+      "id": "f0a56122",
+      "metadata": {},
+      "outputs": [],
+      "source": [
+        "ms = OMEZarrMultiscale(\n",
+        "    image=image,\n",
+        "    coordinateTransformations=[rotation_transform],\n",
+        "    coordinate_systems=[world_cs],\n",
```
will-moore, line 141: Having these two arguments use different naming styles is inconsistent. `coordinateTransformations` and `coordinate_systems`. Either use camelCase or snake_case for both.

docs/source/advanced/transforms/add_transforms_to_multiscales.ipynb
```diff
@@ -0,0 +1,168 @@
...
+      "execution_count": 8,
+      "id": "8177bd16",
+      "metadata": {},
+      "outputs": [],
+      "source": [
+        "rotation_transform = {\n",
+        "    \"type\": \"rotation\",\n",
+        "    \"rotation\": [\n",
+        "        [0, -1],\n",
+        "        [1, 0]\n",
```
will-moore: When opening the resulting image in napari-ome-zarr (with v0.6 support) the resulting image doesn't appear rotated with this transform. It would be nicer if it used a matrix like
```
[0.7, 0.7],
[-0.7, 0.7]
```
So we get an obvious rotation:

Image

docs/source/advanced/transforms/add_transforms_to_multiscales.ipynb
jo-mueller: 90a714b4559c9c506d19f5e183869cadb19535e4

docs/source/advanced/transforms/add_transforms_to_multiscales.ipynb
jo-mueller: 9806829af9fa15b4d97c340e629bb252677bdfa0

docs/source/advanced/transforms/add_transforms_to_multiscales.ipynb
jo-mueller: 9806829af9fa15b4d97c340e629bb252677bdfa0

docs/source/advanced/transforms/add_transforms_to_multiscales.ipynb
jo-mueller: Good idea!

jo-mueller: I also added the screenshot to the docs

will-moore: Thanks for that - However, I realised I cheated a bit here. The `[[0.7, 0.7], [-0.7, 0.7]]` matrix that I used to generate the screenshot was what I hand-edited in the `zarr.json` *after* creating the sample. When I tried to run the code with those values, I got a `pydantic` validation error that it wasn't a proper rotation.

```
  Value error, Provided matrix is not a pure rotation matrix: [[ 0.7  0.7]
 [-0.7  0.7]] [type=value_error, input_value={'type': 'rotation', 'rot...put': {'name': 'world'}}, input_type=dict]
```

 I needed to use much more accurate values for this to work - anything less than this failed:
```
        [0.707107, 0.707107],
        [-0.707107, 0.707107]
```

The original numbers (taken from https://ome.github.io/ome-ngff-validator/?source=https://livingobjects.ebi.ac.uk/idr/zarr/test-data/v0.6.dev4/idr0050/4995115_output_to_ms.zarr/4995115_cropped_700_700_400_400_rot45.zarr) were:
```
        [0.7071067811865476, 0.7071067811865475],
        [-0.7071067811865475, 0.7071067811865476]
```

jo-mueller: In the notebook, the rotation matrix is derived from `np.pi/4` in the current version, so that should be precise enough :)

will-moore:
Something's not right about the image.png in the notebook.

```
$ sphinx-build -b html docs/source/ docs/build/html
...
/Users/wmoore/Desktop/ZARR/ome-zarr-py/docs/source/advanced/transforms/add_transforms_to_multiscales.ipynb:100005: WARNING: image file not readable: advanced/transforms/attachment:image.png [image.not_readable]
...
```

Screenshot 2026-06-25 at 17 16 52

pyproject.toml
```diff
@@ -26,7 +26,7 @@ dependencies = [
     "scikit-image>=0.19.0",
     "toolz",
     "rangehttpserver",
-    "ome-zarr-models >= 1.6",
+    "ome-zarr-models==1.8.0rc0",
```
will-moore, line 29: So, I guess we're just waiting on either a release of ozmp or switch to an alternative package?

jo-mueller: I mean...rc0 is available on Pypi and the functionality used from there is very unlikely to change, so I think this needn't block us?

jo-mueller:
> Something's not right about the image.png in the notebook.

Fixed in 74a23d653a1e664712d8d9e8a453fdf30b9b2c0d

will-moore:
This is looking good 👍 

Possibly for a follow-up PR...
I wondered whether we could make it easier for users by not asking them to provide the "output" coordinateSystem, in this case the "world".
We already know the axes (same as the "intrinsic" coordinateSystem, unless we're dropping/adding/re-ordering axes) and the output.name is given in the transform, so we could auto-generate the `world_cs`, instead of requiring duplication of that info.

NB: I actually wondered whether it is *invalid* in the spec to specify different axes order or even different `unit` for the world, if all we're doing is "rotation"? There's currently no validation, so I can give it any unit etc.

```
data = np.random.rand(256, 256)
image = OMEZarrImage(data=data, axes="yx", name="My Image",
                     axes_units={"y": "micrometer", "x": "micrometer"})
rotation_transform = {
    "type": "rotation",
    "rotation": [
        [np.cos(np.pi/4), -np.sin(np.pi/4)],
        [np.sin(np.pi/4), np.cos(np.pi/4)]
    ],
    "input": {"name": "intrinsic"},
    "output": {"name": "world"}
}

world_cs = {
    "name": "world",
    "axes": [
        {"name": "x", "type": "space", "unit": "micrometer"},
        {"name": "y", "type": "space", "unit": "micrometer"}
    ],
}

ms = OMEZarrMultiscale(
    image=image,
    coordinate_transformations=[rotation_transform],
    coordinate_systems=[world_cs],
    default_coordinate_system_name="intrinsic"
    )
```

jo-mueller:
> We already know the axes (same as the "intrinsic" coordinateSystem, unless we're dropping/adding/re-ordering axes) and the output.name is given in the transform, so we could auto-generate the world_cs, instead of requiring duplication of that info.

Yeah I was thinking about it and then decided against it because there could be weird transforms that add or permute axes, in which case the axes info from the output coordinate system were less clear. But I agree, it would make the life of users easier not having to think about this aspect and we could still make it more explicit in the future if someone has a usecase that would be prohibited by that API.

jo-mueller:
> NB: I actually wondered whether it is invalid in the spec to specify different axes order or even different unit for the world, if all we're doing is "rotation"? There's currently no validation, so I can give it any unit etc.

Also, about this: At the end of the day `cyx` are just axis names without any realy meaning attached and only RFC4 will change that. In a post RFC4-world, one would have to ensure that the anatomical orientation of a coordinate system after a 90° rotation is still in sync with the input coordinate system.

jo-mueller:
So....should we merge and get on with #583? 🤩

jo-mueller:
@will-moore one more thing I'm thinking of adding here: If the `OMEZarrMultiscale` class is used to also write `coordinates` or `displacement` fields, then the `"type"` field in the axes metadata must be inferred correctly. I am thinking of adding an new keyword argument to pass this parameter explicitly to the `OMEZarrImage` class. The kwarg would default to `None` and we infer the type from the axes name as we currently already do it:

- `t` -> `"type": "time"`
- `c` -> `"type": "channel"`
- `z/y/x` -> `"type": "space"`

The new kwarg could then look something like this:

```
axes_types = {"c": "displacements", "z": "space", "y": "space", "x": "space"}
```

will-moore:
@jo-mueller That looks consistent with other params for defining axes 👍

will-moore:
Maybe not due to this PR, but I'm getting an error when building/running the docs.
`/Users/wmoore/Desktop/ZARR/ome-zarr-py/docs/build/html/reports/basic/write_image.err.log` complaining about `axes="zyx"` at the last line, when it should be `axes="czyx"`

`ValueError: axes length (3) must match number of dimensions (4)`

```
write_image(data, path, axes="zyx", fmt=FormatV04()) 
```

jo-mueller:
@will-moore fixed 👍

will-moore:
Docs warning for version number when writing transforms etc looks good.
And code warning also makes sense if you try it... 👍 

```
python test_v06_image.py 
/Users/wmoore/opt/anaconda3/envs/omeroweb_zarrv3/lib/python3.12/site-packages/ome_zarr_models/v06/multiscales.py:76: UserWarning: Coordinate transformations defined in multiscales > coordinateTransformations can currently not be converted to v0.5, as they are not supported in this version.
  return self._to_v05()
```

will-moore (approved):
Thanks - looking good 👍

PR #583, opened 2026-05-18
Introduce scene class

jo-mueller:
Based on #515

This is a first stab at an ome-zarr-py implementation of the [scene metadata/layout](https://ngff.openmicroscopy.org/specifications/dev/index.html#scene-metadata) that will be part of 0.6. It follows closely and builds upon the design principles of the `NgffMultiscales` class that's introduced in #515 and mirrors the design of the `NgffHCSPlate` I propose in https://github.com/ome/ome-zarr-py/pull/570

## TODOs:

- [x] Add tests
- [x] Add documentation on how to update/append scenes
- [x] Keep track internally of underlying graph structures, i.e. in an `nx.DiGraph` instance?


jo-mueller:
@kevinyamauchi ready for beta-testing, I would say. Still missing:

- `OMEZarrImage` needs another keyword to pass the `axis_type` (i.e., channel, displacement). Right now we infer these from the `axes` kwarg (i.e., `czyx`), but we 'll need more flexibility once https://github.com/ome/ngff-spec/pull/145 is merged (displacements require axis type `"displacement"`
- `coordinateTransformations/whatever` are currently not parsed into the class on read, which might lead to trouble when building the scene graph and it includes coordinates or displacements
- waaaaay more testing

jo-mueller:
Also: needs a rebase

ome_zarr/writer.py
```diff
@@ -780,12 +780,11 @@ def _write_pyramid_to_zarr(
...
-    _axes = _get_valid_axes(len(pyramid[0].shape), axes, axes_units=axes_units, fmt=fmt)
-    dims = _extract_dims_from_axes(_axes)
 
-    # Normalize `scale` so every axis in `dims` is represented.
-    # Missing axes are allowed and default to 1.0.
-    scale = {d: scale.get(d, 1.0) for d in dims}
+    # make sure every axis is represented in `scale`;
+    # coerce to 1.0 if not provided
+    # but don't allow missing axes to avoid silent errors
+    scale = {d: scale.get(d, 1.0) for d in axes}
```
kevinyamauchi, line 786: I'm not sure I understand the comment. How are missing axes disallowed?

Also, I see that this was the previous behavior, so I see the value in keeping the behavior consistent, but I think we should be cautious about coercing values. I think such coercion can hide genuine mistakes.

pyproject.toml
```diff
@@ -14,7 +14,7 @@ description="Implementation of images in Zarr files."
 license = "BSD-2-Clause"
 license-files = ["LICENSE"]
 
-requires-python = ">3.11"
+requires-python = ">=3.12, <4.0"
```
kevinyamauchi: 👀 Python 4?

tests/test_scene.py
```diff
@@ -0,0 +1,169 @@
+import pathlib
+
+import numpy as np
+import pytest
+import zarr
+
+from ome_zarr import OMEZarrImage, OMEZarrMultiscale, OMEZarrScene
+
+
+class TestScene:
```
kevinyamauchi: I think it would be helpful to add some comments to these tests to explain the intent. In particular a docstring-like comment for each test explaining what the goal is and a single-line comment for the `assert` statements explaining what they are testing.

ome_zarr/classes/scene.py
```diff
@@ -0,0 +1,393 @@
...
+    CoordinateSystem,
+)
+from ome_zarr_models.v06.scene import SceneAttrs
+from pydantic import TypeAdapter
+from zarr.storage import StoreLike
+
+from .image import OMEZarrMultiscale
+
+
+class OMEZarrScene:
```
kevinyamauchi, line 19: I think this class does a good job parsing the metadata and exposing the coordinate systems and images. Is that the goal of the class? Or is there an intention to do more (e.g., get a collection of images + transforms for a given coordinate system)?

kevinyamauchi (commented):
I think this is going in a good direction @jo-mueller ! I have left some comments below. 

My main feedback is that I think it would be helpful to sketch out a few use cases for the OMEZarrScene class so we can reason about the necessary API. For example, I could imagine it could be useful for users to be able to get the transforms for all images in a given coordinate system.

pyproject.toml
jo-mueller: Copied over from [here](https://github.com/clbarnes/transformnd/blob/main/pyproject.toml). I played around a bit with the requirements because I was having trouble making tests pass 😅

ome_zarr/classes/scene.py
jo-mueller: > I think this class does a good job parsing the metadata and exposing the coordinate systems and images. Is that the goal of the class? 

Yes, absolutely! Also, it is currently hidden in a private attribute (`scene._graph`), which contains the built transformnd graph, which could be very useful in other applications to

- pass a coordinate system and an image (i.e., an `OMEZarrMultiscale`) and expose transforming image -> coordinate_system_X somehow, and if it's just a tutorial on how to do that. For scenarios like displacements/coordinates, this would be extremely valuable for many users of the scene.
- In napari, one could load the scene in a widget, pull all the coordinate systems into a list and select the desired coordinate system in a dropdown. The graph traversal would then return the correct affine (if the transforms are that simple) and then apply it to all layers from the scene.
- Generally, in napari it would be super practical to load the scene into an `OMEZarrScene` object and write something like a `to_napari_layer` around the `OMEZarrMultiscale` class. That way, we can centralize a lot of the parsing behavior and make our lives easier in the future.

But I'm still pondering a bit on how that could look like, API-wise. It's probably best to put this in documentation first and then see what emerges as useful.

ome_zarr/writer.py
jo-mueller: It's a bit hidden, but in the `OMEZarrImage` class, this check is done:

```python
        # validate and normalize scale dict
        if (scale_set := set(self.scale)) != (axes_set := set(self.axes)):
            if diff := scale_set.difference(axes_set):
                raise ValueError(
                    f"Scale contains invalid ax(i)(e)s: {diff}. Valid axes are: {axes_set}"
                )

            warnings.warn(
                f"Scale value not provided for ax(i)(e)s '{axes_set.difference(scale_set)}'. "
                f"Using default scale of 1.0.",
                stacklevel=2,
            )
```

so that you'd get a warning if a scale value has been pre-set that may not have been passed as part of the scale argument. This coercion in writer.py is always shadowed by that in the `OMEZarrImage` class, so we might as well remove it 🤔

jo-mueller:
@kevinyamauchi thanks for the feedback!

> For example, I could imagine it could be useful for users to be able to get the transforms for all images in a given coordinate system.

That's definitely a use case I have in mind, but I'm not entirely sure how to expose it. In this implementation, I am relying mostly on transformnd to do the magic of traversing the graph. The two options I see towards exposing this more to users would be:

- Make the `_graph` attribute public, which is essentially the transformnd graph object, which in itself is built on a networkx graph that would be stored under `_graph.graph`. But the transformnd object also has a method `get_sequence(cs_1, cs_2)` that does the graph traversal for you. Exposing the transformnd graph would then require users to look up how to interact with it at transformnd to traverse graphs/apply transforms
- Provide a `traverse_scene(...)` mmber function that would wrap the transformnd logic. The tricky part is then to construct an API that takes and returns what one would expect, which I'm afraid could come down to reverse-engineering a lot of transformnd functionality. I.e., should the result be an affine matrix? What if the resulting transform cannot be serialized into a single affine? Should it be a (sequence of) transformation dictionaries akin to what the spec writes out?

cc @clbarnes

jo-mueller:
Continued over at #612

PR #584, opened 2026-05-21
Update docs for metadata parameter of write_image

keller-mark:
The `**` for the `**metadata` kwarg was confusing to me. It wasn't initially apparent to me via the type anotations or docstring whether I should do `write_image(omero={})` vs `write_image(metadata={"omero":{}})`. Only the latter works.

```py
write_image(
    image=img_arr,
    # other args here
    metadata={
        "omero": {
            "name": img_name,
        }
    }
)
```


jo-mueller:
@keller-mark I think this looks good, but since the merge of #515 you'll need to rebase. OR check out the new API introduced there, which should make writing the omero metadata a tad easier!

PR #587, opened 2026-06-03
Allow zipped zarr stores as an input for SpatialData

pennycuda:
This PR goes with the following PR to SpatialData that adds the functionality to read zipped zarr stores in as SpatialData objects. 
https://github.com/scverse/spatialdata/pull/1141

This was motivated by [HuBMAP Consortium's](https://hubmapconsortium.org/) desire to reduce the number of files in our filesystem while maintaining our requirement that our output files be easily readable by our users. HuBMAP has already built other infrastructure to support zipped zarr stores for AnnData objects as well.

Please let me know what changes I need to make to get this PR merged; I am happy to work with you all!

### DIFF
--- a/ome_zarr/io.py
+++ b/ome_zarr/io.py
@@ -9,7 +9,8 @@ from urllib.parse import urljoin
 
 import dask.array as da
 import zarr
-from zarr.storage import FsspecStore, LocalStore, StoreLike
+from zarr.core.group import Group as ZarrGroup
+from zarr.storage import FsspecStore, LocalStore, StoreLike, ZipStore
 
 from .format import CurrentFormat, Format, detect_format
 from .types import JSONDict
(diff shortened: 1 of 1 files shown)


will-moore:
Hi @pennycuda - thanks for the PR,

We are actually moving away from the `io` functionality in ome-zarr-py at the moment. There are a lot of issues such as the ones you're seeing when we try to wrap the zarr-python group & store creation, and that code shouldn't be needed anymore.

We have already stopped recommending the use of `parse_url()` for writing - instead you can pass a Zarr Group to `write_image(data, group,...)`.

We are close to merging a new API for reading and writing (#515) which will mean we should be able to fully deprecate the usage of `parse_url()` and `io.py`.

It would be great if you could have a look at that PR and check that you can work with Zip stores - I don't think we've explicitly been testing that cc @jo-mueller (NB: it's easier to read the updated docs if you build them rather than reading the notebooks).

 Thanks!

pennycuda:
@will-moore Thank you for your response! I will check out that PR and see if it works with my SpatialData updates for zip stores then.

pennycuda:
Hi again @will-moore ! I have a couple of updates from this morning. My colleague @keller-mark found a SpatialData PR that is much closer to being merged than mine but does not currently work with zipped stores because of it's ome-zarr-py IO usage. https://github.com/scverse/spatialdata/pull/1107
I am now working off of this SpatialData branch, and I pulled in @jo-mueller branch from #515 for ome-zarr-py which also does not work with `ZipStore`s as is. I will provide an update once I can get everything working together. 

Thank you! 
Penny

jo-mueller:
Hi @pennycuda thanks a lot for working on this, reading from Zipstores is definitely a very much desired featured here. We'll merge #515 as soon as possible so that contributors have an easier time finding the right development head 👍

melonora:
Hi @pennycuda, thanks for the PR. I would also like to point you to this: https://github.com/zarr-developers/zarr-python/pull/3612. This allows users to move specific zarr groups to any destination store. Also, avoiding the whole decoding / encoding step (copying raw bytes). It took a while to get this in, but we would take this particular method into account for SpatialData in order to have it working as efficiently as possible.

This might be useful for when people are submitting data to HuBMAP.

pennycuda:
Thanks @melonora for pointing me to that! I will look at those changes to see how they would pertain to my SpatialData changes. I'm closing this PR bc I just opened another one #587 for the new image class that was merged into ome-zarr-py yeterday. You might hear from me soon on another github thread. :)

PR #590, opened 2026-06-10
Include translations in multiscales

### LINKED ISSUE #403: `coordinateTransformations` generated for 0.4 are scale-only
The most common methods of image downsampling result in a translation of the downsampled image, but this code for generating `coordinateTransformations` metadata only returns scale transformations, which will be incorrect for almost all multiscale pyramids.

https://github.com/ome/ome-zarr-py/blob/56f72b06d4912ba5156fe54f913d19df895b9e9e/ome_zarr/format.py#L260-L271

Suggested fix:  generate translation transforms

will-moore:
Fixes #403.

This is on top of #515. Can rebase once that's merged.

Encouraged by discussion at https://github.com/ome/ngff/issues/89#issuecomment-4657609621, this adds translation to the dataset coordinateTransformsations generated on downsampling.

Need to add/update tests...

cc @jo-mueller

### DIFF
--- a/ome_zarr/classes/image.py
+++ b/ome_zarr/classes/image.py
@@ -180,6 +180,11 @@ class OMEZarrMultiscaleBase:
                 }
             )
 
+        translations = [
+            {d: (scale[d] / 2) - (scales[0][d] / 2) for d in image.axes}
+            for scale in scales
+        ]
+
         # Create Image instances for each pyramid level
         images = []
         datasets = []
@@ -201,6 +206,10 @@ class OMEZarrMultiscaleBase:
                         Scale(
                             type="scale",
                             scale=list(level_scale.values()),
+                        ),
+                        Translation(
+                            type="translation",
+                            translation=list(translations[idx].values()),
                         ),
                     ),
                 )
--- a/tests/test_writer.py
+++ b/tests/test_writer.py
@@ -298,7 +298,7 @@ class TestWriter:
             cts = node_metadata["multiscales"][0]["datasets"][level][
                 "coordinateTransformations"
             ]
-            assert len(cts) == 1
+            assert len(cts) == 2
             transf = cts[0]
             assert transf["type"] == "scale"
             for d in axes:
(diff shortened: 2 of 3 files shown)


almarklein:
BTW, with the 'nearest' downsample method, I imagine it basically does `a[::2]` for all dimensions; the new samples are exactly on top of the samples in the previous layer. This gives the following scheme, and the offsets should be zero:
```
             _______ _______ _______ _______
            |___o___|___o___|___o___|___o___|   level 0
         _______________ _______________ ___
        |_______o_______|_______o_______|___    level 1
 _______________________________ ___________
|_______________o_______________|___________    level 2
```

*Edit: nevermind, with `nearest` it does not matter which of the two pixels is taken, so in this case the offset can be either, I guess. That said, if you first smooth the data and then sample it with `a[::2]` the above schema *does* apply, because you've moved information from the neighbours to the sampled pixels.*

ome_zarr/classes/image.py
```diff
@@ -0,0 +1,1050 @@
...
+                {
+                    d: s * image_scale[d] if d in image_scale else 1.0
+                    for d, s in zip(image.axes, scale)
+                }
+            )
+
+        translations = [
+            {d: (scale[d] / 2) - (scales[0][d] / 2) for d in image.axes}
+            for scale in scales
+        ]
```
jo-mueller, line 188: that's the relevant part, right? I think looks good to me, but I'll know better after a rebase 👍

tests/test_writer.py
```diff
@@ -1513,13 +1740,32 @@ def verify_label_data(
...
-            for transf, expected in zip(cts, transformations):
-                assert transf == expected
+        cts = [
+            d["coordinateTransformations"]
+            for d in node_metadata["multiscales"][0]["datasets"]
+        ]
+
+        axes = list(scale.keys())
+        for level, transfs in enumerate(cts):
+            assert len(transfs) == 1
```
jo-mueller: This assertion needs to be changed with respect to the scale level
```suggestion
            if level == 0:
              assert len(transfs) == 1
            else:
              assert len(transfs) == 2
```

will-moore:
@jo-mueller - rebased and fixed tests

jo-mueller (approved):

jo-mueller (approved):

PR #605, opened 2026-07-24
Ready for 06

jo-mueller:
@will-moore maybe one more thing before shipping the latest release. This PR

- replaces 0.6.dev4 by 0.6 in the written and read versions
- adds some more checks. Prior to this PR, some of the version check clauses missed the check for 0.6 so some of the additional metadata (i.e., `omero` and `image-label`) would never be ingested in case a version 0.6x were encountered.
- Adds 0.6 in the tests more broadly

Rebased on #609


tests/test_writer.py
```diff
@@ -434,12 +444,6 @@ def test_image_class_writer(
 
         assert labels_name in list(image.labels.keys())
 
-        if version.version == "0.4":
-            # Validate with ome-zarr-models-py: only supports v0.4
-            Models04Image.from_zarr(out)
-        elif version.version == "0.5":
-            Models05Image.from_zarr(out)
```
will-moore, line 451: Do we need to remove these checks?

ome_zarr/classes/image.py
```diff
@@ -343,7 +343,7 @@ def to_ome_zarr(
         self,
         group: zarr.Group | str,
         storage_options: list[dict[str, Any]] | dict[str, Any] | None = None,
-        version: Literal["0.6.dev4", "0.5", "0.4"] = "0.5",
+        version: Literal["0.6", "0.5", "0.4"] = "0.5",
```
will-moore: The default value needs to be `"0.6"` so that we write the latest version by default.

will-moore:
In order that the `write_image(data)` method creates v0.6 data, we need a `FormatV06` class to be the `CurrentFormat`.
If we also add this to the tests, that looks like...

```
diff --git a/ome_zarr/format.py b/ome_zarr/format.py
index 3b2c33e..583fe18 100644
--- a/ome_zarr/format.py
+++ b/ome_zarr/format.py
@@ -384,4 +384,14 @@ class FormatV05(FormatV04):
         return {"name": "default", "separator": "/"}
 
 
-CurrentFormat = FormatV05
+class FormatV06(FormatV05):
+    """
+    Changelog: added FormatV06 (May 2025): writing not supported yet
+    """
+
+    @property
+    def version(self) -> str:
+        return "0.6"
+
+
+CurrentFormat = FormatV06
diff --git a/tests/test_writer.py b/tests/test_writer.py
index 836b746..4c7bb58 100644
--- a/tests/test_writer.py
+++ b/tests/test_writer.py
@@ -31,6 +31,7 @@ from ome_zarr.format import (
     FormatV03,
     FormatV04,
     FormatV05,
+    FormatV06,
     format_from_version,
 )
 from ome_zarr.scale import _build_pyramid
@@ -58,6 +59,7 @@ TRANSFORMATIONS = [
 FORMAT_VERSIONS = [
     pytest.param(FormatV04, id="V04"),
     pytest.param(FormatV05, id="V05"),
+    pytest.param(FormatV06, id="V06"),
 ]
```

This creates a few failures which I haven't looked into yet....

tests/test_writer.py
jo-mueller: We don't have to, but since the metadata is essentially ingested with ome-zarr-models-py, we are kind of already validating on-the-fly

ome_zarr/classes/image.py
jo-mueller: That's actually a good question. Should we default to 0.6 already? My feeling would be to stay with 0.5 a little longer and change to 0.6 after more tools have caught up?

jo-mueller:
> In order that the write_image(data) method creates v0.6 data, we need a FormatV06 class to be the CurrentFormat.

right...I was hoping to remove the `Format` class at some point (all of its functionality is now present elsewhere). But for the sake of going forward with 0.6, I could add the `FormatV06` class here and mark it for later removal?

ome_zarr/classes/image.py
will-moore: I think it's probably nicer to change the default behaviour at the same time as v0.6 support is released (major release). That's just a single breaking change, and the release announcement and docs say "v0.6 is now the default, if you want v0.5 then specify it".

If you change the default sometime later, that's an additional breaking change that's harder to keep track of.
The mantra "we always write the latest version by default" is easy to understand and remember.

jo-mueller: Ok. Changed the default in dc07c5585c5c2b030b1fc4d58891961ad9d22622

jo-mueller:
> This creates a few failures which I haven't looked into yet....

oh boy. Jep, quite a few. I think this requires a bit of refactoring until all functions that are writing data somehow are 0.6 compliant one way or another 🙈

jo-mueller:
@will-moore ok, this is turning out super-hard, because there are still more than writer paths (class-based, `write_image`/`write_labels`, `write_multiscale`/`write_multiscale_labels`), all of which would somehow need to be updated to write 0.6, where only the first one (class-based) is really fit for it at this point. And I kind of don't want to update code that we want to deprecate in the long run anyway.

So my suggestion:

- I add an option in the `OMEZarrMultiscale` class that it can be constructed either
  - from a single `OMEZarrImage` and the pyramid is built under the hood. The `write_image` and `write_labels` already use that under the hood.
  - **from a list of `OMEZarrImage`s**: This would be new and would be equivalent to writing a pre-computed pyramid. This could then be used in the `write_multiscale` and `write_multiscale_labels` function

This way, we would get a few props
- The testing data for all these three paths (class based, `write_image`, `write_multiscales`) would effectively take the same path through the writer and could be simplified
- Compliance with 0.6 would need to sit only at one place (inside the `OMEZarrMultiscale` class)
- There **are** cases where creating `OMEZarrMultiscale` from pre-computed pyramids is very desirable, for instance for  (down-)loading data. In this case, the resolution levels are effectively precomputed and one would decidedly not want to create the downscaled copies from a potentially large image. If we were to do this, we would download a big image and downscale instead of downloading a small one to begin with.

In the meantime, I would revert the changes here to https://github.com/ome/ome-zarr-py/pull/605/commits/dc07c5585c5c2b030b1fc4d58891961ad9d22622 (before the introduction of the `FormatV06` class) so we can ship the pre-release.

jo-mueller:
Oh no. And tests are failing with something compression related 💀

will-moore:
Just trying to see what changed in the failing tests, looking at installed versions. Since the tests passed previously, `zarr 3.3.0` has been released, and it's only installed on python 3.12 and 3.13 (failing tests). Older zarr is used on the py 3.11 tests which are passing. But I don't see anything obvious in https://github.com/zarr-developers/zarr-python/releases/tag/v3.3.0

will-moore:
Your suggestion sounds good.

Re: deprecating Format classes - I see that `image.to_ome_zarr()` creates `FormatV05()` when writing v0.6 because the underlying `_write_pyramid_to_zarr()` method expects a Format, so trying to untangle these is a going to be a bit of work.

jo-mueller:
> so trying to untangle these is a going to be a bit of work.

Jep. My approach is to remove the metadata writing entirely from the `_write_pyramid_to_zarr` entirely, because the written metadata there is overwritten by the ozmp `model_dump` later anyway. That will make the `write_multiscale_metadata` obsolete.

`_write_pyramid_to_zarr` will only need to check the format to see whether it needs to go for zarr v2 or zarr v3. PR coming soon!

jo-mueller:
For now, I'll pin zarr-python<3.3.0

pyproject.toml
```diff
@@ -19,7 +19,7 @@ requires-python = ">3.11"
 dependencies = [
     "numpy",
     "dask>=2025.2.0,!=2025.12.*,!=2026.1.*,!=2026.2.*",
-    "zarr>=3.0.0",
+    "zarr>=3.0.0, <=3.3.0",
```
will-moore: I think you don't want `=` there, just `<3.3.0`

jo-mueller: Right.

jo-mueller:
Ok, this is looking good. Pinging @d-v-b here - maybe he knows whether something has changed on the zarr side that could have breaking effects? For reference, tests here started failing in https://github.com/ome/ome-zarr-py/pull/605/commits/a9da128a6fe3a9cfc6fada18f40c4bb11c4f3443 after the release of zarr-python 3.3.0.

d-v-b:
can you link me to a log with the test failures in it?

jo-mueller:
https://github.com/ome/ome-zarr-py/actions/runs/30620108525/job/91124450262

d-v-b:
got it, fix incoming

jo-mueller:
Ok, will rebase on #609 or wait until that is in.

ome_zarr/classes/image.py
```diff
@@ -853,13 +853,17 @@ def _write_additional_meta_data(
...
-            elif version == "0.5":
+
+            # in >=0.5, omero metadata goes in the ome attr
+            elif version == "0.5" or version.startswith("0.6"):
                 if "ome" not in group.attrs:
                     raise ValueError("OME-Zarr attributes not found in group")
                 ome = cast(dict, group.attrs["ome"])
-                omero_dict["version"] = version
+                if version == "0.5":
+                    omero_dict["version"] = version
```
will-moore: I think this is some legacy from v0.4 when e.g. we had `"multiscales": {"version": "0.4"...}` but we don't need to set `omero_dict["version"]` for any versions AFAICT.

ome_zarr/classes/image.py
```diff
@@ -840,7 +840,7 @@ def __init__(
     def _write_additional_meta_data(
         self,
         group: zarr.Group,
-        version: Literal["0.6.dev4", "0.5", "0.4"] = "0.5",
+        version: Literal["0.6", "0.5", "0.4"] = "0.5",
```
will-moore: There's no code difference below between version `0.5` and `0.6` but the default should still be `0.6`. Same for a couple of other `def _write_additional_meta_data()` definitions.

docs/source/basic/write_image.ipynb
```diff
@@ -57,14 +57,14 @@
         "As a last step, we write the multiscale image to disk using the `to_ome_zarr` method, which will create a valid OME-ZARR file that can be read by any OME-ZARR compatible viewer.\n",
         "\n",
         "```{hint}\n",
-        "The demonstrated writer method below defaults to writing OME-ZARR version `0.5`,\n",
-        "but also supports writing OME-ZARR version `0.6.dev4` and `0.4`.\n",
+        "The demonstrated writer method below defaults to writing OME-ZARR version `0.6`.\n",
+        "Writing OME-ZARR version `0.5` and `0.4` needs to be explicitly specified.\n",
```
will-moore, line 61: The `write_image()` section below says:

"The default version of OME-NGFF is v0.5, which is based on Zarr v3"

which is a bit confusing given this hint, partly because of using `OME-NGFF` there and `OME-ZARR` here, but also because it doesn't explicitly refer to the behaviour of `write_image()` and `write_multiscale()`. Since those methods *still* default to v0.5 (and don't support v0.6), this probably needs an equivalent "hint" (or something equally visible) to make that clear.

E.g. NB: `write_image()` and `write_multiscale()` do not yet support `v0.6` and will write v0.5 by default.

will-moore:
A couple of minor points above, but otherwise looking good 👍

ome_zarr/classes/image.py
jo-mueller: Ok, I fixed it a bit more globally in aa7b03c94ba5421cbda641fa397b220f2c81af6f

ome_zarr/classes/image.py
jo-mueller: ✅ 40c36f765dfea295a7e752c305d30d2fb566778a

docs/source/basic/write_image.ipynb
jo-mueller: 61bcaee89e5c6db2975360d167d84f2a5b800919

jo-mueller:
@will-moore ok good to go 👍

jo-mueller:
I think merging this also constitutes a pre-release of 0.19.0. Would love to see `OMEZarrScene` in there, but that doesn't have to be out as fast as writing 0.6 in the first place.

will-moore:
This is looking good now. Everything working for me.

The only issues I see is in building the docs:

 - I sometimes see `ContainsArrayError`. e.g.

```
ContainsArrayError                        Traceback (most recent call last)
Cell In[5], line 4
      1 from ome_zarr.writer import write_image
      2 
      3 path = "test_ngff_image.ome.zarr"
----> 4 write_image(data, "test_ngff2.ome.zarr", axes="czyx")
```

Actually, in that line it should use the `path` variable from the previous line. I think it was replaced by a new name to avoid this error in the past in https://github.com/ome/ome-zarr-py/commit/0f01ed349893af1d90faf6aaf5cb62b488a71cac. 
Since the name is unique in the docs now, I guess this is happening due to previous builds that aren't cleaned up. It would be nice if there was a way to clean up or build in a new space each time (not sure where these actually get written to disk?)

 - Also seeing `image.labels == None` when reading in the docs builds. BUT this isn't a problem when I copy the same code into a new python file and run it:

e.g. here...

Screenshot 2026-08-14 at 12 11 30

and also at..

Screenshot 2026-08-14 at 12 10 55

docs/source/basic/write_image.ipynb
will-moore, line 164: remove the `path` assignment from previous line, or use it in this line

jo-mueller: 48fc07f424f0081482f5751b55cc4776b55cbefa

jo-mueller:
@will-moore re the docs issue - is that something you think would need to be fixed here?

will-moore:
No, as long as the deployed docs don't suffer from the same issue - The current docs seem to be OK and are not showing these errors 👍

will-moore (approved):
LGTM, thx 👍

PR #607, opened 2026-07-30
feat: reverse-populate OMEZarrMultiscale class with omero metadata on read

jo-mueller:
Small PR to make sure that we can extract OMERO Metadata via the `OMEZarrMultiscale` classes. For context:

In the write direction, we can pass `contrast_limits`, `channel_names`, `channel_colors` arguments to the `OMEZarrMultiscale` class, which are translated into appropriate `omero` metadata entries. On read, this isn't possible, though. This PR adds minimal translation to the reader path of the `OMEZarrMultiscale` class.

TODO:

- [x] testing

### DIFF
--- a/ome_zarr/classes/image.py
+++ b/ome_zarr/classes/image.py
@@ -1041,35 +1041,37 @@ class OMEZarrMultiscale(OMEZarrMultiscaleBase):
                 self._omero = Omero.model_validate(omero_dict)
 
                 # reverse-populate channel_names, channel_colors, contrast_limits from omero metadata
-                if self._omero.channels is not None:
-                    channel_names = [
-                        ch.label if ch.label is not None else f"Channel {i}"
-                        for i, ch in enumerate(self._omero.channels)
-                    ]
-                    channel_colors = [
-                        (
-                            ch.color
-                            if ch.color is not None
-                            else DEFAULT_COLORS[i % len(DEFAULT_COLORS)]
-                        )
-                        for i, ch in enumerate(self._omero.channels)
-                    ]
-                    contrast_limits = [
-                        (
-                            ch.window.start if ch.window.start is not None else 0,
-                            (
-                                ch.window.end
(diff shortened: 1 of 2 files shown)


ome_zarr/classes/image.py
```diff
@@ -1039,6 +1039,45 @@ def _read_additional_metadata(
...
+                            start = getattr(ch_window, "start", 0)
+                            end = getattr(
+                                ch_window,
+                                "end",
+                                self._images[0].data.dtype.itemsize * 255,
+                            )
+                            contrast_limits.append((start, end))
+                        else:
+                            contrast_limits.append(
+                                (0, self._images[0].data.dtype.itemsize * 255)
```
will-moore, line 1075: Might be nicer to use `np.iinfo` to get the min/max as these will distinguish e.g. `unit8` from `int8` etc:

```
>>> a = np.array([1,2,3], dtype="int8")
>>> np.iinfo(a.dtype).min
-128
>>> np.iinfo(a.dtype).max
127
>>> a.dtype.itemsize * 255
255
```

But we need different handling for floats, using `np.finfo` and cast np value to `float`:

```
>>> a = np.array([1,2,3], dtype="float64")
>>> np.finfo(a.dtype).min
np.float64(-1.7976931348623157e+308)
>>> float(np.finfo(a.dtype).min)
-1.7976931348623157e+308
>>> float(np.finfo(a.dtype).max)
1.7976931348623157e+308
```

The other possibility here, which is used in e.g. vizarr and ome-zarr.js is to use the actual min and max values of the smallest resolution array to provide the start and end values if they're missing.

jo-mueller: > The other possibility here, which is used in e.g. vizarr and ome-zarr.js is to use the actual min and max values of the smallest resolution array to provide the start and end values if they're missing.

That's a neat idea 👍

PR #608, opened 2026-07-31
feat: Accept pyramids

jo-mueller:
Re [this comment](https://github.com/ome/ome-zarr-py/pull/605#issuecomment-5141258404)

This PR introduces a medium-sized refactor of the `OMEZarrMultiscale` class, to not only accept a single instance of an `OMEZarrImage`, but also a *list of prebuilt `OMEZarrImage`s*. This has a number of applications:

- The older write functions `write_multiscale` and `write_multiscale_labels` can now also reroute all their internal functionality thourgh the use of the `OMEZarrMultiscale` class. This channels all writing through a single function and should make maintenance easier.
- There are usecases of working with pre-built pyramids which weren't supported nicely by the `OMEZarrMultiscale` framework before. I.e., remote ome-zarrs are essentially pre-built pyramids and ingesting these is made much easier with this change.
- We can deprecate some of the additional write pathways (i.e., `write_multiscale`, `write_image`) with a clear replacement.
- We can deprecate the `Format` class machinery along with these functions.

Rebased on https://github.com/ome/ome-zarr-py/pull/609

### DIFF
--- a/tests/test_reader.py
+++ b/tests/test_reader.py
@@ -156,7 +156,6 @@ class TestInvalid:
 
     def test_invalid_version(self):
         grp = create_zarr(str(self.path))
-        grp = zarr.open_group(str(self.path), mode="a")  # open in append mode
 
         # update version to something invalid
         attrs = get_metadata(grp)
--- a/tests/test_reader.py
+++ b/tests/test_reader.py
@@ -156,6 +156,8 @@ class TestInvalid:
 
     def test_invalid_version(self):
         grp = create_zarr(str(self.path))
+        grp = zarr.open_group(str(self.path), mode="a")  # open in append mode
+
         # update version to something invalid
         attrs = get_metadata(grp)
         attrs["multiscales"][0]["version"] = "invalid"
--- a/tests/test_writer.py
+++ b/tests/test_writer.py
@@ -42,7 +42,6 @@ from ome_zarr.writer import (
     write_labels,
     write_multiscale,
     write_multiscale_labels,
-    write_multiscales_metadata,
     write_plate_metadata,
     write_well_metadata,
 )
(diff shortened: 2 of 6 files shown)


jo-mueller:
Ok, tests are failing here for the same reason they do over at #605 

@melonora do you happen to know of any particular changes regarding compression in later zarr versionsthat could cause a failure here? 😕

melonora:
Will have a look

jo-mueller:
@melonora already fixed by #609 - thanks, though!

jo-mueller:
Throughts from @jni and @tomaz-vieira:

o	In the future, we might want to introduce class methods (i.e., from _signle_scale, from_pyramid) to the OMEZarrMultiscale class so typing becomes easier and its clear which type is expected. Ultimately, creating the OMEZarrMultiscale through its constructor could even be deprecated.
o	To get there, we could leave #608 in place as it is and add the respective class methods later.

jni:
As discussed in meeting, I'm ok with this change. Going forward, it might be better to use classmethods. Scikit-image recently moved Transforms towards that type of constructor, e.g.

https://github.com/scikit-image/scikit-image/blob/cb1adbd63149f69e9b10be55e686b1fc457dd1a4/src/_skimage2/transform/_geometric.py#L326-L355

and

https://github.com/scikit-image/scikit-image/blob/cb1adbd63149f69e9b10be55e686b1fc457dd1a4/src/_skimage2/transform/_geometric.py#L396-L413

So now you can do:

```
noop = AffineTransform.identity(ndim=4)
estimate = AffineTransform.from_estimate(ref_pts, mov_pts)
```

rather than overloading many (incompatible) kwargs in the same init.

That is, you never have to decide, if someone passes 3-dimensional points and ndim=4, whether to ignore ndim or raise an error.

tests/test_writer.py
```diff
@@ -2440,4 +2201,4 @@ def write_labels_class_API(self, fmt):
 
 
 if __name__ == "__main__":
-    pytest.main([__file__])
+    pytest.main([__file__ + "::TestLabelWriter::test_two_label_images"])
```
will-moore: Is this a temp change to run a single test with `$ python test_writer.py` ?

I tend to do: e.g.

```
$ cd tests/
$ pytest test_writer.py -k test_two_label_images
```

tests/test_reader.py
```diff
@@ -156,6 +156,8 @@ def initdir(self, tmpdir):
 
     def test_invalid_version(self):
         grp = create_zarr(str(self.path))
+        grp = zarr.open_group(str(self.path), mode="a")  # open in append mode
```
will-moore: Why is this change needed?
If it IS needed, the previous line can be removed.

tests/test_writer.py
jo-mueller: Ah. I think that's just leftovers from my debugging. Will remove :+1:

jo-mueller: c81c49ad4a10e236faebb5a8426dbbbf9938dc46

PR #609, opened 2026-07-31
fix(tests): support plain-string codec attributes in zarr >= 3.3

d-v-b:
zarr 3.3.0 changed codec configuration attributes such as `BloscCodec.cname` and `BytesCodec.endian` from `Enum` members to plain strings, breaking the `.name` accesses in the label writer storage options tests (12 failures).

Add a `_codec_value` helper that normalizes both representations, so the tests pass against zarr 3.2 and 3.3 alike. Test-only change; the library code never accessed these attributes.

Assisted-by: ClaudeCode:claude-opus-5

### DIFF
--- a/tests/test_writer.py
+++ b/tests/test_writer.py
@@ -70,6 +70,16 @@ def pytest_generate_tests(metafunc):
         metafunc.parametrize("array_constructor", ARRAY_CONSTRUCTORS)
 
 
+def _codec_value(value):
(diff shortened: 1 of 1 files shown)


jo-mueller (approved):

jo-mueller:
@will-moore any objections to merge?

jo-mueller:
Review with @jni @tomaz-vieira Merging!

PR #610, opened 2026-08-01
docs: convert notebooks to jupytext scripts & centralize dependencies

### LINKED ISSUE #604: Docs build: use jupytext/mystmd to allow markdown as source instead of ipynb
Jupyter notebooks can be annoying to maintain because they are hard to diff, and you can't really edit them in a plain text editor except for the simplest changes. (And still I make mistakes there. 😂)

With jupytext or mystmd, you can store the notebooks in markdown and then they get executed and rendered from that source. The napari/docs repo uses this configuration so we should be able to copy it to here.

### LINKED ISSUE #632: Docs raise errors on local builds
Raised in #623

> Not a bug in this PR, but I got an error building docs because `docs/source/api/classes.rst` has `.. autoclass:: ome_zarr.classes.image.OMEZarrScene` which should be `.. autoclass:: ome_zarr.classes.scene.OMEZarrScene`

Fixed by a2b8b7be7e6fd8e87338bd4b25a4bc5727d32b95

> I also get an error because `test_ngff_image.ome.zarr` is used as an output path in 2 places in the docs, so I get a `ContainsArrayError()` in `build_custom_pyramid.html`.
Although I also get the same `ContainsArrayError` when writing to `test_ngff2.ome.zarr` in basic/write_image.html, and that path isn't duplicated. This error(s) may not show up when deploying the docs "for real", but they are annoying because they may hide other genuine errors.
> 
> labels_metadata page is giving me:
> 
> ```
> AttributeError                            Traceback (most recent call last)
> Cell In[6], line 2
>       1 image = OMEZarrMultiscale.from_ome_zarr("labels_metadata_example.zarr")
> ----> 2 image.labels["test_labels"].image_label
> 
> AttributeError: 'OMEZarrMultiscale' object has no attribute 'image_label'
> ```
> 
> I don't really understand why I'm seeing this error, as the following line `image.labels["test_labels"].image_label.model_dump()` seems to run OK.
> Do you see any problem here?

_Originally posted by @will-moore in https://github.com/ome/ome-zarr-py/issues/623#issuecomment-5437581046_

jo-mueller:
Fixes #604 
Fixes #632 

Rebased on https://github.com/ome/ome-zarr-py/pull/609

Edit: I also added a small piece of code into the conf.py which will clean up any existing .zarr files in the source folders written by previous executions of the docs.

*Edit*: This pushes something over the edge that I meant to do a while ago already: In the current state of main, dependencies for tests, documentation and build are de-centralized in multiple places:
- `docs/requirements.txt`
- `requirements/requirements-dev`
- `requirements/requirements-test``
- `pyproject.toml`

I now centralized all dependencies inside `pyproject.toml`, which should make managing this easier in the future.


jo-mueller:
@jni is that roughly how it is done over at napari? In VSCode, it's relatively easy to spawn these scripts from jupyter notebooks. It's probably equally easy to turn them into markdown files with executable python code blocks.

jni:
> is that roughly how it is done over at napari?

Yes! Jupytext allows many formats, .py with comments is one, md with python blocks is another. It's a matter of preference which one is preferred, and I guess it depends on the balance of narrative vs code.

You're probably missing a whole bunch of dependencies to get napari windows to spawn, I think this is the most important one:

https://github.com/napari/shared-workflows/blob/200f011f642fd89598422aa14d361113a9bd8059/.github/workflows/build_docs.yml#L76-L80

jni:
Seems to be working fine!

It will be nice in follow-up PRs to add visualisation with napari, but I think it's best to do that later since this build is working. I spot-checked a couple of pages to make sure that the output was the same.

jo-mueller:
> It will be nice in follow-up PRs to add visualisation with napari, but I think it's best to do that later since this build is working.

Agree! Depending on who shows up later today on the napari-ome-zarr-py call, I'll probably merge it later today 👍

jo-mueller:
@jni @will-moore @kevinyamauchi could i have eyes here once more? Some more changes here since the last time I updated this PR:

- updated all jupytext documents
- excluded the files from the tests (turned out pytest also ran these as part of the tests, but maybe that's not what we want)
- added a code snippet that deletes all .zarr files inside the docs folder before building the docs so we don't run into file-location-not-empty issues.

I guess it would be nice to merge this before doing more updates elsewhere otherwise updating this along the way will be a bit of a pain ^^"

will-moore:
Docs built fine, with just one warning:

```
/Users/wmoore/Desktop/ZARR/ome-zarr-py/docs/source/advanced/transforms/index.md:: WARNING: toctree contains reference to nonexisting document 'advanced/transforms/reading_scenes' [etoc.ref] [docutils]
```

Pages look good and certainly nicer to have `.py` files rather than `.ipynb`.

However, I think it is nice to have these files run as part of the tests. That should help keeping them valid and up to date?

👍

jni:
> I think it is nice to have these files run as part of the tests. That should help keeping them valid and up to date?

whoops 😂 

>   E   ModuleNotFoundError: The requested file is part of the scikit-image distribution, but requires the installation of an optional dependency, pooch. To install pooch, use your preferred python package manager. Follow installation instruction found at https://scikit-image.org/docs/stable/user_guide/install.html

Looks like we should add pooch and/or scikit-image[data] to the test dependencies.

jo-mueller:
Well this pushes something over the edge that I meant to do a while ago already: In the current state of main, dependencies for tests, documentation and build are de-centralized in multiple places:
- `docs/requirements.txt`
- `requirements/requirements-dev`
- `requirements/requirements-test``
- `pyproject.toml`

I now centralized all dependencies inside `pyproject.toml`, which should make managing this easier in the future.

Regarding executing docs as part of pytest: I just realized that calling `uv run pytest` or just pytest, for that matter, will just execute python files anywhere in the repo, but not actually track them for coverage. Not only that, all the generated .zarr files from the doc scripts are executed with `/` as the working directory, so they'll all dump zarr folders into the root directory on test which is annoying ^^"

There's probably a way to configure pytest to properly track the python scripts in the docs, but I think that's for a different PR.

jni:
🥳 all green!

pyproject.toml
```diff
@@ -47,9 +47,33 @@ authors = [
...
+    "napari",
+    "pyqt6",
+]
+
 [dependency-groups]
 tests = [
     "pytest",
+    "pytest-cov",
+    "pooch",
+    "pydantic==2.12.5",
```
will-moore, line 76: Minor point: `ome-zarr-models` currently has `"pydantic >= 2.11.5, <2.13",` so I'm not sure we need pydantic here at-all?

will-moore:
Thanks for all the cleanup etc.

So, just to clarify - we're not running the docs python as part of regular tests, but that's the same as before this PR and we can think about addressing that later?

The README testing section says `pip install tox` and `tox`, which still works. Anything we want to update there based on this PR?

pyproject.toml
jo-mueller: The pydantic pin was introduced in https://github.com/ome/ome-zarr-py/pull/543 by @pwalczysko - maybe he can give us a tip on whether this is still needed? Otherwise I agree, ome-zarr-models brings pydantic and I'm not sure if we shoot ourselves in the foot at some point if maybe both ome-zarr-models and ome-zarr-py bring their own set of pydantic restrictions.

jo-mueller:
> So, just to clarify - we're not running the docs python as part of regular tests, but that's the same as before this PR and we can think about addressing that later?

Yes

> The README testing section says pip install tox and tox, which still works. Anything we want to update there based on this PR?

Yes! Good catch, thanks. *edit*: Updated in 0b3ba327a758724ec382351a6f0b8bde657748df

pyproject.toml
will-moore: That fix in https://github.com/ome/ome-zarr-py/pull/543 was to avoid `pydantic==3.13.x` which is also covered by `ome-zarr-models` requirements of `"pydantic >= 2.11.5, <2.13"`

will-moore (approved):
LGTM, thx 👍

pyproject.toml
jo-mueller: Ok, pin released in 80c41a5c78a8d20b34acae8cb49b8d9a085baa9e because it's also pinned in ozmp [here](https://github.com/ome-zarr-models/ome-zarr-models-py/blob/6c30f107ac6a415a9e4005d1b5375c5c1da025a1/pyproject.toml#L9).

PR #611, opened 2026-08-03
feat: set scaler to not supersede `scale_factors` arguments

### LINKED ISSUE #606: write_labels overwrittes per-default custom scale_factors
Hello, 

I noticed a minor bug(?) in the function `writer.write_labels`:

By default, `write_labels` seems to overwrite custom arguments passed in for `scale_factors`. This appears to be because the `scaler` argument — although deprecated — still defaults to a value other than `None`. In contrast, `writer.write_image` uses `scaler=None` as its default. The result is asymmetric behavior between the two functions: given the same arguments, they write different pyramids.

Thanks for taking a look into this!
Best,
Johann

jo-mueller:
Fixes #606 

Rebased on #609

### DIFF
--- a/ome_zarr/writer.py
+++ b/ome_zarr/writer.py
@@ -1169,7 +1169,7 @@ def write_labels(
     labels: np.ndarray | da.Array,
     group: zarr.Group | str,
     name: str = "labels",
-    scaler: Scaler | None = Scaler(order=0),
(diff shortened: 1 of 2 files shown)


jo-mueller:
@Reviewed with @jni @tomaz-vieira

PR #612, opened 2026-08-04
Ome zarr scene

jo-mueller:
Clean version of #583. The commit history there is a bit garbled up, especially since I rebased #515 a few times before merging and #583 was forked off that branch earlier.

## New: OMEZarrScene

Builds upon the `OMEZarrMuzltiscale` class framework. Provides the following methods:

- `from_ome_zarr()`: Read a zarr group into a scene class instance to access transforms, coordinate systems and contained ome zarr images easily through the class-based API
- `to_ome_zarr()`: Serialize an ome-zarr scene to disk, with the option to append already existing scenes if writing would be too heavy otherwise

Attributes:
- `images`: Dict of essentially `{group: image}`, where `image` is an instance of `OMEZarrMultiscale`
- `coordinate_transformations`: List of ome-zarr-models-py transform instances
- `coordinate_systems` List of coordinate systems in the `scene` metadata (if they exist)
- `coordinate_displacements`: Same like `images`: Dict that lists stored coordinates or displacements arrays.

Features:
- Graph traversal: Uses [transformnd](https://github.com/clbarnes/transformnd) to build and traverse a transformation graph. The graph is currently a hidden feature (`scene._graph)` but I could imagine exposing it more in the future if we come up with a neat API for it


jo-mueller:
Rebased on #605

jo-mueller:
Both transformnd and ome-zarr-models-py in require `python>=3.12` in their latest versions, so these tests are failing here. Python 3.11 is still [a year away from EOL](https://devguide.python.org/versions/) - so are we cool with dropping support for it?

jo-mueller:
Both ozmp and transformn use the [TypeVar ](https://docs.python.org/3/whatsnew/3.12.html#:~:text=The%20new%20syntax%20allows%20declaring%20TypeVarTuple%20and%20ParamSpec%20parameters%2C%20as%20well%20as%20TypeVar%20parameters%20with%20bounds%20or%20constraints%3A)feature from Python 3.12 extensively, so dropping 3.11 is necessary if we want to use these pacakges in conjunction

ome_zarr/classes/scene.py
```diff
@@ -0,0 +1,469 @@
...
+        possible_coordinate_systems = []
+        for group in self.images:
+            if group == path:
+                img = self.images[group]
+                for cs in img.metadata.coordinateSystems:
+                    if cs.name == name:
+                        return cs
+        return None
+
+    def _build_graph(self):
```
Tomaz-Vieira, line 162: Do you intend for the `OMEZarrScene` to be mutable? Otherwise maybe this method should be folded into `__init__` as there would be no other situation in which one would want to call it.

ome_zarr/classes/scene.py
```diff
@@ -0,0 +1,469 @@
...
+from zarr.storage import StoreLike
+
+from .image import OMEZarrMultiscale
+
+
+class OMEZarrScene:
+    def __init__(
+        self,
+        images: list[OMEZarrMultiscale] | dict[str, OMEZarrMultiscale],
+        coordinate_transformations: Sequence[AnyTransform] | list[dict[str, Any]],
```
Tomaz-Vieira, line 24: Can't you just accept only the `Sequence[AnyTransform]`? I don't think we gain much by having the constructor perform this kind of conversion internally. If a client has raw dicts, they can parse it themselves and handle failures however they find appropriate.

ome_zarr/classes/scene.py
```diff
@@ -0,0 +1,469 @@
...
+        coordinate_systems: (
+            Sequence[CoordinateSystem] | Sequence[dict[str, Any]] | None
```
Tomaz-Vieira: Just as above, I don't think there is much benefit to do the `CoordinateSystem.model_validate` internally; if we need a `CoordinateSystem`, we should say that we need a `CoordinateSystem`.

Also, you can remove the ` | None` and default to `()` so the type of `coordinate_systems` is always a `Sequence` and you have one less type to worry about

ome_zarr/classes/scene.py
```diff
@@ -0,0 +1,469 @@
...
+            self.images = images
+
+        # parse coordinate systems and transforms
+        self.coordinate_systems = self._parse_coordinate_systems(coordinate_systems)
+        self.coordinate_transformations = self._parse_transforms(
+            coordinate_transformations
+        )
+        self.coordinates_displacements = coordinates_displacements
+
+        self.metadata = SceneAttrs(
```
Tomaz-Vieira: if feels like you could avoid having `self.metadata` altogether, as it might be relevant only when serializing, and you can create it on the fly then. 

Also, if you make this class mutable, you have to make sure that the metadata is always in sync with `self.coordinate_systems` and `self.coordinate_transformations`

ome_zarr/classes/scene.py
```diff
@@ -0,0 +1,469 @@
...
+        path: str | None
+            Optional path to disambiguate coordinate systems with the same name. If None, will return the first match with the given name.
+
+        Returns
+        -------
+        CoordinateSystem or None
+            The matching CoordinateSystem object, or None if no match is found.
+        """
+        if path is None:
+            # coordinate system can only be in top-level
```
Tomaz-Vieira: I would not have expected this from the signature. I would expect an intrinsic coordinate system to be findable by its name as long as the name doesn't collide with a scene-coordinate-system

ome_zarr/classes/scene.py
```diff
@@ -0,0 +1,469 @@
...
+from pydantic import TypeAdapter
+from zarr.storage import StoreLike
+
+from .image import OMEZarrMultiscale
+
+
+class OMEZarrScene:
+    def __init__(
+        self,
+        images: list[OMEZarrMultiscale] | dict[str, OMEZarrMultiscale],
```
Tomaz-Vieira, line 23: what do you think of using something like `Pure(Posix)Path` instead of `str` to disambiguate names from paths?

ome_zarr/classes/scene.py
```diff
@@ -0,0 +1,469 @@
...
+                aff = np.eye(max(aff.shape))
+                aff[: aff.shape[0], : aff.shape[1]] = aff
+                tnd_transform = tnd.transforms.Affine(aff, spaces=spaces)
+
+        elif transform.type == "displacements":
+            path_to_dfield = transform.path if transform.path is not None else ""
+            if zarr_context != "" and path_to_dfield != "":
+                path_to_dfield = f"{zarr_context}/{path_to_dfield}"
+
+            if self.coordinates_displacements is not None:
```
Tomaz-Vieira, line 449: I find this function to be a bit dangerous, as it is indirectly required as part of `__init__` via `_build_graph`, but it implicitly requires `self.coordinatea_displacements` to have been initialized beforehand.

ome_zarr/classes/scene.py
```diff
@@ -0,0 +1,469 @@
...
+                transform.input.path if transform.input.path is not None else ""
+            )
+            output_path = (
+                transform.output.path if transform.output.path is not None else ""
+            )
+
+            # zarr_context prepends path with relative path from root
+            # to keep track of global location of coordinate systems in the zarr store
+            if zarr_context != "" and input_path != "":
+                input_path = f"{zarr_context}/{input_path}"
```
Tomaz-Vieira: I feel like this type of code would be made easier if `zarr_context` was a `PurePosixPath`. But more importantly, I think that combining `path` with `name` into a string can generate ambiguous results if `name` happens to look like a path (e.g. with slashes and double dots).

So to be on the safe side, I'd say it's better to identify a space by something like

```python
Tuple[PurePosixpath, str]
```

which is never ambiguous

ome_zarr/classes/scene.py
```diff
@@ -0,0 +1,469 @@
...
+        store: StoreLike
+            A zarr-compatible storage backend (e.g., directory path, in-memory store, etc.)
+        overwrite: bool
+            If True, overwrite all images in the store with the current state of the scene.
+            If False, only write new images that haven't been written before. Existing images in the store will be left unchanged.
+
+        """
+        import shutil
+
+        import tqdm
```
Tomaz-Vieira: I would argue against having human-friendly progress indicators because we never know where the library is going to be used:

If someone uses this library as part of a TUI app, the progress bar would completely wreck the UI. If someone pipes the output into a file, they'd get many lines of progress bars.

Maybe you could have a progress callback parameter? Alternatively you could expose the tasks to the caller (by returning something like a `Iterator[WritingTask]`), so that clients can perform the writes however they want, and log progress however they want.

ome_zarr/classes/scene.py
```diff
@@ -0,0 +1,469 @@
...
+
+            # Write the displacement image
+            subgroup = zarr_group.create_group(
+                f"coordinateTransformations/{disp_path}", overwrite=overwrite
+            )
+            disp_img.to_ome_zarr(subgroup, overwrite=True, version="0.6")
+
+        # Always update scene metadata
+        metadata_dict = self.metadata.model_dump()
+        metadata_dict = _recursive_pop_nones(metadata_dict)
```
Tomaz-Vieira: maybe use `model_dump(exclude_none=True)`? Even so, maybe we shouldn't do that at all, because we don't know if all the fields that contain a `None` have a default value. If ever there is a field that is something like

```python
some_field: int | None # notice: no default value
```

then this would corrupt the metadata.

ome_zarr/classes/scene.py
```diff
@@ -0,0 +1,469 @@
...
+        zarr_group = zarr.open(store, mode=mode)
+
+        # Create a subgroup for each image using its path key
+        for img_path, img in tqdm.tqdm(self.images.items(), desc="Writing images"):
+            # Skip if already written (incremental mode)
+            if not overwrite and img_path in zarr_group:
+                continue
+
+            # Write the image
+            subgroup = zarr_group.create_group(img_path, overwrite=overwrite)
```
Tomaz-Vieira, line 252: If possible, I'd suggest writing to a temporary, ideally invisible, directory right next to the output. That way you can make the write atomic: Either you write everything and then rename the directory to the target, or you fail along the way, but don't end up with a partially-created, corrupted zarr archive.

ome_zarr/classes/scene.py
```diff
@@ -0,0 +1,469 @@
...
+
+        # Load scene metadata
+        scene_metadata = BaseSceneAttrs.model_validate(zarr_group.attrs.get("ome", {}))
+        transformations = scene_metadata.scene.coordinateTransformations
+        coordinate_systems = scene_metadata.scene.coordinateSystems
+
+        # Load all image subgroups, keyed by their zarr path
+        images = {}
+        for tf in transformations:
+            if hasattr(tf, "input"):
```
Tomaz-Vieira: could a transformation ever **not** have an input and output fields? Maybe you meant to check if they were `None`?

ome_zarr/classes/scene.py
```diff
@@ -0,0 +1,469 @@
...
+        scene = OMEZarrScene(
+            images=images,
+            coordinate_transformations=transformations,
+            coordinate_systems=coordinate_systems,
+            coordinates_displacements=coordinates_displacements,
+        )
+
+        return scene
+
+    def __setattr__(self, name: str, value: Any) -> None:
```
Tomaz-Vieira: As mentioned above, I don't think you should be storing `metadata` at all. But if you do decide to store it, I would say you should do this logic inside a `@property` or some accessor method, as it is very surprising to encounter this kind of magic. Also, if you ever change your field names, you'd have to remember to change them here as well.

ome_zarr/classes/scene.py
```diff
@@ -0,0 +1,469 @@
...
+                self.metadata = self.metadata.model_copy(
+                    update={"coordinateSystems": parsed_coordinate_systems}
+                )
+
+        else:
+            # Default behavior for all other attributes
+            super().__setattr__(name, value)
+
+    @staticmethod
+    def _parse_transforms(
```
Tomaz-Vieira, line 353: Again I'd argue that clients should know what they have in hand, and if they need to parse some raw dicts, they should do it themselves

ome_zarr/classes/scene.py
jo-mueller: My motivation to support both was to shield users from having to deal with ome-zarr-models-py classes. If I accepted only `Sequence[AnyTransform]` here, that would have users need to go through
1. reading and understanding the spec
2. building the respective ome-zarr-models-py instance
3. passing it to the scene

My original intention behind the class-based API was to shield/abstract the model class handling as much as possible, but I would also find it a bit overkill to *not* allow ozmp objects. So...accepting both seemed like a reasoanble trade-off.

ome_zarr/classes/scene.py
jo-mueller: Same argument as in https://github.com/ome/ome-zarr-py/pull/612#discussion_r3719117677, I'm hesitant to force people to deal with ozmp objects if it can be avoided.

> Also, you can remove the  | None and default to () so the type of coordinate_systems is always a Sequence and you have one less type to worry about

Legit!

ome_zarr/classes/scene.py
jo-mueller: > if feels like you could avoid having self.metadata altogether, as it might be relevant only when serializing, and you can create it on the fly then.

The intent was to allow a path to edit the metadata of an existing scene, but you're right about the self.coordinate_systems and self.coordinate_transformations attributes, those would need to be properties that get their content from self.metadata in a `getter` method 🤔

ome_zarr/classes/scene.py
jo-mueller: Could totally imagine expanding this. The function would then return, say, a dict of `{zarr_group: [coordinate_systems]}`?

ome_zarr/classes/scene.py
jo-mueller: Fair. I think at the time when this was written, transformnd had a concept of `Spaces` that allowed only strings that has since matured into tuples as space identifiers (or any hashable object? not sure). The `path:name` convention was my workaround for it at the time, but can definitely be changed.

ome_zarr/classes/scene.py
jo-mueller: I think I added it at some point for my own purposes, but maybe the better way to do the following:

- remove the `tqdm` bar here
- set `compute=False` in the loop that writes the individual images and collect all the delayed dask objects that are returned by the `to_ome_zarr` functions
-  If `compute=True` was passed to `scene.to_ome_zarr(...)`, we could simply put this at the end:
  ```python
  if compute:
    da.compute(*delayed)
  ```

Users could then create a [dask progress bar](https://docs.dask.org/en/stable/diagnostics-local.html#progress-bar) outside of the writer application:
```python
pbar = ProgressBar()
pbar.register()

scene.to_ome_zarr(..., compute=True)
```

and that would be a better progress bar anyways, because it would be more informed of the total progress. The *downside* is that I don't know whether it imposes a lot of burden on dask to handle a potentially *massive* job graph 🤔

ome_zarr/classes/scene.py
jo-mueller: Oh man, I totally wasn't aware of the `exclude_none=True` flag! There's even a utility function in `utils.py` that recursively browses written json dicts and pops `None`s 🤦‍♂️ .

This is clearly the easier way.

Aside from that, there are no places in the spec where `None`/`null` has informative value, so removing `None`s is always going to be fine.

ome_zarr/classes/scene.py
jo-mueller: > could a transformation ever not have an input and output fields?

Only if it is placed inside a wrapper transform (i.e., `sequence`). That's why all transforms have `input` and `output` as globally optional fields, even if the restrictions are tighter in the context where the transforms appear. Hence all the annoying type checking here 🙄

ome_zarr/classes/scene.py
jo-mueller: That's probably also a remainder of a past implementation when I designed the `OMEZarrScence` class as a `dataclass`, which abstracts the `property` feature away and hence required using the `__setattr__` to alter fields. I would like to have `metadata` editable (to a degree, i.e., if you knowe what you are doing), but it requires keeping it in sync with the transforms/system attributes, which in turn requires getters and setters 👍

ome_zarr/classes/scene.py
jo-mueller: Hm...I have a subjective dislike for in-line designed function because I find them notoriously difficult to debug and test and I like clean(-ish) inits 🤔

ome_zarr/classes/scene.py
jo-mueller: Pinging @m-albert for thoughts on this :) To summarize what's happening here:

- The writer function we discuss about writes a number of ome-zarrs in a loop, each of which can be computed there and then *or* the delayed objects are collected for later.
- The collected delayed objects can then be sent off to compute all at once.

ome_zarr/classes/scene.py
jo-mueller: b13e0f512a180ea59f59240e4a7c1c595ca4a5e8

ome_zarr/classes/scene.py
jo-mueller: 4fd655b5402c291979ddaf699d9badbdb8372aa7

ome_zarr/classes/scene.py
jo-mueller: faee5e1258ac6692b83a4f487d809a8d8510f102

ome_zarr/classes/scene.py
jo-mueller: Or just a list of tuples?

ome_zarr/classes/scene.py
jo-mueller: 529b9cf40eac1b92d7f07eb7d298b6256665c4b9

ome_zarr/classes/scene.py
jo-mueller: 529b9cf40eac1b92d7f07eb7d298b6256665c4b9

ome_zarr/classes/scene.py
Tomaz-Vieira: I don't think this shields the user from ozmp classes, because you are still using them internally, so users are implicitly expected to either know those classes or have read the raw spec to even be able to craft a raw dict that has the proper shape and types.

Further, if the user makes a mistake, an exception will be thrown further away from where the mistake was (i.e.: it will be thrown from within `OMEZArrScene.__init__` instead of where the user created the parameters. Further, depending on how the ozmp parses the raw dicts, it's totally possible that a user might pass in some dict that has fields that are silently ignored while parsing them.

By only taking the ozmp models (**or whatever other structured class you might want**), the users can be naturally guided into doing the right thing: They come to your class, see that you need some ozmp objects. Then they can go to ozmp class definitions and try to create those, following the docs and type hints there. Once they managed to create those, they can come back to `OMEZArrScene` and look at the next argument type, and so on and so forth.

And if they are using some IDE or LSP that understands the type hints, they can tab-complete to discover fields, which they can't do on raw dicts.

ome_zarr/classes/scene.py
Tomaz-Vieira: > The intent was to allow a path to edit the metadata of an existing scene

But if the metadata is just `coordinate_systems` and `coordinate_transforms`, then editing those fields **is** editing the metadata, right?. I don't think you need to stick exactly to the on-disk shape, except for when you write it back down to disk

ome_zarr/classes/scene.py
Tomaz-Vieira: Nice =)
Yeah, `transformnd` will very politely take any Hashable now =)

ome_zarr/classes/scene.py
Tomaz-Vieira: I kinda think that you should always return the dask object rather than calling `da.compute` yourself. If you take a `compute: bool` argument, then your function must return two different types, and it's harder to tell at a glance which one a client is gonna get. Also, it's trivial for the user to call compute on their own, so we won't be helping them much by doing it ourselves.

> burden on dask to handle a potentially massive job graph

If this turns out to be a problem, maybe you could return an iterator that produces the pieces of the compute graph. Then the caller can process them however they wish:

```python
delayed_computations: Iterator[TheDaskThing] = scene.to_ome_zarr()
for dc in delayed_computations:
    print(f"progress: X%")
    da.compute(dc)
    
# or alternatively

delayed_computations: Iterator[TheDaskThing] = scene.to_ome_zarr()
giant_graph = # collect all delayed_computations
pbar = ProgressBar()
pbar.register()
da.compute(giant_graph)
```

And it is easy to make convenience methods using this approach

ome_zarr/classes/scene.py
Tomaz-Vieira: Maybe this should be two methods? Like `find_by_path` and `find_by_name`? Because `find_by_name` can return multiple coordinate systems, and a client probably also wants to get the path where they occur. On the other hand, `find_by_path` would return one CS (or `None`), and there is no need to return a path, because the caller already knows it.

jo-mueller: Ok, check out dd1070f5b4ef1c4b769344323d834415efb3e491!

ome_zarr/classes/scene.py
jo-mueller: Just noticed that the `compute` parameter hadn't been there until 7db011d6d335b156e99c4f8fc94ad62cb9a71455

jo-mueller: > Also, it's trivial for the user to call compute on their own

I disagree. Knowing some Python + read/write ome-zarrs is one thing, knowing Python+read/write ome-zarrs+dask (and how to compute/schedule/etc jobs) is necessarily a smaller subset of expertise a user may or may not have.

I would think it's alright to have compute as a `default=True` parameter and if you need some more advanced dask/scheduling stuff, then ome-zarr-py will let you do it

Tomaz-Vieira: Fair enough. Though I'd still argue that it would be better to have two methods, one that calls compute and another that doesn't (perhaps with a scarier name for those who don't know about Dask), just so that the return type doesn't change with the parameters.

m-albert: Thanks for the ping! Super cool work here!

> I would think it's alright to have compute as a default=True parameter

> `delayed_computations: Iterator[TheDaskThing] = scene.to_ome_zarr()`

> just so that the return type doesn't change with the parameters

Maybe a way to have all of these would be to always return an iterator over dask arrays, but in the case of `compute=True` simply return the dask arrays backed by the previously written zarr stores ([return_stored=True](https://docs.dask.org/en/latest/generated/dask.array.to_zarr.html)).

To minimize dask graph blow up, I'd say it's probably a good idea to write one array at the time, or in small batches:

```python
import dask
from itertools import batched  # Python 3.12+

for array_batch in batched(darrays, 2):
    writes = [
        array.to_zarr(path_for(array), compute=False)
        for array in array_batch
    ]
    dask.compute(*writes)
```

jo-mueller: > just so that the return type doesn't change with the parameters.

Legit concern. In the `OMEZarrMultiscale` class, the `to_ome_zarr` function always returns a list, which is empty if `compute=True`. Not pretty, but clear.

ome_zarr/classes/scene.py
jo-mueller: I would say that's a feature that's beyond this PR because it would also extend to the writer functionality elsewhere, and not just for the `OMEZarrScene` class.

ome_zarr/classes/scene.py
jo-mueller: My argument would be that the intersection of people with Python + Pydantic + spec knowledge is smaller than that of Python + spec knowledge. So if you are versed with ome-zarr-models-py instances that's fine and you can pass these, but I wouldn't want to make it a pre-requisite.

The other thing is that exposing this choice would make it much harder for us in the future to swap out the model backend, if we ever chose to. For instance, [yaozarrs](https://github.com/tlambert03/yaozarrs) is pretty good in that respect and I wouldn't want a choice of a different model backend lead to a completely different API.

ome_zarr/classes/scene.py
jo-mueller: > have read the raw spec to even be able to craft a raw dict that has the proper shape and types.

Jep. For the `OMEZarrMultiscale` and `OMEZarrImage` classes, I think they are designed now in such a way, that this is actually not needed. Where it comes to the transforms and the scene, writing an API around the transforms stuff that allows you do use the Scene class without also knowing about the underlying spec is much harder. So you're right here, you cannot build a scene transform graph without knowing how these are structured (yet).

> Further, if the user makes a mistake, an exception will be thrown further away from where the mistake was

The error might fly in the init, but wouldn't you always find the traceback from the ozmp error at the end of that? Assuming that that's where users would look for the error, this would put the pressure on ozmp to throw informative warning messages, which I think is reasonable. Personally, I don't care so much which tool throws the error in how deep a traceback, as long as it shows me what I need to do/change.

>  Then they can go to ozmp class definitions and try to create those, following the docs and type hints there.

As the sole ozmp maintainer right now AND writer of the spec, I would rather write it very cleanly in the spec, link to the respective sections here and make sure that ozmp throws useful errors. 

> And if they are using some IDE or LSP that understands the type hints, they can tab-complete to discover fields, which they can't do on raw dicts.

THAT is a good point. I would argue that as a newcomer, your starting point would likely be the examples in the spec document and your workflow would be something like `SomeClass.model_validate(some_dict)`, for which you'd also start with the dict.

The other argument I'd bring against narrowing ourselves to ozmp objects is something I wrote [below](https://github.com/ome/ome-zarr-py/pull/612#discussion_r3728869241): I don't want to promote the choice of ozmp as a model backend to be part of the API just yet. There are [other libraries](https://github.com/imaging-formats/yaozarrs) for that, too, If we decide at some point to swap out the model implementation to reduce duplicity, I wouldn't want to change the API, too.

ome_zarr/classes/scene.py
jo-mueller: Ok, in the latest change, `_build_graph()` is called in init, but also if either of coordinate_transformations or coordinate_systems are changed, which justifies it being a private method better!

ome_zarr/classes/scene.py
jo-mueller: Maybe a PathLike union of `PurePosixPath` and str? That would clearly communicate what's expected here?

ome_zarr/classes/scene.py
jo-mueller: ✅ 0b1ad99f2278544b53a326d68bdc4db748be6838

ome_zarr/classes/scene.py
jo-mueller: Ok, I removed tqdm in 9f851e164f4eb42148751df9a610a20a125ad20a. Maybe we continue this in a separate issue?

jo-mueller:
@will-moore @kevinyamauchi @melonora @Tomaz-Vieira I think this is as good as it gets if you want to give it another look?

ome_zarr/classes/scene.py
```diff
@@ -0,0 +1,520 @@
...
+from .image import OMEZarrMultiscale
+
+
+class OMEZarrScene:
+    def __init__(
+        self,
+        images: list[OMEZarrMultiscale] | dict[str, OMEZarrMultiscale],
+        coordinate_transformations: Sequence[AnyTransform] | list[dict[str, Any]],
+        coordinate_systems: Sequence[CoordinateSystem] | Sequence[dict[str, Any]] = (),
+        coordinates_displacements: dict[str, OMEZarrMultiscale] | None = None,
```
kevinyamauchi:
```suggestion
        coordinate_displacements: dict[str, OMEZarrMultiscale] | None = None,
```

I _think_ it should be coordinate_displacement (no "s"), but I'm not 100% sure.

ome_zarr/classes/scene.py
kevinyamauchi: I think this is a fair point. I could go either way on this. If we want to leave it as is, I would add some tests to ensure `self.coordinates_displacements` is properly initialized in the init.

That being said, it feels to me that all of this conversion code should be pulled off the class into modular functions a util module. It makes it easier to test them and then also the passing of the displacements/transforms becomes explicit.

kevinyamauchi (commented):
I think this is looking good, thanks! I think it makes sense to get this version in so that we have a class to read/write Scenes. We can iterate on the API for querying the Scene in a follow-up.

We discussed this in the bi-weekly [ome-zarr-py dev call](https://ngff.openmicroscopy.org/community/), but just to have it on the record, I agree with @Tomaz-Vieira that it makes sense to restrict the API to accepting the model objects and omit the dictionary input for now. My main rationale are:
- I think even if the user constructs a dictionary, they will still have to look up the model object API to know how to structure the dictionary
- with good documentation we can make it clear where the classes can be imported from.
- it is much easier to expand the API than deprecate it, so I think it makes sense to start slim

The other main point is that I think we need tests for the `_build_graph()` and `_ozmp_tf_to_tnd`. They will be helpful to catch if the libraries (e.g., transformnd) change their API and we have to make an update.

ome_zarr/classes/scene.py
kevinyamauchi: (I'm okay with making the change in the follow-up you are planning)

will-moore:
EDIT: fixed by updating to `ome-zarr-models-1.8.0rc2` - previously had `ome-zarr-models-1.8.0rc0`

For the record, keeping the error below

<details><summary>Validation Error</summary>

When building docs, the `advanced/transforms/scene_2d_to_3d.html` page fails with

```
transform_to_3d = {
    "type": "sequence",
    "input": {"path": "cells3d_slice", "name": "physical"},
    "output": {"path": "cells3d", "name": "physical"},
    "transformations": [
        {
            "type": "projectAxis",
            "createdOutputs": [0, 1]
        },
        {
            "type": "translation",
            "translation": [0, 30, 0, 0]
        }
    ]
}
```

```
ValidationError                           Traceback (most recent call last)
Cell In[6], line 1
----> 1 scene = OMEZarrScene(
      2     images=[ngff_ms, slice_ms],
      3     coordinate_transformations=[transform_to_3d]
      4 )
...
ValidationError: 1 validation error for tagged-union[function-after[_ensure_consistent_input_output(), Identity],function-after[_ensure_consistent_input_output(), MapAxis],function-after[_ensure_consistent_input_output(), Translation],function-after[_ensure_consistent_input_output(), Scale],function-after[check_metadata_set(), function-after[_ensure_consistent_input_output(), Affine]],function-after[check_rotation_matrix(), function-after[check_metadata_set(), function-after[_ensure_consistent_input_output(), Rotation]]],function-after[_ensure_consistent_input_output(), function-after[_ensure_consistent_input_output(), Sequence]],function-after[_ensure_consistent_input_output(), Displacements],function-after[_ensure_consistent_input_output(), Coordinates],function-after[_ensure_consistent_input_output(), Bijection],function-after[_ensure_consistent_input_output(), ByDimension]]
sequence.transformations.0
  Input tag 'projectAxis' found using 'type' does not match any of the expected tags: 'identity', 'mapAxis', 'translation', 'scale', 'affine', 'rotation', 'sequence', 'displacements', 'coordinates', 'bijection', 'byDimension' [type=union_tag_invalid, input_value={'type': 'projectAxis', 'createdOutputs': [0, 1]}, input_type=dict]
```

</details>

jo-mueller:
@will-moore 

> Not clear what's not valid here?

Also not sure. In the latest build the error doesn't fly anymore so I'm inclined to believe that it's RTD-related.

ome_zarr/classes/scene.py
jo-mueller: f7181623d3af8be0495fc9752050813a0d962a40

jo-mueller:
Ok, I think that's all I would want to do here. Otherwise #623 is the better way to continue the work on this, as exposing the ozmp classes a bit more not only affects the `OMEZarrScene`, but also the `OMEZarrMultiscale`, which also accepts additional transforms. Feel free to merge when tests have passed (then I can also put https://github.com/ome/napari-ome-zarr/pull/165 on more stable feet 🙂 )

PR #622, opened 2026-08-18
refactor: class-based API in sample data creator

jo-mueller:
Broken out from #608 .

This PR uses the class-based API in the sample creation functions in the `data` module. A necessary step towards eventually deprecating and removing the `Scaler` class.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -19,7 +19,7 @@ repos:
           - --autofix
 
   - repo: https://github.com/astral-sh/ruff-pre-commit
-    rev: v0.15.18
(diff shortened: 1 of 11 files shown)


jo-mueller:
@will-moore small refactoring change I broke out of #608 where it doesn't belong. Essentially, I'm just modifying the `create_zarr` function to use the image-classes under the hood instead of the deprecated `Scaler` class.

jo-mueller:
@will-moore ok, tests are coming out green here, too 👍

will-moore (approved):
Working fine, code looks good and tests passing 👍

PR #623, opened 2026-08-18
No validation in init

jo-mueller:
Built on top of #612 - this class changes the API of the `OMEZarrMultiscale` and `OMEZarrScene` classes so that they expect validated model class instances instead of raw dicts.

- [x] Made the changes in the classes
- [x] updated tests
- [x] updated docs with more hints

### DIFF
--- a/docs/source/_toc.yml
+++ b/docs/source/_toc.yml
@@ -42,7 +42,7 @@ parts:
     chapters:
     - file: api
       sections:
-      - file: api/image
+      - file: api/classes
       - file: api/writer
       - file: api/reader
       - file: api/io
--- a/docs/source/advanced/transforms/add_transforms_to_multiscales.ipynb
+++ b/docs/source/advanced/transforms/add_transforms_to_multiscales.ipynb
@@ -20,8 +20,9 @@
       "outputs": [],
       "source": [
         "import numpy as np\n",
-        "from ome_zarr import OMEZarrImage, OMEZarrMultiscale\n",
-        "from ome_zarr_models.v06.coordinate_transforms import Rotation, CoordinateSystem"
+        "from ome_zarr_models.v06.coordinate_transforms import CoordinateSystem, Rotation\n",
+        "\n",
+        "from ome_zarr import OMEZarrImage, OMEZarrMultiscale"
       ]
     },
     {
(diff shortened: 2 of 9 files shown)


jo-mueller:
@will-moore @kevinyamauchi @Tomaz-Vieira this is the change we talked about last week - if any of you could give it a look, I think merging this could constitute the proper release of 0.19.0?

will-moore:
First feedback - just on building docs...

Not a bug in this PR, but I got an error building docs because `docs/source/api/classes.rst` has `.. autoclass:: ome_zarr.classes.image.OMEZarrScene` which should be `.. autoclass:: ome_zarr.classes.scene.OMEZarrScene`

I also get an error because `test_ngff_image.ome.zarr` is used as an output path in 2 places in the docs, so I get a `ContainsArrayError()` in `build_custom_pyramid.html`.
Although I also get the same `ContainsArrayError` when writing to `test_ngff2.ome.zarr` in basic/write_image.html, and that path isn't duplicated. This error(s) may not show up when deploying the docs "for real", but they are annoying because they may hide other genuine errors.

labels_metadata page is giving me:

```
AttributeError                            Traceback (most recent call last)
Cell In[6], line 2
      1 image = OMEZarrMultiscale.from_ome_zarr("labels_metadata_example.zarr")
----> 2 image.labels["test_labels"].image_label

AttributeError: 'OMEZarrMultiscale' object has no attribute 'image_label'
```

I don't really understand why I'm seeing this error, as the following line `image.labels["test_labels"].image_label.model_dump()` seems to run OK.
Do you see any problem here?

jo-mueller:
> Not a bug in this PR, but I got an error building docs because docs/source/api/classes.rst has .. autoclass:: ome_zarr.classes.image.OMEZarrScene which should be .. autoclass:: ome_zarr.classes.scene.OMEZarrScene

@will-moore Definitely in scope, as only this PR adds the `OMEZarrScene` class to the documented classes in the API reference 🙊 

I created an issue for the other raised points, will fix in a follow-up PR 👍

kevinyamauchi (approved):
This looks good to me! Thanks, @jo-mueller !

> if any of you could give it a look, I think merging this could constitute the proper release of 0.19.0?

I think once the docs build it working again, a release makes sense.

jo-mueller:
@will-moore any objections here? I'd follow up on the docs issues over in https://github.com/ome/ome-zarr-py/issues/632

will-moore:
Looks good. 
After building the docs, it's kinda nice to list the zarrs (as in #632) and then view in validator...

NB: we need the `-f` here because the `ome_zarr view` doesn't recognise a "scene". I'll open a PR to fix...

```
$ ome_zarr view docs/source/advanced/transforms/test_example_scene.zarr -f
```

Screenshot 2026-08-28 at 12 09 10

will-moore (approved):

PR #628, opened 2026-08-24
Preserve metadata on write

jo-mueller:
Built on top of #623 . In a future (and actually already present) ome-zarr-verse, a zarr group can be both a Scene *and* a plate, i.e., both keys could be present. This PR makes sure that when we are writing to a zarr group in `overwrite=False` aka append mode, we are not deleting metadata that may have already been there under the `ome` key.

✅ Added a test to make sure it works

### DIFF
--- a/docs/source/_toc.yml
+++ b/docs/source/_toc.yml
@@ -42,7 +42,7 @@ parts:
     chapters:
     - file: api
       sections:
-      - file: api/image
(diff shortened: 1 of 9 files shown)


jo-mueller:
Decision: We will defer this to later until it becomes an actual issue. Right now, nuking the metadata is usually fine.

PR #629, opened 2026-08-26
deps: bump ome-zarr-models dependency

jo-mueller:
With ome-zarr-models-py 1.8.0 [released](https://github.com/ome-zarr-models/ome-zarr-models-py/releases/tag/v1.8.0), I'm bumping the dependency here.


will-moore (approved):
LGTM: 👍 💯

PR #631, opened 2026-08-26
Fix failing pre-release test on main

jo-mueller:
@will-moore I'm not *exactly* sure what the action does, but some gh action only running on main was using Python 3.11, which we deprecated support for in some of the recent PRs, so we have failing tests on master at the momet.


will-moore (approved):
LGTM 👍

PR #637, opened 2026-08-28
find_multiscales() handles scenes

will-moore:
This adds support for "scene" to the `ome_zarr view` command.

To test - I found this useful when viewing the output of docs build..

```
$ sphinx-build -b html docs/source/ docs/build/html

# list various zarrs...
$ find docs/ -type d -name "*.zarr"

# view the scene in validator
$ ome_zarr view docs/source/advanced/transforms/test_example_scene.zarr
```

In the 'finder' command the "scene" shows up as a single item (same as plate) in the BioFile Finder table. For each plate or scene, the FilePath passed to BFF is to the first image (so that we see a zarr thumbnail). Unfortunately the thumbnails aren't working for v0.6 data in BFF yet, but you can still e.g. open with Validator for each scene (opens first image).

```
$ ome_zarr finder docs/source/
```

BFF with the "scene" selected:

Screenshot 2026-08-28 at 15 02 51

Showing Thumbnails:

Screenshot 2026-08-28 at 15 03 05

### DIFF
--- a/ome_zarr/utils.py
+++ b/ome_zarr/utils.py
@@ -154,10 +154,20 @@ def find_multiscales(path_to_zattrs):
         if len(wells) > 0:
             path_to_zarr = path_to_zattrs / wells[0].get("path") / field
             plate_name = os.path.basename(path_to_zattrs)
+            # when used for "finder", show the plate as a single entry with thumbnail
             return [[path_to_zarr, plate_name, os.path.dirname(path_to_zattrs)]]
         else:
             LOGGER.info("No wells found in plate%s", path_to_zattrs)
             return []
+    elif "scene" in zattrs:
+        # use first input image as the representative for the scene
+        path_to_zarr = path_to_zattrs
+        for ct in zattrs["scene"].get("coordinateTransformations", []):
+            input_path = ct.get("input", {}).get("path")
+            if input_path is not None:
+                path_to_zarr = path_to_zattrs / input_path
+                break
+        return [[path_to_zarr, path_to_zattrs.name, os.path.dirname(path_to_zattrs)]]
     elif zattrs.get("bioformats2raw.layout") == 3:
         # Open OME/METADATA.ome.xml
         try:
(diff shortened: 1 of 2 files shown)


ome_zarr/utils.py
```diff
@@ -154,10 +154,20 @@ def find_multiscales(path_to_zattrs):
...
+            # when used for "finder", show the plate as a single entry with thumbnail
             return [[path_to_zarr, plate_name, os.path.dirname(path_to_zattrs)]]
         else:
             LOGGER.info("No wells found in plate%s", path_to_zattrs)
             return []
+    elif "scene" in zattrs:
+        # use first input image as the representative for the scene
+        path_to_zarr = path_to_zattrs
+        for ct in zattrs["scene"].get("coordinateTransformations", []):
+            input_path = ct.get("input", {}).get("path")
```
jo-mueller, line 166: Would we need to check for the input AND the output? I.e., if a transform would contain inputs and outputs which are both coordinate systems in different images, only checking the input would leave one out, right?

jo-mueller (commented):
Cool stuff. Just a minor thing regarding how we find members of the scene. You could consider also using the `OMEZarrScene` here:

```python
scene = OMEZarrScene.from_ome_zarr(...)
paths = list(scene.keys())
```

because by default, `from_ome_zarr` will give you a dict of `{path: OMEZarrMultiscale}`

ome_zarr/utils.py
will-moore: @jo-mueller - A scene is handled the same as for plate above: Only 1 item (row) is added to the BioFile Finder table for a Plate or a Scene and we just need a single thumbnail to represent that Plate or Scene. So we just choose the first input image for the Scene.

jo-mueller: Oh ok, I didn't get that the `find_multiscales()` is used here only to produce a thumbnail for the entire scene 👍

jo-mueller (approved):

PR #639, opened 2026-08-31
allow python 3.12

jo-mueller:
@will-moore not sure why the tests for 3.12 passed on main with `requires-python = ">3.12"` there. This PR relaxe this requirement to  `requires-python = ">=3.12"`.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -14,7 +14,7 @@ description="Implementation of images in Zarr files."
 license = "BSD-2-Clause"
 license-files = ["LICENSE"]
 
-requires-python = ">3.12"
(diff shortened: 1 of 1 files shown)


will-moore (approved):
Seems the last 2 commits to that line have mistaken `>` for `>=`:

`>3.12` instead of `>=3.12`
https://github.com/ome/ome-zarr-py/commit/5394cb12c754dc270848952a9fef377f12f5c88d

`>3.11` instead of `>=3.11`
https://github.com/ome/ome-zarr-py/commit/0865736012bda4fc83a9edd889a783a1245fdd22

But your fix is good 👍

PR #644, opened 2026-09-01
chore: precommit fixes

jo-mueller:
Since I merged #599 earlier, a bunch of precommit checks are failing throughout the repo.


jo-mueller:
@will-moore ok, tests are passing here again. I wasn't sure at first whether the failing tests over at #622 were due to something in the PR itself or precommit-induced (got a bit of a heart-attack when I saw all the failing tests). Turns out, I just had to update all the x-fails in the tests (which are Exception-type specific as well 🙄 

Will rebase #622 on this PR once merged and then we can get that in, too (hopefully)

will-moore (approved):
I'm surprised that pre-commit can tell the difference between when a ValueError and TypeError should be used, but 👍 that it can.

PR #652, opened 2026-09-08
bug: correctly normalize resolution level paths

jo-mueller:
Bug I encountered that occurs when a remote ome-zarr is loaded and stored in a different place. If that remote ome-zarr uses a convention for the resolution levels that is **not** `s0`, `s1`, `s2`, etc but rather `0`, `1`, `2`. In that case, the metadata is copied under the hood in the writer path so that when this instance of `OMEZarrMultiscale` is saved, the resolution levels are normalized to `s0`, `s1`, `s2` and the `image.metadata` is not out of sync with the data that was originally opened.

**But**, the writer only updates the `path` field in the `datasets` field, but not the `datasets > coordinateTransformations > input > path` field 🙈 This is fixed by this PR along with a test for it.

*Edit*: @will-moore @kevinyamauchi I think if this goes in we could/should directly ship it as a patch release?

### DIFF
--- a/ome_zarr/classes/image.py
+++ b/ome_zarr/classes/image.py
@@ -401,7 +401,8 @@ class OMEZarrMultiscaleBase:
                 if transform.input is None:
                     raise ValueError(
                         f"Transform input cannot be None in dataset {idx} "
-                        f"transform {transform}")
+                        f"transform {transform}"
+                    )
                 transform = transform.model_copy(
                     update={"input": transform.input.model_copy(update={"path": path})}
                 )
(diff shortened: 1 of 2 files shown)


will-moore (approved):
Tested with:

```
from ome_zarr import OMEZarrMultiscale
url = "https://livingobjects.ebi.ac.uk/idr/zarr/v0.5/idr0062A/6001240_labels.zarr"
ngff_image = OMEZarrMultiscale.from_ome_zarr(url)
path = "6001240_labels_v0.6.zarr"
ngff_image.to_ome_zarr(path)
```

input paths were correct `s0` etc in the parent image and in the labels image. 👍

will-moore:
Yes, go for the patch release 👍

kevinyamauchi (approved):
LGTM! I agree that a release makes sense. When you say "patch release", do you mean `0.19.2` or do you mean `0.19.1.post1`? My preference is `0.19.2` as it is a bug fix

jo-mueller:
Yes, 0.19.2 👍

Issue #656, opened 2026-09-10
Written image does not match in-memory representation


alanocallaghan:
I was trying to create an example image to test an image reader and stumbled across some odd behaviour that seems incorrect to me.

```python
import numpy as np
from ome_zarr import OMEZarrImage, OMEZarrMultiscale

# create random poisson
size_y = 128
size_x = 200
size_z = 5
size_t = 3
rng = np.random.default_rng(0)
data_tzyx = rng.poisson(lam=10, size=(size_t, size_z, size_y, size_x)).astype(np.uint8)

# specify XYZT shape according to size vars above
image_tzyx = OMEZarrImage(
  data=data_tzyx,
  axes=["t", "z", "y", "x"],
  scale={"t": 1.0, "z": 1, "y": 1, "x": 1},
  axes_units={"t": "second", "z": "micrometer", "y": "micrometer", "x": "micrometer"},
)

## write multiscale to disk
created_tzyx = OMEZarrMultiscale(
    image=image_tzyx,
)
created_tzyx.to_ome_zarr("tzyx.ome.zarr")

## read the written image
read_tzyx = OMEZarrMultiscale.from_ome_zarr("tzyx.ome.zarr")

## compare the size of the base level image...?????
print(created_tzyx.images[0].data.shape)
(3, 5, 128, 200)
print(read_tzyx.images[0].data.shape)
(3, 10, 128, 5)
```

I would expect the dimensions of the images to match.

will-moore:
I'm afraid I'm not seeing that output from your code. I get the expected:

```
(3, 5, 128, 200)
(3, 5, 128, 200)
```

I can only guess that you have something else written there already?

To check in validator:

```
$ ome_zarr view tzyx.ome.zarr
```

<img width="1225" height="1304" alt="Image" src="https://github.com/user-attachments/assets/a3707654-481f-405d-ad0a-0a33bc5a4450" />

alanocallaghan:
Ah! Yes my mistake. I suppose this does reinforce my feeling that when `overwrite=False`, the writer should throw an error or warning if the file already exists.

PR #657, opened 2026-09-10
Throw error if overwrite is False but file exists

alanocallaghan:
Mentioned in #656 

I think `overwrite=False` is a fine default especially for potentially large files, but personally if I ask something to write a file I expect it to either

1. Write the file
2. Tell me why it did not write the file

Rather than

3. Execute as though it had written the file while doing nothing.


jo-mueller (approved):

jo-mueller:
@will-moore thoughts? Otherwise I'd merge and release a patch release 👍

will-moore:
I'm happy with this change, but I'm kinda wandering if a warning is enough, rather than throwing an exception?

In zarr-python, I seem to remember that if you try to write a group/array and it already exists, then you get an exception?
Warnings are easier to miss.

@jo-mueller @alanocallaghan would raising exception be acceptable here?
But the change is an improvement in any case, so 👍 .

alanocallaghan:
An exception is fine, maybe preferred for me; the only reason for choosing a warning is to not introduce a breaking change in case anyone relies on this behaviour.

jo-mueller:
I'm OK with an exception here, too 👍

alanocallaghan:
I removed the write_image_data flag logic as becomes unreachable when throwing an exception here

kevinyamauchi:
Sorry for the late reply, @alanocallaghan ! Thanks for the changes. Overall, they look good to me.

It looks like the test failure is because the path in the line below is being passed in as a `path.Path` object and the check in `to_ome_zarr()` doesn't properly process `Path` objects. This was check was pre-existing code, so your PR helped surface a bug we didn't know about. Nice!

The line from the failing test:
https://github.com/ome/ome-zarr-py/blob/f76e2e0597c9403658855e0131235e6e5e8ff3ef/tests/test_writer.py#L303 

The faulty check:
https://github.com/ome/ome-zarr-py/blob/f76e2e0597c9403658855e0131235e6e5e8ff3ef/ome_zarr/classes/image.py#L345-L349

Since the `Path` object is not a `str`, it goes straight to the `store_exist=True` branch. I think we should change the logic to something like this:

```python
# Determine if store already exists
if isinstance(group, zarr.Group):
    store_exist = True    # zarr.Group was passed in, so it exists
else:
    store_exists = os.path.exists(group)
```

@alanocallaghan , do you think you have time to make this update? No worries if not. I think @jo-mueller or I could push the fix if you prefer. Thanks again!

PR #660, opened 2026-09-17
feat: narrow return types

jo-mueller:
Small QOL feature for the `from_ome_zarr` methods attached to the `OMEZarrMultiscale` and `OMEZarrLabels` classes. Right now, both inherit the `from_ome_zarr` method from a common base class, which has the return type `OMEZarrMultiscale | OMEZarrLabels`. This is annoying for linters elsewhere, because you'd always need to assert whether the returned object is of either class.

In this PR, I am adding small overrides to the `from_ome_zarr` methods in the derived classes `OMEZarrMultiscale` and `OMEZarrLabels`, which casts the return type to the respectively correct one.


kevinyamauchi (approved):
Thanks, @jo-mueller ! I think this will indeed be a quality of life improvement. This looks good to me.

jo-mueller:
@will-moore any objections?

will-moore (approved):
LGTM 👍

Issue #627, opened 2026-08-22
Does `OMEZarrMultiscales` need a transform graph, too?


jo-mueller:
Title. Currently, building a transform graph only happens in the OMREZarrScene` class, but you can also build transform graphs in the metadata of a single multi scales class (albeit simpler ones). Maybe that would best be accessed through the same structure, I.e. a transform graph? Building the scene graph would then simplify to connecting these sub-graphs into a single graph.

jo-mueller:
@clbarnes quick question about this: If I have to instances of `TransformGraph` from transformnd, is it possible to connect them into a joint graph object by adding a single (or more) transform that links the two?

clbarnes:
Not currently, but sounds like it could be a useful feature!

clbarnes:
I have just merged the space removal PR (if you haven't already pinned your transformnd version, you should do that), then added a couple of utilities useful for graph-merging. This is currently being released in v0.8.

```python
for spaced_transform, edata in graph2:
    graph1.add_transform(spaced_transform, edge_data=edata)
```

and then, if there aren't already shared spaces, add the new edges manually.

You can also relabel spaces (e.g. if you want to make sure that there are no overlapping spaces) with `my_transform_graph.relabel_spaces()`, which takes a dict-like mapping or a function from the old space name to the new one.

jo-mueller:
OK, so a plan to do this could be:
 
- Move the `_ozmp_tf_to_and()` function out of the `OMEZarrScene` class into a utility module, which will anyway be helpful for #626
- Add a `_build_graph` function to the `OMEZarrMultiscale` class or the `OMEZarrMultiscaleBase` class.
- In the `_build_graph` function of the `OMEZarrScene` class, first retrieve all the images' individual transform graphs and connect them via the transforms specified in the scene
-...?

PR #662, opened 2026-09-18
Transformnd 0.9.1

clbarnes:
Also makes the scene-building typing a little more robust by having different methods for getting transforms which need to have spaces and transforms which don't.

The helper method explicitly returns None where a transform can't be constructed (with a log warning), so that the graph construction method can ignore it.

Note: affines and rotations do not support the path form. ome-zarr-models has a property to abstract over the inline/ path forms (Affine.affine_matrix / Rotation.rotation_matrix) which, for now, just raises an error if the path form is given.


jo-mueller:
Conflict with #661 incoming 🙈

clbarnes:
Ah right - would it be helpful to highlight the important changes I made here so that they can both go into the same PR? Or if you think you'll be done with #661 pretty soon I can just re-do my changes on top of that.

jo-mueller:
It's probably easier to move ahead with this one first. #661 just moves the transformed stuff to a different module, so that shouldn't be so hard to rebase. I'd rather have tnd 0.9.1 in first so we can make use of its utilities (graph combinations, etc) here as well.

jo-mueller:
Still, we may end up with an ome-zarr-py version that will not be available on conda-forge

ome_zarr/classes/scene.py
```diff
@@ -74,7 +89,7 @@ def __init__(
             coordinateTransformations=tuple(coordinate_transformations),
         )
 
-        self._build_graph()
+        self._graph = self._setup_graph()
```
clbarnes, line 87: I changed setup_graph to create and return the graph so that all instance variables are set explicitly in `__init__`; easier keep track of them that way. It also makes the method easier to test on its own.

ome_zarr/classes/scene.py
```diff
@@ -58,14 +71,16 @@ def __init__(
...
-
         """
         # Coerce list to dict keyed by metadata.name
         if isinstance(images, list):
             self.images = {str(img.metadata.name): img for img in images}
         else:
             self.images = images
 
+        if coordinate_displacements is None:
+            coordinate_displacements = dict()
```
clbarnes, line 77: It didn't look like there was a distinction between None and an empty dict; this makes type checking easier.

ome_zarr/classes/scene.py
```diff
@@ -166,18 +181,23 @@ def _build_graph(self):
...
+                tnd_transform = self._ozmp_tf_to_tnd_spaced(
+                    tf,
+                    zarr_context="",
+                    source_cs=source_cs,
+                    target_cs=target_cs,
+                )
+            except UnsupportedTransformation as e:
+                # error message contains transform info
+                logger.warning("Skipping unsupported transformation: %s", e)
+                continue
```
clbarnes, line 190: These methods skip (with a loud warning) rather than error to support the case where certain edges depend on transforms we don't support yet. Better to leave a graph with holes in (which will be necessary in future with extensions) rather than just refuse to do anything.

ome_zarr/classes/scene.py
```diff
@@ -342,132 +368,201 @@ def from_ome_zarr(cls, store: StoreLike):
...
 
-    def _ozmp_tf_to_tnd(
+    def _ozmp_tf_to_tnd_spaced(
         self,
         transform: AnyTransform,
         zarr_context: str = "",
         source_cs: CoordinateSystem | None = None,
         target_cs: CoordinateSystem | None = None,
-    ) -> tnd.base.Transform:
+    ) -> tnd.Spaced:
```
clbarnes: Raises an error rather than returning None; consumers can catch this.

ome_zarr/classes/scene.py
```diff
@@ -342,132 +368,201 @@ def from_ome_zarr(cls, store: StoreLike):
...
+            )
+
+        spaces = (
+            (input_path, transform.input.name),
+            (output_path, transform.output.name),
+        )
+
+        return tnd.Spaced(t, *spaces)
+
+    def _setup_vectorfield_args(
```
clbarnes: This method can be reused for coordinates, once they're supported.

ome_zarr/classes/scene.py
```diff
@@ -342,132 +368,201 @@ def from_ome_zarr(cls, store: StoreLike):
...
+            UnsupportedTransformation
+                Transformation cannot be recreated in transformnd.
+        """
+        import numpy as np
 
-        tnd_transform = None
         # Example for an affine transformation (this will depend on the actual structure of AnyTransform)
-        if transform.type == "affine":
-            aff = np.asarray(transform.affine)
+        if isinstance(transform, ozmt.Affine):
```
clbarnes: `isinstance` here rather than `transform.type == ...` makes it play nice with type checking.

ome_zarr/classes/scene.py
```diff
@@ -342,132 +368,201 @@ def from_ome_zarr(cls, store: StoreLike):
...
+                transform.translation,
             )
 
-        elif transform.type == "rotation":
-            affine_matrix = np.eye(len(transform.rotation) + 1)
-            affine_matrix[:-1, :-1] = transform.rotation
-            tnd_transform = tnd.transforms.Affine(affine_matrix, spaces=spaces)
+        elif isinstance(transform, ozmt.Rotation):
+            try:
+                rot = transform.rotation_matrix
```
clbarnes: rotation_matrix in ozmp is intended to abstract over inline or path-form transformations, but the path-form isn't implemented, so we handle that here.

ome_zarr/classes/scene.py
```diff
@@ -342,132 +368,218 @@ def from_ome_zarr(cls, store: StoreLike):
...
             ]
-            tnd_transform = tnd.base.TransformSequence(
-                tnd_sub_transforms,
-                spaces=spaces,
+
+            return tnd.transforms.ByDimension(tnd_sub_transforms)
+
+        elif isinstance(transform, ozmt.Sequence):
+            ts = transform.transformations
+            if not transform.transformations:
```
clbarnes: Only passes in coordinate systems where they're actually applicable - i.e. start and end of the sequence. If we have to e.g. infer dimensionality mid-sequence, that's a different problem. This is the case for identity and projectaxis.

ome_zarr/classes/scene.py
```diff
@@ -342,132 +368,233 @@ def from_ome_zarr(cls, store: StoreLike):
...
+                        self._ozmp_tf_to_tnd(
+                            ts[0], zarr_context, source_ndim, target_ndim
+                        )
+                    ]
+                )
+
+            inner = [self._ozmp_tf_to_tnd(ts[0], zarr_context, source_ndim, None)]
+            for t in ts[1:-1]:
+                src_ndim = inner[-1].ndims.target
+                inner.append(self._ozmp_tf_to_tnd(t, zarr_context, src_ndim, None))
```
clbarnes: Here we only give coordinate system information to the transforms it actually applies to.
Only identity and projectAxis require a dimensionality and in both cases can infer it if given only one.

ome_zarr/classes/scene.py
```diff
@@ -342,132 +368,233 @@ def from_ome_zarr(cls, store: StoreLike):
...
+            tnd.transforms.Scale(list(img.scale.values())[1:]),
+            VECTOR_FIELD_AXIS,
+        )
+
+    def _ozmp_tf_to_tnd(
+        self,
+        transform: AnyTransform,
+        zarr_context: str = "",
+        source_ndim: int | None = None,
+        target_ndim: int | None = None,
```
clbarnes: As this method only uses coordinate systems' dimensionality, and sometimes needs to be used where we know the dimensionality but not any other information about the coordinate system (i.e. within sequence or bydimension), we only pass in the ndim here.

This could be a problem if we wanted to e.g. validate that the vector field is in the right position, but we weren't doing that here anyway.

ome_zarr/classes/scene.py
```diff
@@ -342,132 +368,235 @@ def from_ome_zarr(cls, store: StoreLike):
 
         return scene
 
-    def _ozmp_tf_to_tnd(
+    def _ozmp_tf_to_tnd_spaced(
```
jo-mueller: Not sure I'm a fan of double-wrapping the conversion into `ozmp_tf_to_tnd_spaced` and `_ozmp_tf_to_tnd` functions? I think this can all just be a single conversion function.

ome_zarr/classes/scene.py
```diff
@@ -342,132 +368,235 @@ def from_ome_zarr(cls, store: StoreLike):
...
+                ) from e
             if aff.shape[0] == aff.shape[1]:
-                tnd_transform = tnd.transforms.Affine(
-                    transform.affine,
-                    spaces=spaces,
-                )
+                return tnd.transforms.Affine(aff)
             else:
                 aff = np.eye(max(aff.shape))
                 aff[: aff.shape[0], : aff.shape[1]] = aff
```
jo-mueller: Bug I discovered here (not introduced by this PR),  can fix it here if we're already at it: this overwrites the affine matrix with an identity 😬

jo-mueller (changes requested):
I *think* I'm generally fine here, but I think the entirely logic can just go inside a single function instead of the double-wrapping here. This also makes the additional helper functions at the bottom (e.g., `_get_ndim`) redudnant, I think?

jo-mueller:
Another thing I caught elsewhere: in transformnd's `Displacement` class, the index transform projects from physical space into array coordinates, right? This is typically handled the other way round for the resolution level transforms in the spec, so we'd need a `.invert()` for the conversion of the displacements (and coordinates) transform here.

ome_zarr/classes/scene.py
clbarnes: This is actually quite important to the refactor. By separating the spaced and unspaced conversions, we can throw more informative errors when we're missing space information but need it, and don't have to make up spaces or add a bunch of if-this-then-that logic when we're converting wrapped transforms. It's a long function that could use decomposing anyway, and splitting it into the purely transform-concerned part (which can be called recursively without issue), plus the space-validating part, is a good separation.

The split is even cleaner once it's in another module. The graph-builder in the Scene class only needs to call the spaced conversion function, and the unspaced function is simply an internal implementation detail which nobody else needs to care about.

ome_zarr/classes/scene.py
clbarnes: Good point! While we're here, what's the deal with treating square affines differently to rectangular? I thought the spec was clear that stored affines should not include the bottom row, i.e. we should always do the write-into-identity form? I've done that (effectively) in my latest commits.

clbarnes:
Good point about the index transformation inversion, I'll add that.

I'm ok with keeping the small helper functions; they're only called a couple of times each but handling nulls is annoying in python, and the extra function makes it more clear what's happening inline.

As I mentioned in the comment reply, I am pretty set on keeping the separation between the transform conversion and then adding spaces on top of the unspaced transform. In general quite keen on keeping functions shorter and with clear responsibilities, even if they're only used within the one public function.

ome_zarr/classes/scene.py
jo-mueller: Can probably be removed, this is probably a historic remnant. I could imagine that I originally used this to parse rotation matrices into affines before I added the conversion for rotation matrices as a separate clause, which is the correct thing to do. So...that distinction here could be removed, too.

ome_zarr/classes/scene.py
jo-mueller: Ok, that sounds clean to me. Maybe add a note about the responsibilities iun the docstring so that's clear for others to infer in the future ?

jo-mueller:
> As I mentioned in the comment reply, I am pretty set on keeping the separation between the transform conversion and then adding spaces on top of the unspaced transform. In general quite keen on keeping functions shorter and with clear responsibilities, even if they're only used within the one public function.

Sounds good to me, thanks for the explanation!

I think the codecov complaining comes mostly from more escapes. Much of that should be captured by a test with an invalid transform. At this point, the effort of moving the whole code around the `ozmp_tf_to_tnd` functionality into a `utils.py` in the same module (which #661 does) could also be part of this PR? THis would then supersede #661 and make this functionality easier to test and cover.

Another nitpick: In ome-zarr-py I mostly refer to displacements as displacements and I think tnd calls this vector fields? Maybe we can stick to the displacements naming convention here?

Thanks for working on this!

clbarnes:
> At this point, the effort of moving the whole code around the ozmp_tf_to_tnd functionality into a utils.py in the same module

Agreed - I did already start that in another branch but didn't submit it because it would have made the changes much harder to review. I might do it in a second round if that's OK?

> In ome-zarr-py I mostly refer to displacements as displacements and I think tnd calls this vector fields?

In tnd I use "vector fields" to refer to coordinates + displacements. As the helper method here would apply equally to both (although ozp currently only implements displacements - I added coordinates in this PR), I went with vector fields there.

I'll add some more tests and refactor out into a separate module.

jo-mueller:
> I might do it in a second round if that's OK?

Folow-up PR you mean? Cool 👍

jo-mueller:
Re tests: Feel free to scavenge from #661

jo-mueller:
> In tnd I use "vector fields" to refer to coordinates + displacements. As the helper method here would apply equally to both (although ozp currently only implements displacements - I added coordinates in this PR), I went with vector fields there.

In the argument to the instantiation of the scene, this is called `coordinate_displacements`, but fair if it's too bulky here.

jo-mueller:
@clbarnes copilot flagged a few more issues, I leave it to you to assess their relevance

clbarnes:
This has necessitated adding a get_coordinate_system method to the Multiscale class, so that a graph associated with a scene can resolve the child multiscales' coordinate systems (and maybe even their grandchild labels' coordinate systems).

jo-mueller:
> This has necessitated adding a get_coordinate_system method to the Multiscale class, so that a graph associated with a scene can resolve the child multiscales' coordinate systems (and maybe even their grandchild labels' coordinate systems).

Sounds we are already converging a bit towards resolving #627 as well? 😅 Ultimately, this will make many things a lot easier for consumers. For instance in https://github.com/ome/napari-ome-zarr/pull/169 I am building some specialcasing around resolving transforms inside single multiscales. This will be a lot easier if the `OMEZarrMultiscale` would own its own transformation graph object.

